Issue #3, opened 2014-04-09
Get arrays from BytesIO objects without a copy


mdboom:
Is it even possible?

PR #46, opened 2014-10-23
WIP: Generalized WCS -- try #3

mdboom:
This is a work-in-progress to get the generalized WCS stuff serializing to/from ASDF.

All of this is written against embray/astropy@9307fbafe741 (the compound modeling branch).

There's a few things in modeling that I either don't understand or need improvement:

- [ ] Models don't have equality defined, so it doesn't seem to be possible to serialize a model from disk, reload it and compare it against the original.  Is that a hard thing to add?

- [ ] I've had to use some private APIs related to compound models (see `compound.py` below).  Any thoughts about how to provide a public API to what I need?

- [ ] modeling only has an AffineTransformation2D.  What would be involved in making it n-dimensional?  In general, I don't think we should have separate classes for different dimensionality, with the exception of the very-specialized projections (TAN, XPH etc.) where n-dimensionality doesn't make sense.

### DIFF
--- a/pyasdf/asdftypes.py
+++ b/pyasdf/asdftypes.py
@@ -3,11 +3,10 @@
 
 from __future__ import absolute_import, division, unicode_literals, print_function
 
-import os
-
 from astropy.extern import six
 from astropy.utils.misc import InheritDocstrings
 
+from . import tagged
 from . import versioning
 
 
@@ -52,18 +51,20 @@ class AsdfTypeMeta(type):
         cls = super(AsdfTypeMeta, mcls).__new__(mcls, name, bases, attrs)
 
         if hasattr(cls, 'name'):
-            if 'yaml_tag' not in attrs:
-                cls.yaml_tag = format_tag(
-                    cls.organization,
-                    cls.standard,
-                    versioning.version_to_string(cls.version),
-                    cls.name)
+            if isinstance(cls.name, six.string_types):
+                if 'yaml_tag' not in attrs:
+                    cls.yaml_tag = cls.make_yaml_tag(cls.name)
 
-            AsdfTypeIndex._type_by_cls[cls] = cls
-            AsdfTypeIndex._type_by_name[cls.yaml_tag] = cls
+                AsdfTypeIndex._type_by_name[cls.yaml_tag] = cls
+            elif isinstance(cls.name, list):
+                for name in cls.name:
+                    AsdfTypeIndex._type_by_name[cls.make_yaml_tag(name)] = cls
+            elif cls.name is not None:
+                raise TypeError("name must be string or list")
 
-            for typ in cls.types:
-                AsdfTypeIndex._type_by_cls[typ] = cls
+        AsdfTypeIndex._type_by_cls[cls] = cls
+        for typ in cls.types:
+            AsdfTypeIndex._type_by_cls[typ] = cls
 
         return cls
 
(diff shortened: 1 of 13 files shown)


mdboom:
@embray: This has been updated to work with current astropy master and is passing the tests (hooray!).  The one outstanding issue is using the `._tree` member of compound models in order to introspect the tree and serialize it to YAML.  (See `compound.py` for what exactly it does).  Any thoughts about how to make a public API for that?  Maybe just add a property `tree` that returns `_tree`?

embray:
That might be a start.  How do you feel about the expression tree interface itself?  I think it's fairly easy to use, but part of me wants something "better" where exactly what that is remains undefined.

I'm also increasingly unhappy with how the tree is represented internally.  For a binary tree like this it would be easy to represent the nodes as a flat list rather than a nested data structure, while still providing the same user-level interface.  There are several algorithms that I feel like would be simpler if the tree were represented this way, rather than having to traverse the nodes as is currently necessary.  But that's more of an internal detail...

mdboom:
@embray: On the naming issue: How about this:

- `&`: concatenate
- `|`: compose

nden:
@mdboom What is the plan for this PR? I am using it locally but it would be easier if it's merged now, even if things are changed later.

mdboom:
I've been working on it a bit but mainly been sidetracked.  I'm fine with merging it and continuing to improve it incrementally, though.

mdboom:
I'll give Travis-CI a chance to test this, and then merge.

nden:
Great! Thanks!

And a question - given an asdf file, I see how I can read/open it, but how can I generate a model object from it?

mdboom:
The work to join `jwst_lib.models` and ASDF is still ongoing.  It's been the bane of my existence for some time...

nden:
Sorry, overloaded term again. How do I generate an astropy.modeling model? For example, given a rotated2d.asdf file (and since there's a schema for it), what is the command to get a `modeling.Rotation2D` object. I feel like I'm missing something simple.

mdboom:
:laughing: overloading models again...

You would do `asdffile.tree['my_model']` to get it (where `my_model` is just wherever you happened to store the transform in the tree.  The conversion from modeling objects to YAML and back happens automatically.

nden:
Here's my asdf file:

```
#ASDF 0.1.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/0.1.0/
--- !core/asdf
smallrotation: !transform/rotate2d
  angle: 2.4
...
```
I get a dictionary
```
 rot=pyasdf.AsdfFile.read('rot.asdf')
 rot.tree['smallrotation']
Out[33]: {u'angle': 2.4}
```
Is my asdf file wrong?

mdboom:
Ah -- it's a bug with this PR.  The transform classes weren't getting imported (and the unit tests did this, so they weren't catching this).  I'll have an update here momentarily.

mdboom:
I think this is working now, but @nden, would you mind double-checking before I merge?

nden:
Yep, it works.

nden:
I think the name of this method `asdffile.tree` is confusing. I didn't think it was a bug since it returned a dictionary and I thought it represented the tree of models.

mdboom:
It's the tree of objects stored in the ASDF file.  http://asdf-standard.readthedocs.org/en/latest/tree.html  

You can store transforms anywhere in it -- it's just JSON/YAML data structures.  I agree it's not the greatest name -- it's what I've been using to refer to the main heirarchical content of the ASDF file.  Any suggestions?

nden:
I guess I expected `.tree` to return the YAML data structure and something else to create the object, something like "instantiate" which is a terrible name. May be it's just me,  not very good at names.

PR #50, opened 2014-12-23
Add support for auto-inlining of arrays

mdboom:


mdboom:
This should be done after #64 and updated to use the new API if necessary.

PR #54, opened 2015-01-27
Replace float128 and complex256

migueldvb:
Changing from np.float128 and np.complex256 to np.longdouble and longcomplex that exist in 64 and 32 bit systems (as float96 and complex192 objects).


mdboom:
Shouldn't we do all of them (to support non-Intel systems)?  Or at least all that we find in the currently installed Numpy...

migueldvb:
It looks like  this change works fine with 32-bit and 64-bit Linux systems. np.longdouble is just an alias to np.float128 on 64bit and np.float96 on 32 bit, and the same for the longdouble alias. I think that this would also work on Windows but I have not tried it...

mdboom:
I see.  In that case, I have no objection.  All that's important here is that *all* complex and float types on a given installation of Numpy are covered.

migueldvb:
The last commit replaces the data types by the generic names that are available in the Numpy data type API. They should be the same on 64-bit as the previous ones and I think that only `longlong` and `longdouble` may differ in 32-bit platforms. These types could work also on non-Intel systems but I haven't tested it.

mdboom:
Ok -- I'm a bit lost.  What's the advantage of using the generic names?  I would think that using the specific names, we'd be sure to have them all covered, but with the generic names some may be missed...

migueldvb:
I think the generic names would be more portable, but maybe we only need to replace the float128 and complex256 by the generic names since these may not be available on some systems. In that case you can ignore the last commit.

mdboom:
Actually (and I wasn't aware of this until looking for it just now), Numpy has a class heirarchy for these things: http://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html#specifying-and-constructing-data-types

So we should be able use `numpy.floating.__subclasses__()`, `numpy.complexfloating.__subclasses__()` and `numpy.integer.__subclasses__()` for these.

mdboom:
And, specifically, we'd need all subclasses recursively, using something like this: http://code.activestate.com/recipes/576949-find-all-subclasses-of-a-given-class/

migueldvb:
Thanks, I will check that method.

mdboom:
Are you cool with https://github.com/migueldvb/pyasdf/pull/1 ?  If so, go ahead and merge that and then I can merge this.

PR #55, opened 2015-01-27
Add numpy dependency to setup.py script

migueldvb:


mdboom:
Isn't this implied by the astropy dependency?

migueldvb:
I think that is right, probably this is not necessary.

PR #56, opened 2015-02-03
add polynomial implementation

nden:
This PR goes with #49 in asdf-standard.

### DIFF
--- a/pyasdf/tags/transform/__init__.py
+++ b/pyasdf/tags/transform/__init__.py
@@ -6,4 +6,4 @@ from __future__ import absolute_import, division, unicode_literals, print_functi
 from .basic import *
 from .compound import *
 from .projections import *
-from .polynomials import *
+from .polynomial import *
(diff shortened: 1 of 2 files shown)


nden:
@mdboom Can you clarify how to deal with a custom inverse for the polynomial type. I tried adding an "inverse" property to the schema and setting it in the implementation as `model.inverse` but this didn't work.

mdboom:
The inverse property exists in the "base class" of transform that all other transforms include, and it's handled there as well.  But it may not be working, because it was kind of half-baked when things got merged.  I'll poke at it for a bit.

mdboom:
This PR will need some tests, which probably means updating the asdf-standard git submodule to point to spacetelescope/asdf-standard#49 so Travis works.

mdboom:
#57 makes inverse transforms work.  I'll merge it as soon as Travis passes.

nden:
I'm not sure how to update the submodule to a PR.

nden:
@mdboom I'm working on tests (but running into some issues). Will open a separate PR.

mdboom:
Oops.  Sorry I merged too soon...

mdboom:
Also, what I merged fails on Travis currently...  (I think the github page hadn't updated for me and I was seeing an earlier build that was passing).  I'm just going to remove this merge from master -- you should be able to create a new PR from your branch and we'll just do everything altogether once the tests are written and passing.

PR #59, opened 2015-02-05
Add polynomials

nden:
@mdboom Submitting this as a new PR. Also added tests and they all pass locally.

### DIFF
--- a/pyasdf/tags/transform/polynomial.py
+++ b/pyasdf/tags/transform/polynomial.py
@@ -79,10 +79,10 @@ class PolynomialType(TransformType):
 
         if n_dim == 1:
             model = modeling.models.Polynomial1D(coefficients.size - 1)
-            model.parameters[:] = coefficients
+            model.parameters = coefficients
         elif n_dim == 2:
             shape = coefficients.shape
-            degree = shape[0]
+            degree = shape[0] - 1
             if shape[0] != shape[1]:
                 raise TypeError("Coefficients must be an (n+1, n+1) matrix")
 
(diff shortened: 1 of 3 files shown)


asdf-standard
```diff
@@ -1 +1 @@
-Subproject commit f6421d381ba2b6308259b781de6434f05c879338
+Subproject commit cf9f9b8ac4d7a830f12fed5bc64f3078d5e31b9a
```
mdboom: I think Travis is failing because this points to one commit from the end of your nden/asdf-standard/add-polynomials branch, rather than 300f62, which has the fixes for the examples.

nden:
:fireworks:

mdboom:
:tada:

pyasdf/tags/transform/polynomial.py
```diff
@@ -0,0 +1,121 @@
...
+            if shape[0] != shape[1]:
+                raise TypeError("Coefficients must be an (n+1, n+1) matrix")
+
+            coeffs = {}
+            for i in range(shape[0]):
+                for j in range(shape[0]):
+                    if i + j < degree + 1:
+                        name = 'c' + str(i) + '_' +str(j)
+                        coeffs[name] = coefficients[i, j]
+            model = modeling.models.Polynomial2D(degree, **coeffs)
```
embray: I'm hoping at some point to come up with a better, easier to use interface for specifying polynomial coefficients.  There are *so many places* where essentially this same code is duplicated and I hope to fix that.  Just something for the "TODO" pile.

nden: @embray I went back and forth between defining the polynomials in terms of individual coefficients or a coefficient matrix. At the end I think it's more convenient if they are initialized as individual coefficients. This allows to set constraints in a straightforward way or quickly check/set values. This was als the opinion of others who I talked with. However I agree the need for something like this comes often so properties/methods on the polynommial class `from_coeffs` and `to_coeffs` would be good.

embray: Yeah, I agree individual coeffs is fine.  Maybe supporting a coefficient matrix is nice too, but optional.  I just meant basically what you wrote in the last sentence, as well as a more flexible template for generating coefficient names (this goes back a while now to a discussion we had about SIP coefficients, for example).  This will be in astropy.modeling at some point...

PR #61, opened 2015-02-10
Round-trip the transform name

### LINKED ISSUE #60: transform.name does not roundtrip
This demonstrates the problem:

```
rot=models.Rotation2D(23, name='rotation')
fa=AsdfFile()
fa.tree={'r': rot}
fa.write_to('rot.asdf')
frot=AsdfFile.read('todel.asdf')
frot.tree['r'].name is None
Out[38]: True
```

`name` is already part of the basic transform schema. Is there a way to add this to the general transform `to_tree` method instead of adding it to every subclass of `TransformType`?

mdboom:
Fix #60.


nden:
Thanks for the quick response.

PR #64, opened 2015-02-25
Change API for setting all array storage

mdboom:
Rather than `exploded=True` which doesn't really scale to other things.


mdboom:
This is part of #51 that I've pulled out so I can proceed on other things.

PR #65, opened 2015-02-25
Use a class to manage structs

mdboom:
Part of #51 that I've pulled out so I can move on with other things.

### DIFF
--- a/pyasdf/util.py
+++ b/pyasdf/util.py
@@ -6,8 +6,11 @@ from __future__ import absolute_import, division, unicode_literals, print_functi
 
 import itertools
 import math
+import struct
 
+from astropy.extern import six
 from astropy.extern.six.moves.urllib import parse as urlparse
+from astropy.extern.six.moves import zip as izip
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


pyasdf/util.py
```diff
@@ -118,3 +121,85 @@ def calculate_padding(content_size, pad_blocks, block_size):
...
+            fields[i] = val
+        return struct.pack(self._fmt, *fields)
+
+    def unpack(self, buff):
+        """
+        Unpack the given binary buffer into the fields.  The result
+        is a dictionary mapping field names to values.
+        """
+        args = struct.unpack_from(self._fmt, buff[:self._size])
+        return dict(izip(self._names, args))
```
embray: Nice little recipe.  I wonder if the explicit `pack` and `unpack` are really necessary--this could just use keyword lookup/assignment built into it to do that automatically.  

Maybe to speed things up don't do struct packing/unpacking on every update/lookup.  But just pack into the struct format in the `__bytes__` method :D

mdboom: I'm not sure I follow.  It doesn't do struct packing/unpacking on every update lookup.  It does one pack/unpack per header (i.e. the whole thing at once).  Update is special because it only updates the parts of the file that really need to be changed, without having to read what's already there.  If the dictionary was maintained in the object, that would be more difficult because then it would have to track what changed, when the calling code already knows that.

embray: My point was basically why have special methods on this object that return a dictionary, when `__getitem__` and `__setitem__` could give it a mapping-like interface on its own, and make the rest of the logic internal.  I don't see why it's necessary to spit out a dict, make changes only to the dict, and the slurp it back in later.

PR #68, opened 2015-02-25
Add to_yaml command

mdboom:
This command takes an ASDF file and makes it 100% YAML by inlining all of the data.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -62,6 +62,9 @@ number of basic operations:
   - ``implode``: Convert an ASDF file in exploded form into a
     self-contained file.
 
+  - ``to_yaml``: Inline all of the data in an ASDF file so that it is
+    pure YAML.
+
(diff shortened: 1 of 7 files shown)


nden:
@mdboom How should I use this? 

The first example below works - writes out the numpy array as yaml array.

```
f.tree={'test': np.array([[1,2],[3,4]])}
f.write_to('test.asdf', auto_inline=10)
```

But this doesn't - still writes the numpy array as binary.

```
aff=models.AffineTransformation2D(matrix=[[1,2],[3,4]])
f.tree={'test': aff}
f.write_to('test.asdf', auto_inline=20)
```

mdboom:
@nden: Indeed, you've found an issue here.  I'll look into it.

mdboom:
Fix in #70, which I'll merge as soon as Travis passes.

PR #71, opened 2015-02-26
Add 3D rotations

nden:


pyasdf/tags/transform/projections.py
```diff
@@ -69,6 +69,74 @@ def assert_equal(cls, a, b):
         assert_array_equal(a.angle, b.angle)
 
 
+class Rotate3DType(TransformType):
+    name = "transform/rotate3d"
```
mdboom: All three of these types have the same YAML tag name, so if the parser encounters that tag, it doesn't know which of the three types to create (it will probably just make the last defined one below).  What's the difference between `RotateNative2Celestial` and `RotateCelestial2Native` anyway?

If we need a separate type to distinguish between the direction, we'll need another YAML tag -- or use the same YAML tag but another key on the object to specify the direction.  (That will require updating the schema, which I just merged without reading this -- sorry).

nden: Duh, right. And yes, it's just the direction. I'll update the schema.

mdboom:
Travis seems to indicate some sort of indentation problem.

nden:
Added all 3D rotations here as well although this didn't go very well. Somehow I didn't notice the conflict the first time. I think it's fine now but I can do a new PR if necessary.

nden:
I am closing this one and will create a new one. The rebase apparently did not work well and locally I am getting tons of errors. It's faster to create a new PR.

PR #75, opened 2015-03-17
Add checksum support

mdboom:
@embray I'd appreciate your comments on this.

Right now it's using MD5 hashes.  Should we consider a different/longer hash?  Any strong argument in favor of storing the hash type in the file (and thus opening this up to multiple hash algorithms in the future)?

This implementation only validates the checksums when an explicit flag is passed to `read`, to avoid unnecessarily reading the entire file.  It always writes checksums to the file, unless the data was memmapped and never accessed at all, since in that case we can assume the existing checksum is still valid.  This does mean that updating the file in place will result in streaming in the entire contents of any data that was accessed (it would be nice to do that only if *writes* were made, but I don't think there's a way to get that info from memmap).  I *think* that might be a good tradeoff to encourage people to store checksums in the file, while avoiding the read-time downsides.  But I could imagine also either (a) adding an option to not update the checksums on writing (and thus store zeros in the checksum field to indicate "no checksum") or (b) make not writing checksums the default (which I fear would limit their use in the wild).  Any thoughts or experience from pyfits?

### DIFF
--- a/pyasdf/block.py
+++ b/pyasdf/block.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division, unicode_literals, print_functi
 
 from collections import namedtuple
 import copy
+import hashlib
 import io
 import os
 import struct
@@ -86,7 +87,8 @@ class BlockManager(object):
                 return x.offset
         self._blocks.sort(key=sorter)
 
-    def read_internal_blocks(self, fd, past_magic=False):
+    def read_internal_blocks(self, fd, past_magic=False,
+                             validate_checksums=False):
         """
         Read all internal blocks present in the file.
 
@@ -95,14 +97,18 @@ class BlockManager(object):
         fd : GenericFile
             The file to read from.
 
-        past_magic: bool, optional
+        past_magic : bool, optional
             If `True`, the file position is immediately after the
             block magic token.  If `False` (default), the file
             position is exactly at the beginning of the block magic
             token.
+
+        validate_checksums : bool, optional
+            If `True`, validate the blocks against their checksums.
         """
         while True:
-            block = Block().read(fd, past_magic=past_magic)
+            block = Block().read(fd, past_magic=past_magic,
+                                 validate_checksum=validate_checksums)
             if block is not None:
                 self.add(block)
             else:
@@ -446,7 +452,7 @@ class Block(object):
         ('allocated_size', 'Q'),
         ('used_size', 'Q'),
         ('data_size', 'Q'),
-        ('checksum', 'Q')
+        ('checksum', '16s')
     ])
 
     def __init__(self, data=None, uri=None, array_storage='internal'):
@@ -457,6 +463,7 @@ class Block(object):
         self._fd = None
         self._offset = None
         self._compression = None
+        self._checksum = None
         self._memmapped = False
 
         self.update_size()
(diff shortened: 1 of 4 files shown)


embray:
I don't see enormous benefits to allowing multiple checksum algorithms--this isn't for cryptographic purposes so md5 is fine.  So I lean YAGNI on that, and it wouldn't be hard to add in a later version if needed.

I'll have to chew a bit though on the question of when/when not to update the checksum.  I totally agree with only *validating* the checksum on read upon request for obvious reasons.  

I've never been totally happy with how pyfits handles checksums, but I haven't made much effort to change it just for fear of breaking how things already work.  Although I will say I think worrying about whether or not the file was mmap'd is a red herring.  Unless the user has completely replaced the data array with a new one, it's next to impossible to know whether the user has changed any bytes in the array regardless whether it was mmap'd or read fully into main memory.  So if they want to maintain the checksum it requires recalculating even if no changes were actually made (unless the only changes were appends).

mdboom:
> Unless the user has completely replaced the data array with a new one, it's next to impossible to know whether the user has changed any bytes in the array regardless whether it was mmap'd or read fully into main memory.

Just to clarify, since wasn't clear.  You're absolutely right.  In the specific case of pyasdf, though, the data isn't even memmapped on first opening of the file, only when the data is actually accessed for the first time.  So in that event, we do know if the user never accessed the data array, that the underlying data couldn't possibly have changed (except perhaps by some other thing that has opened the file, but we don't protect against that in general anyway).  But the minute the array is memmapped, all bets are off, and we have to assume it was changed.

pyasdf/block.py
```diff
@@ -521,6 +528,46 @@ def compression(self, compression):
...
+
+    def _set_checksum(self, checksum):
+        if checksum == b'\0' * 16:
+            self._checksum = None
+        else:
+            self._checksum = checksum
+
+    def _calculate_checksum(self, data):
+        m = hashlib.new('md5')
+        m.update(self.data)
```
embray: I haven't super kept on the details of how arrays are written, but I wonder if this could avoid an extra loop over all the data if instead it were performed in chunks while simultaneously streaming the data block to disk.  Of course this would only work for a random access write since you'd have to go back to write the checksum.  But for typical writes to disk I wonder if there's room for improvement here--compute the checksum on some block of the data while it's still paged in for writing, then be done with it for good.

embray:
>  In the specific case of pyasdf, though, the data isn't even memmapped on first opening of the file, only when the data is actually accessed for the first time. So in that event, we do know if the user never accessed the data array, that the underlying data couldn't possibly have changed (except perhaps by some other thing that has opened the file, but we don't protect against that in general anyway). But the minute the array is memmapped, all bets are off, and we have to assume it was changed.

Right.  It's the same way in PyFITS--if the `.data` attribute as never accessed we can assume the data wasn't modified and can't avoid recalculating the checksum entirely.  I need to double check but I'm pretty sure it is in fact smart enough not to do that (unless there was no checksum in the first place and the user has requested that one be added).

embray:
Yeah, the relevant bit of code is here: https://github.com/astropy/astropy/blob/master/astropy/io/fits/hdu/base.py#L553

The `self._data_loaded` attribute is the one that specifies whether the data has been accessed or not.  It just assumes that if the data was loaded it may have been modified.

pyasdf/block.py
mdboom: Seems plausible -- I'll poke around.

embray: Just a thought--might be annoying to do right now depending how far along it has to pass checksum-related handling.  Maybe in the future it will be cool to add a high-level chunked write routine that can apply any number of filters in the process like checksum computing and compression :D

mdboom: Yeah -- at the moment, we just write the array in one go (which obviously does chunking somewhere along the line underneath), but that doesn't give us a good spot in *our* code to do the checksum calculation on the side.  I might defer that for now in that case...

embray: Seems fine--it was just a half-baked thought.  Possibly worth pursuing in the future.

PR #77, opened 2015-03-18
Fix seek_until so it can find delimiters on block boundaries

mdboom:


pyasdf/generic_io.py
```diff
@@ -531,19 +531,23 @@ def read_until(self, delimiter, delimiter_name=None, include=True):
...
 
     def seek_until(self, delimiter, include=True):
+        last_block = b''
         while True:
             block = self.read_block()
             if not len(block):
                 break
-            index = re.search(delimiter, block)
+            blocks = last_block + block
+            index = re.search(delimiter, blocks)
```
embray: I might be misunderstanding, but couldn't this do something like:
```python
index = re.search(delimiter, block) + len(last_block)
```
and overall avoid the concatentation on the previous line?

embray: Nevermind, I see now why that wouldn't work.  Though I guess it could also do
```python
index = re.search(delimiter, last_block)
if index is None:
    index = re.search(delimiter, block) + len(last_block)
```
and still avoid the concatenate.  I don't know how large these blocks are or if it matters though.

embray: Okay, I see now from the rest of the code where this comes from.  I think maybe the concatenate isn't a big deal, but it might make a noticeable difference if seeking a long ways for the delimiter in question.  I'm curious now to see what difference it makes :)

mdboom: Yeah -- the concatenate is to deal with the case where the regular expression match falls on a block boundary.  Ideally, the Python re module would allow for this (searching through a list of strings or something), but it doesn't.

mdboom:
Note: needs a test still.

pyasdf/generic_io.py
embray: I see now. I don't know what I was thinking--I don't think there's a better way (at least not if you insist on using the re module...) :\

PR #78, opened 2015-03-19
implements 3d rotations

nden:


mdboom:
It looks like Travis is failing because the example (over in asdf-standard) is missing a required "direction" property.

nden:
@mdboom Are you OK with this? It's not new but a copy of #71 to avoid dealing with merge conflicts.

PR #79, opened 2015-03-20
Add commands module to define asdftool entry point

migueldvb:
This commit adds the entry point `asdftool ` in the setup.py script


mdboom:
Thanks.  I think that got moved at some point.

pyasdf/__init__.py
```diff
@@ -14,7 +14,7 @@
 from ._astropy_init import *
 # ----------------------------------------------------------------------------
 
-__all__ = ['AsdfFile', 'Stream', 'open', 'test']
+__all__ = ['AsdfFile', 'Stream', 'open', 'test', 'commands']
 
 from .asdf import AsdfFile
 from .stream import Stream
```
mdboom: We also need a `from . import commands` here.

migueldvb: Thanks, I have added that line.

Issue #82, opened 2015-03-30
Failed test in transform schema


migueldvb:
One of the schema tests fails with a KeyError. This is the end of the traceback output from `python setup.py test`:

```
cls = <class 'pyasdf.tags.transform.projections.Rotate3DType'>
node = {'phi': 12.3, 'psi': -1.2, 'theta': 34}
ctx = <pyasdf.asdf.AsdfFile object at 0x7f7e2f75d390>

    @classmethod
    def from_tree_transform(cls, node, ctx):
        print(node)
>       if node['direction'] == 'native2celestial':
E       KeyError: 'direction'

pyasdf/tags/transform/projections.py:83: KeyError
From file: rotate3d.yaml
=================== 1 failed, 222 passed in 16.35 seconds ====================
```

PR #83, opened 2015-03-30
pulled update to asdf-standard submodule

migueldvb:
The direction key is added to the rotate3d schema. Updating the asdf-standard submodule
fixes issue #82.


mdboom:
Thanks.

PR #84, opened 2015-03-31
selector models

nden:
@mdboom The failing example uses this code (unfortunately you need https://github.com/spacetelescope/gwcs/pull/6 )

```
from gwcs import selector
a = np.zeros((5,6),dtype=np.int32)
mask = selector.SelectorMask(a)
f = AsdfFile()
f.tree['mask'] = mask
f.write_to('mask.asdf')
```

### DIFF
--- a/pyasdf/tags/transform/selector.py
+++ b/pyasdf/tags/transform/selector.py
@@ -16,9 +16,10 @@ from .basic import TransformType
 
 __all__ = ['SelectorMaskType', 'RegionsSelectorType']
 
+
 class SelectorMaskType(TransformType):
     name = "transform/selector_mask"
-    types = [gwcs.selector.SelectorMask]
+    types = [selector.SelectorMask]
 
     @classmethod
     def from_tree_transform(cls, node, ctx):
@@ -31,7 +32,7 @@ class SelectorMaskType(TransformType):
 
     @classmethod
     def to_tree_transform(cls, model, ctx):
-        node = {'mask': model.mask.value}
+        node = {'mask': model.mask}
         return yamlutil.custom_tree_to_tagged_tree(node, ctx)
 
     @classmethod
@@ -43,7 +44,7 @@ class SelectorMaskType(TransformType):
 
 class RegionsSelectorType(TransformType):
     name = "transform/selector_mask"
-    types = [gwcs.selector.RegionsSelector]
+    types = [selector.RegionsSelector]
 
     @classmethod
     def from_tree_transform(cls, node, ctx):
@@ -71,7 +72,3 @@ class RegionsSelectorType(TransformType):
         assert_array_equal(a.outputs, b.outputs)
         assert_array_equal(a.selector, b.selector)
         assert_array_equal(a.undefined_transform_value, b.undefined_transform_value)
-
-
-
-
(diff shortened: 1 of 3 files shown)


mdboom:
Here's what's happening: The tag definition wasn't being imported, so then pyasdf doesn't know how to save the RegionsSelector model.  We would have got a nice traceback, except the __repr__ of the object fails when trying to create the error message (because the model can't convert the parameters to a table, possibly because of the issues with including an int or string array as a parameter that you described).  Merging nden/pyasdf#1 and nden/asdf-standard#1 should help (but then I have some other comments).

pyasdf/tags/transform/selector.py
```diff
@@ -0,0 +1,77 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+import numpy as np
+from numpy.testing import assert_array_equal
+
+
+from gwcs import selector
```
mdboom: This creates a dependency of pyasdf of gwcs, which we probably don't want to do.

There are some ways out here (and I'm Cc'ing @perrygreenfield as he might have some thoughts):

1) Move this type to the `wcs` module in pyasdf, so at least only `wcs` stuff, and not core `transform` stuff in ASDF will depend on `gwcs`.

2) Put the region-handling stuff in astropy.

3) Put the region-handling stuff (i.e. the actual code that handles regions, not just the reading to/from ASDF files) in pyasdf.  That's not as strange as it sounds if these classes exist only to handle ASDF WCS -- and if they are more general, (2) is probably the better choice.

4) Put this ASDF type and schema in the gwcs project, in which case this becomes a test bed for "third-party ad hoc" types in ASDF.

But as it stands, I don't think we want to grow a dependency on gwcs here.  I'm ultimately trying to get the astropy dependency out of pyasdf anyway (except to handle the transform stuff), since it's supposed to be non-astro-specific ultimately.  That should be more doable once some more features are added to astropy-helpers.

nden: I agree this is not the place for these types (I was simply trying to test them).
I guess they can go in astropy eventually but I'm trying to avoid this for the time being (until we have some more experience).
I'm aslo assuming there will be a `gwcs.WCS.toasdf()` method (I may be wrong) in which case 1)  and 2) lead to circular imports, right?
How would 4) work in practice? Is it ready to be tested?

mdboom: For (4), you would just move the contents of this file to the gwcs project (assuming the schemas still live in asdf-standard).  If we want to put the schemas elsewhere, that will require some minor changes to pyasdf.

perrygreenfield:
Is option 4 something the least restrictive and easiest to do for now? If we end up happy, we should then think about where it should go with regard to ASDF. In some sense, I think WCS issues are pretty general to science and not astronomy, so in the long run (if there was interest), this kind of machinery has a natural place outside of astropy. Do you agree that 4 is the easiest and least committing approach?

mdboom:
(4) makes sense to me as well.  @nden: See #85 which should make external extensions to ASDF much easier to implement.  I just need to get some docs on there, and it's ready to merge (probably happen on Monday at this point).

nden:
4) would be my preference as well. Where do the schemas live (in asdf-standard I presume but just want to check).

nden:
These were moved to GWCS - spacetelescope/gwcs#19

PR #86, opened 2015-04-02
update submodule automatically

mdboom:


setup.py
```diff
@@ -8,6 +8,8 @@
 import ah_bootstrap
 from setuptools import setup
 
+ah_bootstrap._check_submodule("asdf-standard")
+
```
nden: Ah, so this is the magic incantation! Is it documented somewhere (e.g. in the instructions for using the package-template)?

mdboom: No, and it's not even a public function.  I plan to file a PR against astropy-helpers/package-template to make this public (since it's useful).

embray: :-1: Let's please not rely on non-public functions.  Automatically initializing submodules is certainly convenient, but please wait until something like the discussion in astropy/astropy-helpers#74 is available.  In the meantime not having this is hardly a showstopper.  You can set up an alias like:
```
clone-all = clone --recursive
```

and use that to initialize and update all submodules when cloning a repository.

mdboom:
Closing.  Whatever approach is ultimately taken, it won't be this one.

PR #89, opened 2015-04-07
fix rotations

nden:
Some rotations parameter names have changed in modeling.


mdboom:
:+1:  Will merge once Travis passes.

PR #90, opened 2015-04-07
Add defragment subcommand

mdboom:


embray:
Nice-- I wonder if this shouldn't have a "friendlier" more "familiar" name.  I suggest "defragment"?

mdboom:
Sure.  `defragment` probably makes more sense.

PR #93, opened 2015-04-09
Use C yaml parser

mdboom:
I wasn't aware that pyyaml doesn't use the C libyaml-based parser/dumper by default.


embray:
Nice--any idea what that gets us in terms of speedups?

mdboom:
The YAML parsing itself now runs in about 25% of the time it did before.  It doesn't have a huge effect on overall parsing time (about 15% faster), since JSON Schema validation and model construction dominate.

PR #95, opened 2015-04-10
Improve performance of schema validation

mdboom:
This will now validate and recurse the tree only once per YAML tag.  It still may recurse multiple times over the same parts of the tree, but I don't see a way around that given how JSON schema works.

This is about a 15% improvement in JSON schema validation.

### DIFF
--- a/pyasdf/__init__.py
+++ b/pyasdf/__init__.py
@@ -27,5 +27,7 @@ from . import commands
 
 from jsonschema import ValidationError
 
+class ValidationError(ValidationError):
+    pass
 
 open = AsdfFile.read
--- a/pyasdf/__init__.py
+++ b/pyasdf/__init__.py
@@ -15,13 +15,17 @@ from ._astropy_init import *
 # ----------------------------------------------------------------------------
 
 __all__ = ['AsdfFile', 'AsdfType', 'AsdfExtension',
-           'Stream', 'open', 'test', 'commands']
+           'Stream', 'open', 'test', 'commands',
+           'ValidationError']
+
 
 from .asdf import AsdfFile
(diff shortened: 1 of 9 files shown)


embray:
How did this work before?  When you say "per YAML tag" do you mean per tagged object?  Why does it need to recurse the tree more than once to validate a tagged object against the schema?

mdboom:
Before, the tree was traversed outside of jsonschema, and jsonschema.validate was called on each tagged YAML node with the schema corresponding to that node.  Now, the traversal happens within jsonschema.validate, which seems to be a bit faster (mainly because the validation setup only needs to happen only once, and all of the errors are collated in a single stream).

You can't deal with the whole thing in a single pass because the "entire" schema in effect is actually annotated as it goes by looking up other subschemas based on the YAML tags.  This is of course different from "vanilla" JSON schema validation.

Take this tree as an example:

```
root: !asdf
   a: !A
      b: !B
         c: value
```

It starts at the top and validates `root` against the `asdf` schema.  But the `asdf` schema knows nothing about the `a` element.  Normally, in JSON schema nothing would happen at all.  However, since in ASDF we want to handle tags in arbitrary places, whether they are mentioned in the parent schema or not, we then recurse to find `a`, look up the `A` schema and then validate on that.  By this scheme, it's possible the subtree at `b` will get validated twice, once with a subschema of `A` and again with the schema of `B`.  It would be nice to do it all in one pass, but at least it's not repeating work (since validating against the subschema of `A` and the `B` schema are different things).

embray:
Awesome, thanks for the explanation.  I feel like this was the sort of problem I was getting stuck on when I was first toying with tag-based validation.

PR #96, opened 2015-04-10
Fix checksum tests for 32-bit machines

migueldvb:
The `assert` statements in test_low_level checksum tests fail for 32-bit computers.

### DIFF
--- a/pyasdf/tests/test_low_level.py
+++ b/pyasdf/tests/test_low_level.py
@@ -698,7 +698,7 @@ def test_checksum(tmpdir):
     tmpdir = str(tmpdir)
     path = os.path.join(tmpdir, 'test.asdf')
 
-    my_array = np.arange(0, 64).reshape((8, 8))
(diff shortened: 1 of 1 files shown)


mdboom:
Ack.  I'd actually rather fix this so that the checksum is the same on all platforms.  We can't have a file format where the checksums aren't portable.  I'll have to fire up a 32-bit virtual machine and play around...

mdboom:
It's very puzzling that this is different -- we are just using the Python stdlib hashlib module.  It really *shouldn't* be different.

migueldvb:
Yes, it's surprising. I will try to get the same checksum on a 32 bit
computer.

migueldvb:
For the test_low_level case the asdf file that contains an (8,8) array produced by arange has a datatype of `int32` on 32-bit and `int64` on 64-bits.  Using  a explicit dtype in the arange function fixes the checksum error.

mdboom:
Ah.  Fantastic.  That makes sense.  Thanks for getting to the bottom of this.

PR #98, opened 2015-04-12
Fix error when building docs witn python 3

migueldvb:
This fixes an encoding error that appears when building the docs with python 3.4. However I think that this may not be portable to other python versions. I have not tried it with python 2 yet.

### DIFF
--- a/docs/sphinxext/example.py
+++ b/docs/sphinxext/example.py
@@ -9,6 +9,7 @@ import os
 import shutil
 import tempfile
 import textwrap
+import codecs
 
 from docutils.parsers.rst import Directive
 from docutils import nodes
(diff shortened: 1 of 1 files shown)


docs/sphinxext/example.py
```diff
@@ -74,7 +75,8 @@ def run(self):
         parts = []
         try:
             code = AsdfFile.read(filename, _get_yaml_content=True)
-            code = '{0}{1}\n'.format(ASDF_MAGIC, version_string) + code.strip()
+            print(code.strip())
```
mdboom: Debugging line left in?

mdboom:
Looks fine (at it's passing on Travis on Python 2.x) so I think this is good to merge after removing the print statement.

migueldvb:
That is great. I have removed the print debugging line ant the Travis build is now in progress.

PR #101, opened 2015-04-15
ASDF-in-FITS embedding

mdboom:
This saves an ASDF file as an extra "ASDF" extension inside of a FITS file.  That ASDF file can refer to binary arrays in other extensions in the FITS file.

### DIFF
--- a/pyasdf/asdf.py
+++ b/pyasdf/asdf.py
@@ -316,53 +316,12 @@ class AsdfFile(versioning.VersionedMixin):
                 int(match.group("micro")))
 
     @classmethod
-    def open(cls, fd, uri=None, mode='r',
-             validate_checksums=False,
-             extensions=None,
-             do_not_fill_defaults=False,
-             _get_yaml_content=False):
-        """
-        Open an existing ASDF file.
-
-        By default, any missing values which have a default value
-        specified in the schema will be added to the tree after
-        loading.  To disable this, set ``do_not_fill_defaults`` to
-        `True`.
-
-        Parameters
-        ----------
-        fd : string or file-like object
-            May be a string ``file`` or ``http`` URI, or a Python
-            file-like object.
-
-        uri : string, optional
-            The URI of the file.  Only required if the URI can not be
-            automatically determined from `fd`.
-
-        mode : string, optional
-            The mode to open the file in.  Must be ``r`` (default) or
-            ``rw``.
-
-        validate_checksums : bool, optional
-            If `True`, validate the blocks against their checksums.
-            Requires reading the entire file, so disabled by default.
-
-        extensions : list of AsdfExtension, optional
-            A list of extensions to the ASDF to support when reading
-            and writing ASDF files.  See `asdftypes.AsdfExtension` for
-            more information.
-
-        do_not_fill_defaults : bool, optional
-            When `True`, don't fill in default values in the tree.
-
-        Returns
-        -------
-        asdffile : AsdfFile
-            The new AsdfFile object.
-        """
+    def _open_impl(cls, self, fd, uri=None, mode='r',
+                   validate_checksums=False,
+                   do_not_fill_defaults=False,
+                   _get_yaml_content=False):
         fd = generic_io.get_file(fd, mode=mode, uri=uri)
 
-        self = cls(extensions=extensions)
         self._fd = fd
 
         try:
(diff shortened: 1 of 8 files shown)


mdboom:
This is ready for review.  It will need a corresponding chapter in the specification to describe exactly how this all fits (no pun intended) together, but it may make sense to kick at this proof of concept first.

We could definitely take a similar approach to HDF5 and allay some of the concerns at the Lorentz Center meeting.  (I don't think trying to put the YAML-like structure directly in HDF5's native structures is going to work for a number of reasons, but shoving some YAML in a blob like this might.)

Cc: @embray, @nden, @perrygreenfield

perrygreenfield:
A concise summary of why YAML doesn't map well to HDF5 may be useful (perhaps for the paper too).

mdboom:
@perrygreenfield: Yes -- I agree.  I'm still sort of poking at it -- once I'm convinced, I'll write something up for the paper.

pyasdf/asdf.py
```diff
@@ -406,6 +365,54 @@ def open(cls, fd, uri=None, mode='r',
 
         return self
 
+    @classmethod
+    def open(cls, fd, uri=None, mode='r',
+             validate_checksums=False,
+             extensions=None,
+             do_not_fill_defaults=False,
+             _get_yaml_content=False):
```
embray: Just wondering, that now that we have open as a front-end to _open_impl, does it need to accept the `_get_yaml_content` argument?  I get the impression that's just for internal use anyways.

pyasdf/fits_embed.py
```diff
@@ -0,0 +1,156 @@
...
+from . import block
+from . import util
+
+
+ASDF_EXTENSION_NAME = 'ASDF'
+
+FITS_SOURCE_PREFIX = 'fits:'
+
+
+class _FitsBlock(object):
```
embray: I'm probably misunderstanding, but is it strictly necessary to have a special kind of "fits block" storage?  In my mind there's nothing special to distinguish a binary block containing a FITS file--and the arrays in the FITS blocks would just be accessed like any other array, with the appropriate offsets.  

That said, I'm not against this either, if you think it makes it clearer what's going on.  There may also be some technical reason I'm missing.

pyasdf/fits_embed.py
```diff
@@ -0,0 +1,156 @@
...
+            base = util.get_array_base(arr)
+            for hdu in self._hdulist:
+                if base is hdu.data:
+                    return _FitsBlock(hdu)
+
+        return super(
+            _EmbeddedBlockManager, self).find_or_create_block_for_array(arr, ctx)
+
+
+class AsdfInFits(asdf.AsdfFile):
```
embray: I haven't tried using it yet, but I *think* I like how this works.  I'm trying to picture how this might be implemented on an `astropy.io.fits` end to support reading an ASDF file out of a FITS HDU.

pyasdf/asdf.py
mdboom: That's a good point.  It's only there to make writing some sorts of tests possible.

pyasdf/fits_embed.py
mdboom: This is to support storing an ASDF file inside of a FITS HDU, and that ASDF content refers to data in the regular FITS HDUs in the same file.

So a FITS block references a FITS HDU by an `EXTNAME,EXTVER` pair.  It could be an offset from the start of the file, I suppose, but this would allow the FITS headers in the file to be edited by eg. `astropy.io.fits` without breaking the ASDF portion.

pyasdf/fits_embed.py
mdboom: You can grab the YAML content by pulling out the data from the `ASDF` HDU.  But this class will be necessary in order to make the data linkages work.

That should be in an example, but it would basically be:

```
with fits.open("test.fits") as hdulist:
    with AsdfInFits.open(hdulist) as ff:
        ff.tree[...]
```

pyasdf/fits_embed.py
embray: Cool, that's a great point in its favor then.

pyasdf/fits_embed.py
embray: I guess the one thing I don't understand in this case is why an entire hdulist is passed to AsdfInFits.open?  Is my understanding incorrect that the ASDF file is stored in a single ASDF extension (of which there could in principle be more than one?  Or do we just not want to go there...?)

mdboom: It needs the whole hdulist so it can get at the binary data in other hdus
referenced from the asdf yaml.

embray: Ah, I think I understand this all better now. So this is done in a way that the FITS HDUs are still accessible through the normal FITS interface, while the ASDF extension references the data in those HDUs directly.  I guess that should have been obvious but for some reason I didn't wrap my head around it earlier.  Very cool.

mdboom: I'll need to write this up as an appendix in the ASDF standard document -- just thought it made sense to get some feedback on this first.

embray: Okay. I'm convinced this is a good approach.

PR #116, opened 2015-04-28
Explain context managers in greater detail

### LINKED ISSUE #107: Context manager clarification
Just out of curiosity, why is the current syntax for the context manager:

```python
ff = AsdfFile(tree)
with ff.write_to("example.asdf"):
    pass
```

Would it not make sense for it to be:

```python
with AsdfFile(tree) as ff:
    ff.write_to("example.asdf")
```

? What is the purpose of using ``write_to`` as a context manager?:

```python
with ff.write_to("example.asdf"):
    pass
```

mdboom:
Fix #107.

### DIFF
--- a/docs/pyasdf/context_managers.rst
+++ b/docs/pyasdf/context_managers.rst
@@ -1,0 +1,43 @@
+.. _context-managers:
+
+Context managers
+================
+
+`~pyasdf.AsdfFile` objects can be used as context managers to be used
+with ``with`` statements to help ensure that the underlying files are
+automatically closed.  There is no single pattern to recommend when
+using `~pyasdf.AsdfFile` objects as context managers, but it will
+instead depend on the context (no pun intended) in which they are
+used.
+
+For example, basic writing and reading of a file::
+
+  >>> import pyasdf
+  >>> tree = {'key': 'value'}
+  >>> with pyasdf.AsdfFile(tree).write_to('test.asdf'):
+  ...     pass
+
+  >>> with pyasdf.AsdfFile.read('test.asdf') as ff:
+  ...     print(ff.tree['key'])
+
+The reason that `~pyasdf.AsdfFile.write_to` and
+`~pyasdf.AsdfFile.read` functions do not automatically close the file
+(unless used with ``with`` statements), is to allow for updating the
+file in-place later on.  For example::
+
+  >>> with pyasdf.AsdfFile.read('test.asdf', 'rw') as ff:
+  ...     ff.tree['key'] = 'new value'
+  ...     ff.update()
+
+You may also want to write the same file, with small modifications, to
+a number of different files on disk, in which case a context manager
+can be used for each write.
+
+  >>> ff = pyasdf.AsdfFile(tree)
+
+  >>> with ff.write_to('a.asdf'):
+  ...     pass
+
+  >>> ff.tree['key'] = 'new value'
+  >>> with ff.write_to('b.asdf'):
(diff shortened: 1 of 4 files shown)


docs/pyasdf/context_managers.rst
```diff
@@ -0,0 +1,43 @@
...
+  ...     print(ff.tree['key'])
+
+The reason that `~pyasdf.AsdfFile.write_to` and
+`~pyasdf.AsdfFile.read` functions do not automatically close the file
+(unless used with ``with`` statements), is to allow for updating the
+file in-place later on.  For example::
+
+  >>> with pyasdf.AsdfFile.read('test.asdf', 'rw') as ff:
+  ...     ff.tree['key'] = 'new value'
+  ...     ff.update()
```
astrofrog: Part of me wonders if this example wouldn't make more sense as ``open`` instead of ``read`` (kind of like ``astropy.io.fits.open). That is:

```
>>> with pyasdf.AsdfFile.open('test.asdf', 'rw') as ff:
  ...     ff.tree['key'] = 'new value'
  ...     ff.update()
```

because I find it confusing to read something with read/write mode.

docs/pyasdf/context_managers.rst
```diff
@@ -0,0 +1,43 @@
...
+can be used for each write.
+
+  >>> ff = pyasdf.AsdfFile(tree)
+
+  >>> with ff.write_to('a.asdf'):
+  ...     pass
+
+  >>> ff.tree['key'] = 'new value'
+  >>> with ff.write_to('b.asdf'):
+  ...     pass
```
astrofrog: Can you clarify here why not just do:

```
>>> ff = pyasdf.AsdfFile(tree)
>>> ff.write_to('a.asdf')
>>> ff.tree['key'] = 'new value'
>>> ff.write_to('b.asdf')
```

astrofrog: (i.e. if it's not used as a context manager, you could auto-close?)

mdboom: Hmm...  The context manager is the AsdfFile object.  Not sure if we can actually detect whether it's in a `with` statement or not.

mdboom:
It probably makes some sense to rename `read` to `open` here.  Use of `open` will always require a context manager, since the file needs to remain open to memmap the data.  For that reason, I don't think it makes sense for `write_to` to not require a context manager, even if we could detect that case, because it creates an asymmetry between `open` and `write_to` that will be difficult to remember.

Obviously there's some confusion around this but it's not yet clear to me how to objectively improve the situation.

mdboom:
Closing in favor of #118.

docs/pyasdf/context_managers.rst
embray: You can--just set a flag in its `__enter__` and unset it in its `__exit__` (if the flag is not set it's not being used in a with block).  I still find the `with ff.write_to(...)` pattern odd, but I think it makes more sense in a context where you can tweak additional details of how the file is written before it's closed.  But I don't see any examples of that usage--that's why I always thought it was like that though.

mdboom: The `write_to` context manager was really for streaming support.  That is, you write out the YAML and some blocks and then you write out a streaming block at the end.  But requiring the `with` block just for that sort of advanced corner case ultimately seemed like a bit much.  So #118 fixes that at the expense of a bit of complexity for doing stream writing.

PR #117, opened 2015-04-29
Fix #111: Workaround reference-counting bug.

### LINKED ISSUE #111: Feature request: auto-close files when using context managers
In the following example:

```python
import numpy as np
from pyasdf import AsdfFile

tree = {'test': np.array([1,2,3])}

f = AsdfFile(tree)
f.set_array_storage(tree['test'], 'inline')
f.write_to('data.asdf')

for i in range(1000):
    with AsdfFile.read('masked.asdf') as f2:
        np.sum(f2.tree['test'])
```

I am running into:

```
OSError: [Errno 24] Too many open files
```

It would be nice if ``read`` could work as a normal context manager and auto-close the file.

mdboom:
If a memmap'd array is used as a mask in a numpy MaskedArray, that
memmap is never free'd.  Using a view on the memmap'd array seems to
workaround this without copying the mask array.

### DIFF
--- a/pyasdf/tags/core/ndarray.py
+++ b/pyasdf/tags/core/ndarray.py
@@ -243,7 +243,12 @@ class NDArrayType(AsdfType):
 
     def _apply_mask(self, array, mask):
         if isinstance(mask, (np.ndarray, NDArrayType)):
-            return ma.array(array, mask=mask)
+            # Use "mask.view()" here so the underlying possibly
+            # memmapped mask array is freed properly when the masked
+            # array goes away.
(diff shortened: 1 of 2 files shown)


astrofrog:
:+1: - works for me!

embray:
That's odd.  Is it a reference cycle or something?

mdboom:
Yes -- it seems to be a reference cycle in `numpy.ma`.  I keep meaning to file a PR for that.  It only happens when the mask array is memmap'd.

PR #118, opened 2015-04-29
Simplify write_to and context manager usage.

### LINKED ISSUE #107: Context manager clarification

### LINKED ISSUE #109: Inconsistent naming of read/write_to?
At the moment, the methods for reading and writing are ``read`` and ``write_to``. These seem inconsistent, and I wonder if it should be either ``read_from/write_to`` or ``read/write``? (the latter would be my preferred choice).

mdboom:
This is an alternative to #116.  Fix #107, fix #109.

This is an attempt to fix up context manager usage at the expense of:

- the streaming API which is now a little less convenient (since it relies on having a raw file handle).  But that's advanced usage anyway

- It's no longer possible to `write_to` and then `update` -- one must `open` and then `update`. 

@astrofrog: Your thoughts?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -39,8 +39,7 @@ transparently:
      'my_array': np.random.rand(8, 8)
    }
    ff = AsdfFile(tree)
-   with ff.write_to("example.asdf"):
-       pass
+   ff.write_to("example.asdf")
 
 .. asdf:: example.asdf
 
--- a/docs/pyasdf/examples.rst
+++ b/docs/pyasdf/examples.rst
@@ -17,14 +17,12 @@ to YAML.  Here we save a dictionary with the key/value pair ``'hello':
    # Make the tree structure, and create a AsdfFile from it.
    tree = {'hello': 'world'}
    ff = AsdfFile(tree)
+   ff.write_to("test.asdf")
 
    # You can also make the AsdfFile first, and modify its tree directly:
    ff = AsdfFile()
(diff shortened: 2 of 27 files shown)


astrofrog:
@mdboom - this looks much more intuitive to me, thanks! I agree that the streaming API is more advanced so it's less important for that to be convenient.

mdboom:
@nden: This is going to be a bit of a breaking API change.  Basically `read` is now `open`, and `write_to` is no longer a context manager (it automatically closes the file and should not be used with `with`).  Let me know if you have any projects that need to be brought over.

embray:
I guess this is already merged, but I think I like this better too.  I've seen other recent cases where people were confused (in a non-ASDF context, I think) about using read/write on objects that are not stream-like objects and don't have the same semantics as, say, file.read and file.write.

embray:
I feel like it might be desirable to salvage something like the streaming API in the future, only even better--allow writing to a file object, but with protections in place to ensure that it only writes to the correct places, etc.  But that can be for another time.  In the meantime I also find this more intuitive :+1:

PR #122, opened 2015-05-01
Fix comment in streaming data example

migueldvb:
Replace reference to `write_to_stream`  in documentation.


mdboom:
Good catch, thanks!

PR #123, opened 2015-05-05
Add table support

mdboom:
Cc: @astrofrog, @embray 

As mentioned in astropy/astropy#3733, and related to spacetelescope/asdf-standard#83.


mdboom:
I've updated this to support recarray columns as discussed in spacetelescope/asdf-standard#83.  This requires astropy/astropy#3759 to be merged first, however.

mdboom:
Note to self: Add a test where one table is explicitly inlined by the writer and another is not (I suspect that doesn't currently work).

PR #126, opened 2015-05-26
Add support for custom validators

mdboom:
And implements the new validators related to ndarray.

[This is part of the ongoing work to support ASDF in jwst_lib.models]


mdboom:
Related to spacetelescope/asdf-standard#85

mdboom:
Incorporates #125.

PR #128, opened 2015-05-26
Support copying of AsdfFile objects

mdboom:
[This is part of the ongoing work to support ASDF in jwst_lib.models]


mdboom:
Incorporates #127

PR #133, opened 2015-05-26
Pass a reader directly to PyYAML

mdboom:
No longer need to read the entire YAML content into memory just to parse it.

### DIFF
--- a/pyasdf/asdf.py
+++ b/pyasdf/asdf.py
@@ -325,22 +325,28 @@ class AsdfFile(versioning.VersionedMixin):
         self._fd = fd
 
         try:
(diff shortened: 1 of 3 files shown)


embray:
Interesting approach.  I like it it!  It's unfortunate that it's necessary at all but oh well.

mdboom:
Yeah -- PyYAML doesn't seem to have an API for saying "parse the first document and stop".  If it did, we could avoid this.

embray:
Might be worth a patch at some point, but in the meantime worth keeping.  I might be able to use this for FITS in some cases too.

PR #138, opened 2015-06-04
Speed improvements

mdboom:


pyasdf/treeutil.py
```diff
@@ -84,13 +132,13 @@ def recurse(tree, seen):
...
-                result = tag_object(tree.tag, result)
+            if hasattr(tree, '_tag'):
+                result = tag_object(tree._tag, result)
         elif isinstance(tree, (list, tuple)):
             new_seen = seen | set([id(tree)])
             result = tree.__class__([recurse(val, new_seen) for val in tree])
-            if isinstance(tree, Tagged):
-                result = tag_object(tree.tag, result)
+            if hasattr(tree, '_tag'):
+                result = tag_object(tree._tag, result)
```
embray: Is this just because the `hasattr` call is faster?

mdboom: Exactly.  Particularly because all of the types actually used are subclasses of `Tagged` so it has to traverse the `__bases__` tree.  Also, duck typing may be handier in the long run.

embray: Okay-- it's too bad there aren't more ways to speed up `isinstance` checks in Python.  I suppose duck-typing is fine if you need to; I can't think of a better way at the moment.

PR #139, opened 2015-06-09
Fix FITS file close handling

mdboom:
The `AsdfInFits` class should not be closing the FITS file, since it's passed in and it doesn't "own" it.

Ideally `HDUList.close()` would close and flush all of it's mmap handles, but it doesn't.  Fixing that is sort of a `astropy.io.fits` issue that is sort of non-trivial to fix (see astropy/astropy#3827), and `pyasdf` shouldn't be trying to patch around that with magic.

Here, the magic is moved to the testing functions only just to get `setup.py test --open-files` to not complain.

### DIFF
--- a/pyasdf/fits_embed.py
+++ b/pyasdf/fits_embed.py
@@ -106,36 +106,14 @@ class AsdfInFits(asdf.AsdfFile):
         self._blocks = _EmbeddedBlockManager(hdulist, self)
         self._hdulist = hdulist
 
-    def __exit__(self, typ, value, traceback):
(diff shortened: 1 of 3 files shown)


embray:
Would it help if I were to get on that?  (I see you already have a solution anyways so maybe it's not urgent.)  I'd still like to fix this though.

mdboom:
Yeah -- it's not urgent, but it would help.  My realization in this PR was essentially that I was papering over something rather than fixing it in the right place...

PR #142, opened 2015-07-06
Minor performance fixes

mdboom:


pyasdf/asdftypes.py
```diff
@@ -108,13 +109,37 @@ def from_yaml_tag(self, tag):
...
     @lru_cache(5)
     def has_hook(self, hook_name):
+        """
+        Returns `True` if the given hook name exists on any of the managed
+        types.
+        """
         for cls in self._all_types:
             if hasattr(cls, hook_name):
                 return True
         return False
```
embray: I honestly haven't kept up with all the details of the internal functioning of this.  What do the hooks do for types?  Is this just a generic way to hook in validation of specific types (though in principle something other than validation as well?)

mdboom: The hooks are run during various phases of saving and loading.  (pre_write, post_write, post_read, etc.)  Certain types (mainly ndarray) need to be able to do things at those phases.  For example, ndarray needs to determine all of the underlying data blocks that are in use through the whole tree, so that unused blocks are garbage collected, duplicate blocks are removed etc.  So it doesn't have much to do with validation -- though we could add pre and post validation hooks someday if there were a need for it that came up.

embray: I see now.  That sounds really useful.  So the reason the `lru_cache` only has 5 entries anyways is because there are only so many hook types?

Is there a list somewhere of all the hook names? Or is it just whatever names get passed at various to points to `AsdfFile.run_hook`?

mdboom: They should probably be documented in one place somewhere, but aren't currently.

embray: It's fine, I was just curious.

PR #143, opened 2015-07-09
Deferred block finding

mdboom:
The current behavior is to find all blocks in the ASDF file by "skipping" along their headers immediately upon opening the file.  While it isn't reading (or even memmapping) all of the data, it is skipping along the file to read the block headers, and if there's a lot of blocks, that overhead could be significant (particularly over a network file system or HTTP with Range headers).

This changes things so the block headers are read lazily when a particular block is requested.  This still isn't random access, since all block headers leading up to a particular block must still be read.  For example, if block 9 of 20 is requested, block headers 0-9 will be read (and cached along the way for later use, so requesting block header 4 afterward wouldn't even hit the disk).

The first block header is always read, since we've usually read that area anyway when finding the end of the YAML tree.

I still plan to implement some sort of an index so that block access can truly be O(1), but I think it will always be the case that we want to support ASDF files without such an index, and that might as well perform as well as we can make it...

Cc: @embray

### DIFF
--- a/pyasdf/asdf.py
+++ b/pyasdf/asdf.py
@@ -280,7 +280,8 @@ class AsdfFile(versioning.VersionedMixin):
 
             - ``inline``: Store the data as YAML inline in the tree.
         """
-        self.blocks[arr].array_storage = array_storage
+        block = self.blocks[arr]
+        self.blocks.set_array_storage(block, array_storage)
 
     def get_array_storage(self, arr):
         """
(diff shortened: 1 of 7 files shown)


embray:
Looks good.  I've been meaning to do the same for pyfits for ages, but keep not getting around to it (probably because it would need deeper changes to old pyfits interfaces).

embray:
As for the index, I think we should still reinstate the PDF-style end-of-file index that I had previously, for exactly this purpose.  Perhaps we could also allow there to be an index at the start of the binary section before the blocks (where the tail index would only be needed for streamed output).  Still I like and agree with your design that such an index isn't needed in the first place.

PR #144, opened 2015-07-09
Add index

mdboom:
This adds a block index to the end of the file.  ~~I'll have to document this over in the spec, but it's basically:

```
�IDX # ASDF BLOCK INDEX
8192
16384
24576
25142
```

where the numbers are the byte index of each block.~~

To the extent that this index format is "versioned" and may change in the future, I think it should follow the version of the file format as a whole (i.e. unlike tag schemas, I don't think this should be individually versioned).

The way the index is read is a little bit heuristic and magical (basically reading chunk-by-chunk backward until finding the index header followed by a list of numbers -- anything else is just rejected and we fall back to the "skipping through" method of finding blocks).  There's still the "it's technically possible, but highly unlikely, that someone's data would look like an index" problem.  I don't know how much I should worry about that.  I thought about using binary values for the indices, but I felt like they would have a greater likelihood of looking like real data, and I don't think there's a real performance advantage to storing these in binary.

Cc: @embray


mdboom:
One more note: The index is automatically not written out if there is a "streamed" (dynamically sized) block at the end of the file.  I'm not sure the two features are compatible.  Probably not a big problem in practice, but you'd have pretty bad performance, for example, with a file with 1000 little blocks followed by a streaming block.

embray:
The reading chunks backwards until a match is found technique is what I used before.  
Rather than having Yet Another Syntax though, why not make the block index itself a YAML document?  Sure it adds a little overhead, but you could require that it ends with '...' and begins with %YAML, which makes finding the thing a little easier, and it's that much less likely to conflict with actual user data.

Since the block index is optional I wonder if we shouldn't add a property to the tree that indicates whether or not it's used at all.  If it's not present then we would know not to go searching for it.  If the tree does indicate that it's present, then it's fairly unambiguous.  This can still be used with streaming because the user can indicate that a block index should be written ahead of time.  That flag can be set in the tree no problem, and the index written once the last binary block is written.

embray:
Indeed, looking at the code for `update` and `write_to` I get the impression that adding a property to the tree indicating presence of a block index should be straightforward to support.

mdboom:
Using YAML makes sense and will allow for expansion later if it ever needs to include more detail.  We may want to be a little more explicit about what it is, perhaps by using a header like:

```
#ASDF BLOCK INDEX
%YAML 1.1
```

(just in case we want to add more YAML "subdocuments" to the format in the future).

I'm not sure about the special key in the tree to indicate the presence of the index.  Updating that value could force the tree to expand or blocks to move etc.  Come to think of it, by changing to YAML and requiring `...` at the end, you could bail out pretty quickly (by reading only a single block at the end of the file) if there happens to be no index.  So I think I'd actually prefer not to have a special key in the tree -- I think that could create expensive moves in order to update it, and if you don't update it, there's potential for it to become out-of-sync with reality and only useful as a "hint".

I sort of see what you're getting at wrt streaming blocks.  The problem is that the calculation of the size of the streaming block changes from "from here to the end of the file" to "from here to the end of the block index".  And it makes one important use case of streaming blocks, opening a file and appending a single row of data and then closing again, more complicated and less efficient.  I think I still lean to saying that the two features aren't compatible rather than having to explain all of the issues that using them both at the same time creates.

embray:
>  We may want to be a little more explicit about what it is, perhaps by using a header like....

Sounds good.

> I think that could create expensive moves in order to update it, and if you don't update it, there's potential for it to become out-of-sync with reality and only useful as a "hint".

I feel like the index itself has the same problem, sort of.  Not so much that you need moves to update it but that the software had better update it lest it become out of sync.  But that's fine, let's forget about any flag for now and see how it goes without it.

mdboom:
I've created a couple of issues for details that I think we can defer for now.  Otherwise I think this is ready to go.

Any further comments, @embray?

embray:
:+1: Actually, I was wondering, have you had a chance to test whether this adds any measurable performance benefits?  Obviously it depends on how the file is structured, how many blocks there are, their size, and the order in which the blocks need to be accessed.

mdboom:
@embray: Very good question, and I should have thought of it before I hastily merged.  The benefits are actually fairly small in my measurements.

I put a timing script here:

https://gist.github.com/mdboom/5769234ab3f96054adc6

(Note, you'll need #149 to run the benchmark).

In the following results the first number is the number of blocks in the file.  The timing is how long it takes to read the last block.  On my Linux laptop with an SSD, I get:

```
10: Indexed: 0.00803089141846, Unindexed: 0.0077440738678
100: Indexed: 0.0669429302216, Unindexed: 0.0636179447174
1000: Indexed: 0.649363040924, Unindexed: 0.639756917953
10000: Indexed: 6.90504193306, Unindexed: 8.13161206245
```

So, until you get to somewhere between 1,000 and 10,000 blocks, there is actually a slight penalty to using the index.

Of course, this all may behave quite differently on a spinning rust drive (which I don't have with me but can test next week) or over a network.

There's probably a safe heuristic here to not output an index when the number of blocks is quite small.

embray:
Well, it's also still useful, for example, in cases where seeking is more expensive, such as HTTP range requests, no?

mdboom:
Yes, I think it will still prove useful -- you have to really push it for it to be useful on an SSD though.  I'm a little surprised by just how much.  I would have expected at least the overhead of making so many system calls to matter.

It seems that when nblocks < 1000, the time spent skipping over all the blocks and the time spent parsing the YAML is roughly the same.  (Tried using Python's JSON parser instead -- it's a little faster, but not much).

mdboom:
On a 5400 RPM disk over USB 2.0, I get:

```
10: Indexed: 0.0230150222778, Unindexed: 0.0208339691162
100: Indexed: 0.0891010761261, Unindexed: 0.126491069794
1000: Indexed: 0.675503015518, Unindexed: 1.12301683426
10000: Indexed: 6.9854631424, Unindexed: 13.2715499401
```

Moral of the story: SSD's are awesome; Indexes do really help on slower media.

embray:
Is that in seconds?  I'm surprised it still takes that long even with the index.  Is that just because the timing script also reads the arrays?

mdboom:
Yes -- in seconds.  The timing script only reads the *last* array.  In the indexed case, most of the time is spent jsonschema validating (63%) and parsing YAML (15%).  With 10000 arrays, the YAML Tree gets rather large (888k), and the index, also being YAML, is fairly large (101k).

PyYAML, even with the C wrappers, has a fairly large "generality penalty".  I've often thought a more focused C wrapper could do much better.

I don't think the `jsonschema` library, good as it is, is super optimized for performance either -- I wonder if there's any low-hanging fruit over there.  Preoptimizing the schemas (e.g. combining layers of `allOf` groups) may help...

PR #146, opened 2015-07-16
Change how HTTP Range connections are handled

mdboom:
To avoid redundant requests to the server, it creates a local copy in a tempfile, and copies over blocks as needed.  This actually greatly simplifies the code and optimizes the network bandwidth at the expense of disk space.


mdboom:
(This is necessarily built on top of #144 -- that should be merged first).

PR #157, opened 2015-07-23
WIP: Fast YAML parser

mdboom:
Based on a discussion of performance in #144, I thought I'd see just how much the YAML parsing time could be improved.

Turns out, a lot.

The new libyaml wrapper here is around 20x faster than the one in PyYAML.  And it still calls into Python to construct every non-scalar object (so is almost as general as PyYAML).

This is all experimental at this point -- not sure I want to pay the price of maintaining something like this, but it's interesting that my hunch of how "Cythonic and Pythonic" PyYAML is really hurts performance.

### DIFF
--- a/pyasdf/setup_package.py
+++ b/pyasdf/setup_package.py
@@ -3,6 +3,9 @@
 
 from __future__ import absolute_import, division, unicode_literals, print_function
 
+
+from distutils.core import Extension
+
 import os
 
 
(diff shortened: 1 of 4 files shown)


embray:
Wow.  First of all, as usual I'm blown away by how much code you turn out in short amounts of time.

Second of all, it's amazing that PyYAML comes with that much overhead (even when using the C parser?)  What is PyYAML doing that has so much overhead that this doesn't do?

mdboom:
The PyYAML wrapper around the C libyaml creates all kinds of Python instances just to pass around events and tokens between different parts of the parser, when an enumerated value would do in most cases.  I think it makes it more general because you can subclass those things and influence their behavior from Python.  But if you don't need to do that and just want the basic data types, it's not very helpful.  The other main difference is that the parsing of scalars (determining what is a string vs. int vs. float etc.) all happens in Python.  (Plus, it's written in some of the most opaque Cython I've ever seen for reasons that are beyond me).

embray:
Yeah, I looked at the CParser class in PyYAML just now and immediately went NOPE and closed it.  But I was able to skim through and understand your code in seconds.

WilliamJamieson:
Closing. This has become very stale.

PR #158, opened 2015-07-23
Software info

mdboom:
Implementation of spacetelescope/asdf-standard#101


embray:
Ah, I had originally envisioned this as something you do at write time only.  But I like this better--this way history can be added at any stage.

PR #160, opened 2015-07-27
WIP: Make astropy a soft dependency.

mdboom:
This makes astropy a soft dependency -- only required for time, units, wcs or transform support.

This is a bit of a hack at the moment and involved copy-and-pasting large chunks of the astropy testing infrastructure into pyasdf.  Once I've settled on what pieces are required, then I'll focus on doing this the "right way" by making those pieces at least usable-without-changes, if not as a separately installable package or something.  @embray's thoughts and advice on how best to handle that would be helpful...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,14 @@ matrix:
         - python: 2.7
           env: SETUP_CMD='build_sphinx -w'
 
+        # Add a couple of configurations without Astropy, since it's a
+        # soft dependency.
+        - python: 2.7
+          env: ASTROPY_VERSION=
+
+        - python: 3.4
+          env: ASTROPY_VERSION=
+
         # Try older numpy versions
         - python: 2.7
           env: NUMPY_VERSION=1.8 SETUP_CMD='test'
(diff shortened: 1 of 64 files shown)


embray:
Oof, that'll be tricky.  Will have to look at this more later and think about it.  I'm increasingly tempted to break up Astropy into multiple packages (most of which would then be vendored...)

mdboom:
Yeah -- the vendored option is the best I can think of as well.  We could make a "astropy-test-infrastructure" (or some better name), and vendor that, installing it to `astropy.tests` as we do now.  Then there's just a few fixes necessary (surprisingly few) to make such a thing usable without the rest of astropy.  We could do that as a proof of concept of vendoring a part of astropy to see how it works, without vendoring *everything*.

embray:
I would be curious to see what it takes to break astropy up as a namespace package, and what problems it causes.  But something like you suggest would be easier to start with...

drdavella:
Closing this since hopefully it will soon be superseded by ongoing work. Also, it's very out of date.

PR #165, opened 2015-07-30
Update WCS to reflect changes in gwcs

mdboom:
Cc: @nden 

The tests could probably be better -- I just grabbed some WCSes from the gwcs tests that seemed to give me good code coverage.  But I made need your help coming up with some WCSes that do unusual things with corner cases etc.


pyasdf/tags/wcs/wcs.py
```diff
@@ -4,99 +4,229 @@
...
     @classmethod
-    def from_tree(cls, node, ctx):
-        return GWCSAxis(**node)
+    def assert_equal(cls, old, new):
+        from ...tests import helpers
+
+        assert old.name == new.name
+        assert len(old.available_frames) == len(new.available_frames)
+        for (old_frame, old_transform), (new_frame, new_transform) in zip(
+                old._pipeline, new._pipeline):
```
mdboom: I had to resort to a private member here.  Any thoughts?

pyasdf/tags/wcs/wcs.py
```diff
@@ -4,99 +4,229 @@
...
-            raise ValueError("Last WCS step must not have a transform defined")
+        return gwcs.WCS(steps, name=name)
 
+    @classmethod
+    def to_tree(cls, gwcs, ctx):
+        frames = gwcs.available_frames
+        steps = []
+        for i in range(len(frames) - 1):
+            transform = gwcs.get_transform(frames[i], frames[i + 1])
+            steps.append(StepType({'frame': frames[i], 'transform': transform}))
```
nden: This will not write to file the entire frame object. The reason is that `available_frames` is a list of the frame names.
To get the frame object use 
```
frame = getattr(gwcs, frame_name)
if frame is None:
    frame = frame_name
```
 Frames can be strings or instances of `CoordinateFrame`. So this will return `None` if a frame is just a string.

pyasdf/tags/wcs/wcs.py
nden: I added a `pipeline` property.

mdboom:
This now implements "reference frame parameters", except for those with more than 2 axes pending some (I think) bugfixes on the gwcs side.

mdboom:
@nden: Do you think this is ready to merge?  (This is the last thing for asdf 1.0)

nden:
@mdboom We discussed the default number of axes with Erik today and agreed on a way to proceed with the default of 2. I need to make changes to the frames in gwcs and then this can be nerged.

mdboom:
@nden: Ready to merge, or do we need further changes here?

nden:
@mdboom I'm working on this today. Sorry it took so long to get back to this.

nden:
As this is written currently there's no way to include distance as an axis in celestial spheres. From a WCS point of view this is probably fine. Taking a minimalistic approach again, we'll add this if/when it is needed. So :+1: to merge.

PR #170, opened 2015-08-21
Rename generated Validator class as ASDFValidator

embray:
I found it super confusing in debugging that the Validator looks to be coming from jsonschema, when in fact it's being created in asdf.schema, and the iter_errors method being overwritten.

This should make the situation a little less confusing.  I was also thinking of adding a custom metaclass to the validator with a more informative `__repr__` telling us exactly what metaschema this validator is for, as well as displaying the `validators` mapping.

### DIFF
--- a/pyasdf/schema.py
+++ b/pyasdf/schema.py
@@ -25,6 +25,9 @@ from . import treeutil
 from . import util
 
 
+YAML_SCHEMA_METASCHEMA_ID = 'http://stsci.edu/schemas/yaml-schema/draft-01'
+
+
 if getattr(yaml, '__with_libyaml__', None):  # pragma: no cover
     _yaml_base_loader = yaml.CSafeLoader
 else:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


pyasdf/schema.py
```diff
@@ -198,61 +201,57 @@ def validate_remove_default(validator, properties, instance, schema):
...
+
+            if ((isinstance(instance, dict) and '$ref' in instance) or
+                    isinstance(instance, reference.Reference)):
+                return
+
+            if _schema is None:
+                tag = getattr(instance, '_tag', None)
+                if tag is not None:
+                    schema_path = self.ctx.tag_to_schema_resolver(tag)
+                    if schema_path != tag:
```
embray: I don't fully understand this--when would it ever be that `schema_path == tag`?

mdboom: If the tag is something that isn't understood by any loaded AsdfTypes.  (In that case, tag_to_schema_resolver returns the tag itself).

embray: Ah, I see, that sort of makes sense.  I'd expect maybe it just returns `None` or some other placeholder value though.

mdboom: It sort of comes from the fact that there is a generic "Resolver" class whose only purpose is to pattern match strings and convert them into other strings.  (The same mechanism is used to map remote URLs to schema files that ship with pyasdf and thus are already on the local disk).  That could be changed, I suppose, but the resolver thing is really flexible...

embray: Okay, I'll look into that.
Mostly I'm just trying to understand so that I can try to implement the generic yaml-schema library.

PR #171, opened 2015-08-21
Validate Python tuples as `type: array` under YAML Schema

embray:
Per the discussion [here](https://github.com/spacetelescope/gwcs/pull/19/files#r37652995), pyyaml will already write out tuples using the YAML sequence type.  However, jsonschema does *not* accept tuples for `type: array`.  In fairness a tuple really isn't like a JavaScript array, but for purposes of *output* validation I feel like it should pass muster (it will never be read back in as a tuple, which is fine).

This PR is built on top of #170, but could just as easily be cleaved apart from it.

Still needs tests.

### DIFF
--- a/pyasdf/schema.py
+++ b/pyasdf/schema.py
@@ -25,6 +25,9 @@ from . import treeutil
 from . import util
 
 
+YAML_SCHEMA_METASCHEMA_ID = 'http://stsci.edu/schemas/yaml-schema/draft-01'
+
+
 if getattr(yaml, '__with_libyaml__', None):  # pragma: no cover
     _yaml_base_loader = yaml.CSafeLoader
 else:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


pyasdf/schema.py
```diff
@@ -198,62 +201,62 @@ def validate_remove_default(validator, properties, instance, schema):
...
-
-    validator.iter_errors = iter_errors
-    return validator
+    meta_schema = load_schema(YAML_SCHEMA_METASCHEMA_ID,
+                              mresolver.default_url_mapping)
+    base_cls = mvalidators.create(meta_schema=meta_schema,
+                                  validators=validators)
+
+    class ASDFValidator(base_cls):
+        DEFAULT_TYPES = base_cls.DEFAULT_TYPES
```
mdboom: We should copy the dictionary here, otherwise this has the side effect of accepting tuples for *all* users of jsonschema in the process.

embray: Oops, right, I meant to do that :)

mdboom:
I think a rebase on master should fix the test failure: The `mmap` changes in astropy.io.fits required changes on pyasdf's end.

mdboom:
I think the Travis failure is due to the mmap changes in astropy.io.fits, which required #173 on pyasdf's end.  I've restarted Travis, but it may require a rebase here in order to pass.

embray:
The tests passed fine--there was just an 0.3% decrease in coverage?

mdboom:
Yeah -- the coverage decrease didn't seem significant, so I ignored it.

PR #179, opened 2015-09-24
Use longest match for resolver

mdboom:


embray:
I don't fully understand the background to this issue.  Could somebody fill me in a bit more?

pyasdf/tests/test_asdftypes.py
```diff
@@ -180,3 +181,27 @@ def url_mapping(self):
...
+        @property
+        def url_mapping(self):
+            return [('http://stsci.edu/schemas/asdf/core/',
+                     'FOOBAR/{url_suffix}')]
+
+    l = extension.AsdfExtensionList(
+        [extension.BuiltinExtension(), FancyComplexExtension()])
+
+    assert l.url_mapping(
+        'http://stsci.edu/schemas/asdf/core/asdf-1.0.0') == 'FOOBAR/asdf-1.0.0'
```
embray: And why would we want an extension to be able to override the path to a core schema?

embray:
@nden -- any idea about this?

nden:
This was to fix a problem with passing more than one custom extensions to`AsdfFile`. It failed to open the file in some cases depending on the order of  the extensions.

embray:
I don't quite understand what the test is supposed to demonstrate though.  It seems to also be demonstrating that a custom extension can override the paths to built-in schemas, which we *probably* don't want.  I also don't understand how the test is demonstrating a "longest match".

pyasdf/tests/test_asdftypes.py
nden: I think it may be simply testing the machinery of the mapping.

embray: Maybe, but I think it exposes a possible loophole we'd probably prefer to close :)

PR #181, opened 2015-12-03
Prevent dumper from splitting new lines.

nden:
This prevents `yaml.SafeDumper` from splitting long lines into multiple lines which in turn causes the pyasdf parser to error in such cases and when the delimiter is a `\n`. This showed up in the context of JWST data models. Here's a minimal example of the problem:

```
from jwst_lib import models
from pyasdf import generic_io, yamlutil
m = models.ImageModel() # meta is not populated
sub={'name': 'FULL', 'xstart': 1, 'ystart': 1, 'xsize': 2014, 'ysize':1023, 'slowaxis': -1, 'fastaxis': 2}
m.meta.subarray = sub
fd = generic_io.get_file('mm', mode='w')
yamlutil.dump_tree(m._instance, fd, m._asdf)
fd.close()
```
Reading the file shows how yaml split the long line.
```
f=open('mm')
f.read()
subarray: {fastaxis: 2, name: FULL, slowaxis: -1, xsize: 2014, xstart: 1, ysize: 1023,\n    ystart: 1}
```

This fixes the immediate problem but is there a better fix. 
@embray @bernie-simon

### DIFF
--- a/pyasdf/yamlutil.py
+++ b/pyasdf/yamlutil.py
@@ -308,4 +308,4 @@ def dump_tree(tree, fd, ctx):
         explicit_start=True, explicit_end=True,
         version=yaml_version,
         allow_unicode=True, encoding='utf-8',
-        tags=tags)
+        tags=tags, width=10000)


embray:
Should get #180 done then re-run this build for this PR.

embray:
> which in turn causes the pyasdf parser to error in such cases and when the delimiter is a \n

Could you demonstrate how this errors?  When I run your example, and follow it up with:
```python
yamlutil.load_tree(open('mm'), m._asdf)
```
it returns
```
{u'meta': {u'date': u'2015-12-09T14:16:23.699', u'subarray': {u'slowaxis': -1, u'name': u'FULL', u'xsize': 2014, u'fastaxis': 2, u'xstart': 1, u'ysize': 1023, u'ystart': 1}}, u'data': {u'datatype': u'float32', u'source': 0, u'shape': [0, 0], u'byteorder': u'little'}, u'err': {u'datatype': u'float32', u'source': 2, u'shape': [0, 0], u'byteorder': u'little'}, u'dq': {u'datatype': u'uint32', u'source': 1, u'shape': [0, 0], u'byteorder': u'little'}}
```

For comparison:
```
In [14]: pprint(sub)
{'fastaxis': 2,
 'name': 'FULL',
 'slowaxis': -1,
 'xsize': 2014,
 'xstart': 1,
 'ysize': 1023,
 'ystart': 1}

In [15]: pprint(yamlutil.load_tree(open('mm'), m._asdf)['meta']['subarray'])
{u'fastaxis': 2,
 u'name': u'FULL',
 u'slowaxis': -1,
 u'xsize': 2014,
 u'xstart': 1,
 u'ysize': 1023,
 u'ystart': 1}
```

So I'm curious to see where it's actually crashing, since there's nothing incorrect about the YAML itself.

embray:
Also when I do
```
m.to_asdf('test.asdf')
pyasdf.open('test.asdf').tree
```
it seems to load fine.

nden:
Yes, this is valid yaml and will load correctly. It fails in the context of JWST data models where it reads from a bytes stream (in some case until a '\n'). Since by default yaml will split a long line into multiple lines of 80 characters, it prevents such file from being read. I'll have to send a failing example offline as it is more involved than this. And it's probably best if this can be fixed in the data models.

embray:
> It fails in the context of JWST data models where it reads from a bytes stream (in some case until a '\n')

Yeah, I'd have to see what you mean.  I don't know what the second "it" in that sentence is--what reads from a bytes stream?

embray:
I think maybe I understand this now.  I think there's a bug in `pyasdf.generic_io._TruncatedReader`.  Still confirming...

embray:
@nden The example code you gave me saved the ASDF tree in a FITS file, which references FITS extensions for the data.  Is that right?  Is it in reading that where the crash occurs?  How do we normally read such files?

nden:
@embray That's right. It crashes when read in: Try this

```
from jwst_lib improt models
models.DataModel(filename)
```

embray:
The original issue here should be fixed by #183.

PR #183, opened 2015-12-10
Fix corner cases in _TruncatedReader--add more tests

embray:
Added a suite of tests for the `_TruncatedReade`r class, testing several tricky edge cases.  Also fixes several edge cases that were broken.  In particular, fixes one case where where a partial read (ex: nbytes=1024) could return content that ends in the middle of the delimiter (for a multi-character delimiter), but then the following read would not return the remainder of the delimiter.

The latter case was (I think) responsible for the bug discussed in https://github.com/spacetelescope/pyasdf/pull/181--in that case the length of the YAML data was such that when the YAML parser read a chunk of the text (in chunk sizes of 1024 bytes, by default), the read was ending inside the delimiter.

The proposed fix in #181 works only in the case of that specific file since it reduced the file size by enough bytes that the first `read()` from within the YAML parser didn't end inside the delimiter.

### DIFF
--- a/pyasdf/generic_io.py
+++ b/pyasdf/generic_io.py
@@ -201,17 +201,30 @@ class _TruncatedReader(object):
         self._delimiter_name = delimiter_name
         self._include = include
         self._initial_content = initial_content
+        self._trailing_content = b''
         self._exception = exception
         self._past_end = False
 
     def read(self, nbytes=None):
         if self._past_end:
-            return b''
+            content = self._trailing_content[:nbytes]
+            if nbytes is None:
+                self._trailing_content = b''
+            else:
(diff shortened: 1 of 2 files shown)


embray:
@nden I think this fixes the specific case that you and Mike (Wolfe) demonstrated, but it would be helpful to have confirmation as well.  Thanks!

embray:
Build failures are due to #180 and so on....

embray:
Rebased on master to get the builds to re-run on the new CI setup.

embray:
Looks like there are still problems with the AppVeyor builds, but I'll work that issue separately and go ahead and merge this.

Issue #186, opened 2016-01-08
Confusion with asdf / pyasdf package name


cdeil:
This package conflicts with another one:

Python package for astro data ASDF format:
- PyPI: https://pypi.python.org/pypi/asdf
- Repo: https://github.com/spacetelescope/pyasdf
- Install: `pyasdf`

Python package for Seismic data ASDF format:
- PyPI: https://pypi.python.org/pypi/pyasdf
- Repo: https://github.com/SeismicData/pyasdf
- Install: `pyasdf`

@krischer @mdboom @embray This is very confusing. Any chance to still change something to avoid the name conflict / simplify this?

embray:
I agree it's confusing.  The question is who has a stronger claim.

embray:
See also #120 of which this is sort of a duplicate.

krischer:
Hi all,

I agree that this is very unfortunate and confusing. For what it is worth we actually tried to rename everything a while ago once we realized the conflict. Some people higher up decided that it is too late to change the name and thus we are stuck with it. I strongly assume the same is true for you as well.

This is all the worse as your format seems to gain some traction and we are also finally done and have the hope and somewhat justified expectation that it will become the new standard for seismology.

I'm sorry - I kind of missed #120. As already noted you have `pyasdf` on readthedocs and `asdf` on pypi. We have the inverse. So we should definitely swap. Whoever ends up getting `asdf` should also rename their `pyasdf` package and install name to make it easier for everyone.

Unfortunately this will still be very confusing :-( What about renaming both packages to `asdf_space` and `asdf_seismo` or something along these lines? We should still retain a hold on the `asdf` and `pyasdf` pypi names to avoid future conflicts.

We should probably also both add short notes to the top of each of our repos/websites that point to the existence of the other format, e.g.

> This is the `Advanced Scientific Data Format` - if you are looking for the `Adaptable Seismic Data Format` please go to http://seismic-data.org

and 

> This is the `Adaptable Seismic Data Format` - if you are looking for the `Advanced Scientific Data Format` please go to http://asdf-standard.readthedocs.org

Looking forward to hear some other opinions/suggestions.

Cheers!

embray:
FWIW our ASDF is for more than astro, so `asdf_space` wouldn't work.  Another problem is that in both our cases the Python package itself is named `pyasdf` so that won't do.

[...]

PR #190, opened 2016-02-24
Rename package to asdf

olebole:
This pull request solves (partially) #186 in the way that the library is renamed from pyasdf to asdf.
I will apply this patch to the Debian version of asdf so that the version 1.0.1-1 there will have this name.

**Please indicate whether this is OK for you or not.** It will be hard to change the name once the package is there, but I want to advertize the format a bit, and enable the users to play with it.


nden:
:+1:

embray:
Wow, thanks @olebole !  This should mostly resolve the issue, other than swapping RTD accounts.

There were a few errors in the docs build--I don't know if those errors were occurring before or not, but I do see some remaining references to "pyasdf" that might have been missed.

We should also rename this repository on GitHub just to be clear.

olebole:
@embray I found the missing piece for the rename; it was the directory `docs/pyasdf`.
For the error in the docs, I have no idea at all. They appear on the lines like
```
.. asdf:: example.asdf
```
which were not renamed by the first commits of this PR. They lead to 
```
/home/travis/build/spacetelescope/pyasdf/docs/index.rst:52: WARNING: Could not parse literal_block as "yaml". highlighting skipped.
```

I first thought of some name conflict and renamed the directive to `asdfdata`; this owever does not help. I could imagine that this happens as well in the original branch (maybe depending on the sphinx release).
Could I request some help here? I am not familar with sphinx extensions.

embray:
It's possible the sphinx extension broke somewhere along the line or isn't being installed properly.  I have a feeling those errors are not due to your changes but I'm not sure (the others were, but that was a simple fix).

embray:
This looks to be related: https://github.com/sphinx-doc/sphinx/issues/2264

@olebole for the "revert" what I really meant was a rebase / squash--no need for an actual revert commit.

embray:
I think after this we can move on--it seems to issue with yaml highlighting is not related to your changes.  Thanks again!

olebole:
OK, I did a squash (hopefully correctly).

olebole:
@embray Could I ask to merge this PR and create a new version then? Otherwise, I would need to create a Debian package that differs quite significantly from the official upstream package.

nden:
Unless there's an objection I will merge this PR later today.

olebole:
Thank you very much! Would you mind to also create a new release and upload to pypi?

nden:
I tagged a new version as 1.0.2. Will upload it as soon as I get permissions on pypi.

olebole:
Now there is a 1.0.2 version registered on pypi, but no files associated with it.

embray:
I think we should be calling this "1.1.0" at the very least considering it completely changed the name of the package :)

embray:
We should also change the name of this GitHub repository to further reduce confusion.

embray:
Similarly I think the dropping of Python 2.6 should be a 1.x release, not a patch release.  Though it's less important for asdf since it doesn't have a wide user base yet.  

But yeah, patch releases should only be to fix bugs--it shouldn't include major interface changes.

nden:
pypi refused to upload the files and we are still trying to sort out the permissions. 
I'll send a note when the files are there, will probably need to tag a new version, so 1.1 may come out as @embray recommends. Changing the name of the repository requires even more coordination with other groups. It will happen eventually, just not right now. 
@embray Do you have permissions for the RTD account and can you assist in switching the accounts?

PR #195, opened 2016-04-15
Added wrapper class for astropy.core.Time, TaggedTime

bernie-simon:
When the asdf validator compares tree nodes against the schema and the
schema contains a tag field, it requres that the node contain the same
tag, or alse the validation fails. Asdf handles this by creating nodes
of type Tagged and its subtypes, all which contain a _tag field that
holds the value matched against the schema. However, only a limited
number of subtypes have been implemented: TaggedDict, TaggedList, and
TaggedString. This commit adds a new type, TaggedTime, to support time
values contained in a tree node.

### DIFF
--- a/asdf/tagged.py
+++ b/asdf/tagged.py
@@ -35,8 +35,8 @@ from __future__ import absolute_import, division, unicode_literals, print_functi
 
 import six
 
+from astropy import time
 from .compat import UserDict, UserList, UserString
-
 
 __all__ = ['tag_object', 'get_tag']
 
@@ -98,6 +98,15 @@ class TaggedString(Tagged, UserString, six.text_type):
                 self._tag == other._tag)
 
 
+class TaggedTime(Tagged, time.Time):
+    """
+    An Astropy time object with a tag attached.
+    """
+    def __new__(cls, instance, tag):
+        self = time.Time.__new__(type(instance), instance)
+        self._tag = tag
+        return self
+
 def tag_object(tag, instance):
     """
     Tag an object by wrapping it in a ``Tagged`` instance.
@@ -109,6 +118,8 @@ def tag_object(tag, instance):
         return TaggedDict(instance, tag)
     elif isinstance(instance, list):
         return TaggedList(instance, tag)
+    elif isinstance(instance, time.Time):
+        return TaggedTime(instance, tag)
     elif isinstance(instance, six.string_types):
         instance = TaggedString(instance)
         instance._tag = tag


asdf/tagged.py
```diff
@@ -109,6 +118,8 @@ def tag_object(tag, instance):
         return TaggedDict(instance, tag)
     elif isinstance(instance, list):
         return TaggedList(instance, tag)
+    elif isinstance(instance, time.Time):
+        return TaggedTime(instance, tag)
```
embray: This is getting to the point where it really ought to be replaced with a dispatch table of some kind--probably as a public interface too so that additional types can be registered.

Since this code was already here though, I'll open a separate issue for that.

asdf/tagged.py
```diff
@@ -98,6 +98,15 @@ def __eq__(self, other):
...
 
 
+class TaggedTime(Tagged, time.Time):
+    """
+    An Astropy time object with a tag attached.
+    """
+    def __new__(cls, instance, tag):
+        self = time.Time.__new__(type(instance), instance)
+        self._tag = tag
+        return self
```
embray: I'm not quite sure what this is doing.  It will never return a `TaggedTime` object per se, because it just creates a plain `Time` object and adds a `._tag` attribute to it.  I admit the implementations of the other `Tagged` types is a bit odd too, but it seems the intent is to act as a wrapper around some type for internal use, rather than a fully functioning instance of that type.

Meanwhile if you wanted to make a `TaggedTime` that is the same as a `Time` object except for having the `._tag` attribute you would call `Time.__new__(cls, instance)`, and I think that should do it.

asdf/tagged.py
```diff
@@ -35,9 +35,9 @@
 
 import six
 
+from astropy import time
```
embray: There's a not-quite-finished pull requested @mdboom started in #160 to make Astropy not a strict dependency of asdf.  It will probably take a bit of work to revive that PR but it would probably be for the best, and this would be at odds with that.

I'm probably missing some context to this PR though, but it should be handled in such a way that doesn't make Astropy a requirement.  (This goes back to my earlier comment about there needing to be an API that *astropy*, when imported, could register additional `Tagged` types...)

bernie-simon:
> This is getting to the point where it really ought to be replaced with a dispatch table of some kind

I don't agree that we're at that point yet. If there were two or three more classes to wrap, I'd say yes.

> I'm probably missing some context to this PR though, but it should be handled in such a way that doesn't make Astropy a requirement.

Since we are wrapping an astropy class, the only way not to make astropy a requirement would be to write a generic object wrapper. I don't think such a wrapper is possible, because the various types being wrapped are too different. 

> if you wanted to make a TaggedTime that is the same as a Time object except for having the ._tag attribute you would call Time.__new__(cls, instance)

The implementation of the astropy.core.time class is a bit arcane and I struggled with writing code to wrap it successfully. One "feature" is that the interface merges cloning an object with creating a new object. If cls is not type(instance), it executes the new object branch, which dies with a error message about the _time attribute not being defined. If cls is type(instance), it executes the clone branch which runs successfully. Adding a _tag field to the cloned object is sufficient to get the validation code to run successfully and was the most economical way of solving the problem.

I'll accept criticism of the way I fixed the validation problem, because I am none too happy with it myself and it's a bit of a hack. But only criticism that is accompanied by working code, as I was unable to resolve problems with the solution that initially seemed more correct.

bernie-simon:
Since this change is needed to support the models code in build 6, I am merging it. Any of the suggested changes can be handled in another branch.

nden:
@bernie-simon Did you actually want to merge this or just close it?

asdf/tagged.py
nden: I haven't looked at the PR but could you place this import statement somewhere in a function or class so that it's not always required.

bernie-simon:
On May 18, 2016, at 2:53 PM, Nadia Dencheva <notifications@github.com> wrote:

> @bernie-simon Did you actually want to merge this or just close it?

Haven�t merged the change yet. I plan to merge it after the Travis tests run. In fact, I rememmber merging it before, but I must be delusional, because the change isn�t there.

nden:
No, you just closed it. You need to reopen the PR if you want tests to run.

asdf/tagged.py
bernie-simon: On May 18, 2016, at 2:55 PM, Nadia Dencheva <notifications@github.com> wrote:

> In asdf/tagged.py:
> 
> > @@ -35,9 +35,9 @@
> >  
> >  import six
> >  
> > +from astropy import time
> 
> I haven't looked at the PR but could you place this import statement somewhere in a function or class so that it's not always required.

The import needs to be at the top level because class TaggedTime subclasses it and it is defined at the top level.

PR #197, opened 2016-05-11
Bring transforms up-to-date with astropy 1.2

nden:
This PR reflects changes made in astropy/modeling v.1.2.

- `Model._custom_inverse` was renamed to `Model._user_inverse` in astropy 1.2.
- Sky rotations were implemented as subclasses of `EulerAngleRotations`. This had the side effect that the model parameters now are the Euler angles `phi`, `theta` ,`psi` and the user parameters which initialize the rotations `lon`, `lat`, and `lon_pole` are returns by properties and are numbers. This is unfortunate but I don't see a simple way around it.

This also removes `numpy 1.6` from the CI matrix.


nden:
The usual sphinx failure, so merging this.

PR #198, opened 2016-05-18
Added wrapper class for astropy.core.Time, TaggedTime

bernie-simon:
When the asdf validator compares tree nodes against the schema and the
schema contains a tag field, it requres that the node contain the same
tag, or alse the validation fails. Asdf handles this by creating nodes
of type Tagged and its subtypes, all which contain a _tag field that
holds the value matched against the schema. However, only a limited
number of subtypes have been implemented: TaggedDict, TaggedList, and
TaggedString. This commit adds a new type, TaggedTime, to support time
values contained in a tree node.


nden:
I'm not sure I understand this. The module docs say that this is intended to create tags for basic types. Time should already be a tag. Could you post a failing example or an example of what you are trying to achieve and also turn it into a test.

bernie-simon:
On May 18, 2016, at 3:05 PM, Nadia Dencheva <notifications@github.com> wrote:

> I'm not sure I understand this. The module docs say that this is intended to create tags for basic types. Time should already be a tag. Could you post a failing example or an example of what you are trying to achieve and also turn it into a test.


Here is the problem. Models validates changes to image metadata in __setattr__ in properties.py. The validation process expects that where there is a tag in the schema the metatdata value that was changed also has a field containing the tag string. In the recent past, this was not handled properly, so any tags in the core schema would cause Models to throw an error. In particular, there is a tag in the core schema for Time, so setting the value of a time field, which happens in model_base.py when an image is updated, would throw an error. 

The fix is to tag the metadata being updated, which means to wrap it in a tagged type, which has the field that the validator is looking for. That code was added to _cast in properties.py and covers all the cases in properties.py where the validator is called.

        tag = schema.get('tag')
        if tag is not None:
            val = tagged.tag_object(tag, val)

So far, so good. But this only fixes Models, there is a new problem in asdf, because it can only handle a limited number of types in tag_object: strings, lists, and dictionaries. The time field is an astropy time object. So I had to write some new code to handle time objects for tag_object:

    elif isinstance(instance, time.Time):
        return TaggedTime(instance, tag)

class TaggedTime(Tagged, time.Time):
    """
    An Astropy time object with a tag attached.
    """
    def __new__(cls, instance, tag):
        self = time.Time.__new__(type(instance), instance)
        self._tag = tag
        return self

Without this fix, tag_object throws an error:

TypeError: Don't know how to tag a <class 'astropy.time.core.Time�>

This does not create a new dependency for asdf, as the astropy time class has serialization code in asdf/asdf/tags/time.py

I hope this clarifies the issue

nden:
@bernie-simon Could you add a Changelog entry under 1.1.0. Also please open an issue to add a test. If it makes sense, the test can be added to the models package with a reference to this PR.

PR #200, opened 2016-06-06
Fixed a memory leak when reading wcs that grew memory to over 10 Gb

bernie-simon:
User reported a that when asdf was used to read an image with multiple
spectra, each with its own wcs, memory use grew until it crashed the
program. The error was traced to the code which read the binary wcs
coefficients stored in the image. That code opened a pseudo file
descriptor which did a seek on the underlying image and called the
numpy frombuffer method to read the coefficients. Each time this
operation was performed memory use grew by 6 Mb, and it was done
almost 1500 times. I believe the cause of the problem is due to an
optimization in the numpy code that allows a slice of an array to use
the same data as the underlying array. The pseudo file descriptor
allocates a large buffer for the image data but never returns it, as
the slice keeps the reference count of the underlying buffer from
going to zero, even though the slice is typically much smaller than
the buffer it is extracted from. The fix replaces the call to
numpy.frombuffer with a call to array.array, which performs the same,
but lacks the optimization in numpy which causes problems in this
case. A test shows the change eliminates the memory leak mentioned,
although there may still be further memory leaks and speed
optimizations in the asdf code.

### DIFF
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -19,6 +19,7 @@ import platform
 import re
 import sys
 import tempfile
+import array as array_class
 
 from os import SEEK_SET, SEEK_CUR, SEEK_END
 
@@ -800,12 +801,12 @@ class MemoryIO(RandomAccessFile):
         fd.seek(tell, 0)
 
     def read_into_array(self, size):
-        result = np.frombuffer(
-            self._fd.getvalue(), np.uint8, size, self._fd.tell())
-        # When creating an array from a buffer, it is read-only.
-        # If we need a read/write array, we have to copy it.
-        if 'w' in self._mode:
-            result = result.copy()
+        buf = self._fd.getvalue()
+        offset = self._fd.tell()
+        if size == -1:
+            result = array_class.array(str("B"), buf[offset:])
+        else:
+            result = array_class.array(str("B"), buf[offset:offset+size])
         self.seek(size, SEEK_CUR)
         return result
 
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -19,7 +19,6 @@ import platform
 import re
 import sys
 import tempfile
-import array as array_class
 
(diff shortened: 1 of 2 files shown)


bernie-simon:
> Good job, Bernie!  That sounds like it would have been a bear to find.

It was difficult, at least for me. For anyone who might need to look for memory leaks in the future, I did not find Pympler, a python package that was designed to track memory leaks, very helpful, mostly because it takes so long to run. The way I located the error was to call 

resource.getrusage(resource.RUSAGE_SELF).ru_maxrss

which returns the amount of memory used by the process and use binary search to track down the spot where memory usage grew dramatically.

nden:
@bernie-simon Thanks for finding the problem! I looked at it for some time and did not figure it out.

I just tested your solution and can confirm that it works for me. However the tests fail. I don't understand the code enough to tell whether the failure is significant or whether we can change the tests.

Perhaps an alternative solution to the problem would be to always make a copy of the array (using the original code). Slices in numpy always return a view in the array not a copy. Is there a reason why we should not use a copy?

bernie-simon:
On Jun 14, 2016, at 10:54 AM, Nadia Dencheva <notifications@github.com> wrote:

> @bernie-simon Thanks for finding the problem! I looked at it for some time and did not figure it out.
> 
> I just tested your solution and can confirm that it works for me. However the tests fail. I don't understand the code enough to tell whether the failure is significant or whether we can change the tests.
> 
> Perhaps an alternative solution to the problem would be to always make a copy of the array (using the original code). Slices in numpy always return a view in the array not a copy. Is there a reason why we should not use a copy?

Which tests failed? I will look at the alternate solution and see if it works.

nden:
The `asdf` `generic_io` tests failed.

bernie-simon:
Because of the failing tests, I'm removing this pull request in place of a new one, coming shortly.

PR #202, opened 2016-06-16
Replace array class code with ndarray class plus copy

bernie-simon:
While my previous fix for the memory leak problem fixed that issue, other code in generic_io.py and its tests expects that the returned object be an ndarray. Since using the frombuffer call followed by a copy to break the reference achieves the same result as using array, I've replaced my previous change to MemoryIO with this new change.

### DIFF
--- a/asdf/tests/test_generic_io.py
+++ b/asdf/tests/test_generic_io.py
@@ -303,7 +303,7 @@ def test_http_connection_range(tree, rhttpserver):
         if len(tree) == 4:
             assert connection[0]._nreads == 0
         else:
-            assert connection[0]._nreads == 6
+            assert connection[0]._nreads == 5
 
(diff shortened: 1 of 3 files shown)


asdf/tests/test_generic_io.py
```diff
@@ -303,7 +303,7 @@ def get_read_fd():
         if len(tree) == 4:
             assert connection[0]._nreads == 0
         else:
-            assert connection[0]._nreads == 6
+            assert connection[0]._nreads == 5
```
nden: I don't think this change is necessary. In any case it causes failures on Travis Ci.

nden:
@bernie-simon Could you please add an entry to the Changelog. 
Other than that and the small inline comment, is this ready to be merged? It'd be good to put a new release out (1.1.1 ? )because the latest release does not work with astropy 1.2.

Issue #209, opened 2016-07-07
version_string is not defined


pllim:
As reported by @hbushouse:

_I, along with at least 1 other user in INS, are running into the following error when attempting to run anything from the jwst package using either the astroconda or jwstb6 environments. I’m running from my Mac laptop (don’t know what the INS user is on). Does anyone have any clue as to what the problem is?_

```
Traceback (most recent call last):
  File "/Users/bushouse/JWST/softw/jwst/scripts/strun", line 14, in <module>
    step = Step.from_cmdline(sys.argv[1:])
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/step.py", line 128, in from_cmdline
    return cmdline.step_from_cmdline(args)
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/cmdline.py", line 269, in step_from_cmdline
    step.run(*positional)
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/step.py", line 322, in run
    self._precache_reference_files(args[0])
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/step.py", line 458, in _precache_reference_files
    model = datamodels.open(input_file)
  File "/Users/bushouse/JWST/softw/jwst/jwst/datamodels/__init__.py", line 195, in open
    return new_class(init, extensions=extensions)
  File "/Users/bushouse/JWST/softw/jwst/jwst/datamodels/image.py", line 34, in __init__
    super(ImageModel, self).__init__(init=init, **kwargs)
  File "/Users/bushouse/JWST/softw/jwst/jwst/datamodels/model_base.py", line 156, in __init__
    pass_invalid_values=self._pass_invalid_values)
  File "/Users/bushouse/JWST/softw/jwst/jwst/datamodels/fits_support.py", line 565, in from_fits
    ff = fits_embed.AsdfInFits.open(hdulist, extensions=extensions)
  File "//anaconda/lib/python2.7/site-packages/asdf/fits_embed.py", line 160, in open
    self = cls(hdulist, uri=uri, extensions=extensions)
  File "//anaconda/lib/python2.7/site-packages/asdf/fits_embed.py", line 146, in __init__
    tree=tree, uri=uri, extensions=extensions)
  File "//anaconda/lib/python2.7/site-packages/asdf/asdf.py", line 87, in __init__
    self.tree = {}
  File "//anaconda/lib/python2.7/site-packages/asdf/asdf.py", line 248, in tree
[...]

PR #210, opened 2016-07-07
Fix version error message

bernie-simon:
This change addresses issue #209 

Asdf throws an error message when it fails to read the version map file. The error message includes the version being searched for in the form of an undefined variable, version_string. I added a line defining this variable so that the exception doesn't throw another exception.
    
Please note that the most likely reason for the version map file not to be found is that the asdf-standard package has not been installed, as that is where the file is found in the distribution.

### DIFF
--- a/asdf/versioning.py
+++ b/asdf/versioning.py
@@ -36,6 +36,7 @@ def get_version_map(version):
                 version_map = yaml.load(
                     fd, Loader=_yaml_base_loader)
         except:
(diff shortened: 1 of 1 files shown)


nden:
@bernie-simon This now fails because `astropy-helpers` as well as `.travis.yml` were updated. It might be easier to create a new PR with the proper change fo rthe version string.

nden:
Superseded by #217.

Issue #207, opened 2016-06-29
asdf-standard not initialized


stscieisenhamer:
When first cloning and doing a 
```
python setup.py install
```
`astropy-helpers` is initialized, but `asdf-standard` is not. I have to manually do
```
git submodule init
git submodule update
```
Is this expected? If not, not clear as to why both are not automatically initialized.

embray:
This is expected, and normal.

There's no way (unforunately) that I know if in git to automatically initialize a submodule when cloning.  This is done in astropy-helpers as a special feature of the ah_bootstrap.py script.  (It has been asked if we could extend this script to also automatically initialize other submodules; I definitely wouldn't object to this but no one has provided a solution).

You can do this a little more concisely with `git submodule update --init`

stscieisenhamer:
Thanks. Didn't know if there was some trick that had been implemented to get this to happen.

embray:
It wouldn't be bad to put in the readme or something at least.

nden:
Closed by #215

PR #215, opened 2016-09-21
Readme

nden:
Address #207


nden:
@stscieisenhamer Is this sufficient?

justincely:
@nden, for clarity I'd recommend putting the submodule init instructions before the standard `python setup.py install` instruction.  For those not used to submodules, it reads like you should run them the other way around.

PR #219, opened 2016-10-10
Handle importing configparser on Python 3.5.2

DavidLutton:
In my case I get an AttributeError for attempting to import ConfigParser, let the except block catch it  
`AttributeError: module 'distutils.config' has no attribute 'ConfigParser'`

Tested with `pip install git+https://github.com/DavidLutton/asdf.git@patch-1` worked in a venv.


nden:
Looking at the error (`from distutils.config`) it looks like you were attempting to install from pypi which has an older version. I believe this has been fixed in master. Try installing from master. A new version will be released in the beginning of November.

PR #221, opened 2016-10-13
Force new blocks to be contiguous

bernie-simon:
The code in asdf assumes a block will be a contiguous block of memory. But recent versions of numpy have used views more heavily and now there are cases when an object instantiated as a bock is not contiguous, which leads to errors when the block is used. The modified code calls numpy.ascontiguousarray to make sure the block memory is contiguous.

This is a revised version of the branch, as several other irrelevant changes were put in the previous version of the branch.

### DIFF
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -757,12 +757,10 @@ class Block(object):
     ])
 
     def __init__(self, data=None, uri=None, array_storage='internal'):
-        if data is None:
-            self._data = data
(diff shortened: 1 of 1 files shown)


bernie-simon:
Closing out an old issue first raised by a problem encountered by Phil.

nden:
@bernie-simon I see the tests failed. Do you know how they failed? 
Also please add a changelog entry.

PR #222, opened 2016-10-17
improve tabular model

nden:
Implement `Tabular.points` as a tuple.
Merge after astropy/astropy#5417 and spacetelescope/asdf-standard#133


nden:
Not sure why coverage decreased by .003 %. Tests are passing so ...merging.

mcara:
_"Not sure why coverage decreased by .003 %."_ Because the number of lines not being tested has increased due to the fact that the case when ``dim`` > 2 is not being tested at all.

nden:
That case was there before. And coverage is supposed to test the "relevant" lines (whatever that means).

Issue #227, opened 2016-11-05
Use of astropy.model.tabular2D


olebole:
asdf unconditionally uses `astropy.model.tabular2D`, which is only introduced in the not yet released version 1.3 of astropy. With the current version 1.2.1 of astropy, the tests will consequently fail:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "asdf/__init__.py", line 37, in <module>
    from .asdf import AsdfFile
  File "asdf/asdf.py", line 15, in <module>
    from . import block
  File "asdf/block.py", line 26, in <module>
    from . import stream
  File "asdf/stream.py", line 6, in <module>
    from .tags.core import ndarray
  File "asdf/tags/__init__.py", line 11, in <module>
    from . import transform
  File "asdf/tags/transform/__init__.py", line 10, in <module>
    from .tabular import *
  File "asdf/tags/transform/tabular.py", line 15, in <module>
    class TabularType(TransformType):
  File "asdf/tags/transform/tabular.py", line 18, in TabularType
    types = [astropy.modeling.models.Tabular2D,
AttributeError: 'module' object has no attribute 'Tabular2D'
```
Please wrap the according code with some version check of astropy, or depend on a minimal version of astropy (once the 1.3 is released). Note however that relying on the newest features will make backporting to older OS versions more difficult (since astropy needs to be backported first).

nden:
Duh! I'll make 1.2.1 fixing this. Thanks for the note!

nden:
@olebole  I've uploaded 1.2.1 to pypi. It runs with older versions of astropy.

olebole:
Thanks, it is now uploaded to Debian as well.

PR #228, opened 2016-11-06
make astropy use conditional to include old versions of astropy

nden:
Allow ASDF to run with older astropy versions by conditionally enabling parts which require newer versions. 
- In astropy 1.2 `Model._custom_inverse` was renamed to `Model._user_inverse`.
- astropy 1.3 will release a `Tabular` model which is already in asdf and is conditionally imported.

Also added testing with two older astropy versions on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,12 @@ matrix:
         - python: 2.7
           env: NUMPY_VERSION=1.10 SETUP_CMD='test'
 
+        # try older astropy versions
+        - python: 3.5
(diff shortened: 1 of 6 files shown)


nden:
This is intended to fix #227.
The failure on Travis is in some `io` tests with older versions of astropy. 
I don't understand why it fails on Travis but it passes locally for me with these versions.

nden:
I have confirmed that tests pass also on Win with astropy 1.2.1.
The Travis failure is probably due to the way the tests are run with older versions of astropy (something that was added in this PR, perhaps incorrectly).

PR #230, opened 2016-12-06
Adding notes to the seismology ASDF

krischer:
Hi guys,

this is a follow up to https://github.com/spacetelescope/asdf/issues/186 and tries to deal with the unfortunate dual use of ASDF as names for data formats. Since we swapped readthedocs domains a couple of weeks back I already had a number of people contact me as they were confused why this suddenly points to a different format.

This PR adds a note to the other ASDF in visible places. Please let me know if there are other important places I should put this.

As a sign of goodwill I put the same note on all of our pages:

* http://seismic-data.org/
* http://asdf-definition.readthedocs.io/en/latest/
* https://github.com/SeismicData/asdf-library
* http://seismicdata.github.io/pyasdf/
* https://github.com/SeismicData/asdf_sextant
* https://github.com/SeismicData/asdf_validate

This, together with this PR, should hopefully minimize confusion among all parties.

Thanks a lot for your help :-)


nden:
@justincely The error in master is now fixed. Could you respond to this and merge when you are ready. Thanks.

justincely:
Thanks for putting this together @krischer!  I see you've put it in asdf-standard too, that was the only other place i'd say it should go.

@nden, after reviewing the rendered PR and looking into it a bit - it looks like GH doesn't support the `.. note::` directive for RST.  If we want a pretty rendering on the README, we'll need to switch to markdown.  [asdf-standard](https://github.com/krischer/asdf-standard/blob/a8946c0cb531264b81b65ed617ee9d685399f73d/README.md) is using md and it looks fine.  Any objections to moving this REAMDE to rst?

nden:
@justincely No objections, feel free to finalize this or suggest other places this should be mentioned. Thanks.

krischer:
Thanks a bunch everyone! Also let me know if any of your users are confused and I can do something about it.

Issue #182, opened 2015-12-10
AsdfInFits API enhancements


embray:
Two possible enhancements I envision to the API when working with FITS-embedded-ASDF:

1) `AsdfInFits.open` currently just accepts an existing `HDUList` object as its first argument.  This means that when reading from a FITS file on disk one has to:

``` python
from pyasdf.fits_embed import AsdfInFits
from astropy.io import fits
asdf = AsdfInFits.open(fits.open('filename.fits'))
```

The two open calls are silly--`AsdfInFits.open` could easily accept any filename or other object accepted by `fits.open`.

2) Relatedly, I think it should be possible to read ASDF directly from a FITS file with `pyasdf.open`.  It's easy to detect that the input is a FITS file (instead of a true ASDF file), and just as easy to detect that it's using the ASDF-embedded-in-FITS convention (which I think should be part of the ASDF Standard if it isn't already, albeit may in an index since it's really more of a usage convention that part of ASDF itself).

PR #241, opened 2017-06-02
Better handling of AsdfInFits

drdavella:
Issue #182 proposed more intelligent handling of FITS files with ASDF extensions. This branch allows ``AsdfInFits.open`` to handle URIs, open file handles, and HDULists, whereas previously it only handled HDULists. Also, ``AsdfFile.open`` now attempts to handle FITS files with ASDF extensions if the given file-like object is not obviously ASDF. Comments on implementation, code style, documentation, etc. welcome.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,7 @@
 - The ``open`` method of ``AsdfInFits`` now accepts URIs and open file handles
   in addition to HDULists. The ``open`` method of ``AsdfFile`` will now try to
   parse the given URI or file handle as ``AsdfInFits`` if it is not obviously a
-  regular ASDF file. [#182]
+  regular ASDF file. [#241]
 
 1.2.1(2016-11-07)
 -----------------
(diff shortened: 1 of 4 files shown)


drdavella:
Whoops, looks like I pushed this a little prematurely since it definitely broke some tests. It's not yet obvious to me what the issue is but it appears to be related to changes to AsdfFile.open. I will resolve on Monday.

perrygreenfield:
Looks reasonable to me.

nden:
@drdavella This needs a change log entry.

nden:
@drdavella I forgot to mention that we have adopted the convention to use the pull request number in `CHANGES.rst`, not the issue number.

Could you add a test case and perhaps an example in the docs. I am currently getting an error trying this with a fits file but my setup may be the problem.

drdavella:
I added an issue (#246) to address the bug you were seeing with missing extensions.

Issue #212, opened 2016-09-19
asdf frames have diverged from the astropy implementation


nden:
More specifically astropy/astropy#5253 implemented `obsgeoloc` and `obsgeovel` as `CartesianRepresentation` objects (used to be `Quantity` objects).
The implementation in asdf needs to be fixed and the two relevant tests in test_wcs.py uncommented.

drdavella:
This change is complete and a pull request will be submitted shortly. One complication is that `CartesianRepresentation` does not define equality, which means that our unit tests do not pass. There is an issue on astropy (https://github.com/astropy/astropy/issues/3182) but it doesn't appear that any work has been done yet.

drdavella:
This was addressed in PR #244 and has been merged.

PR #244, opened 2017-06-08
Asdf frame update

drdavella:
This update addresses issue #212. Astropy updated the representation of the ``obsgeoloc`` and ``obsgeovel`` fields of WCS frames to be ``CartesianRepresentation`` objects where previously they were ``Quantity`` objects. This update brings the ASDF WCS frame implementation in line with   those changes. It relies on a change to the schema of WCS frames in ``asdf-standard`` (currently pending pull request: https://github.com/spacetelescope/asdf-standard/pull/142).

**Note**: The ``CartesianRepresentation`` class from ``astropy`` does not currently define equality. Our unit tests depend on being able to compare an original object with the object resulting from a round-trip tree operation, so they currently will not pass with these changes. I was able to test the changes in the interim by modifying a local checkout of ``astropy``.

### DIFF
--- a/asdf/asdftypes.py
+++ b/asdf/asdftypes.py
@@ -4,8 +4,9 @@
 from __future__ import absolute_import, division, unicode_literals, print_function
 
 import bisect
-import imp
+import importlib
 import warnings
+import re
 
 import six
 
@@ -24,6 +25,9 @@ __all__ = ['format_tag', 'AsdfTypeIndex', 'AsdfType']
 _BASIC_PYTHON_TYPES = set(list(six.string_types) +
                           list(six.integer_types) +
                           [float, list, dict, tuple])
+
+# regex used to parse module name from optional version string
+MODULE_RE = re.compile(r'([a-zA-Z]+)(-(\d.\d.\d))?')
 
 
 def format_tag(organization, standard, version, tag_name):
@@ -325,17 +329,24 @@ class AsdfTypeMeta(type):
 
     @classmethod
     def _has_required_modules(cls, requires):
-        for mod in requires:
-            if mod in cls._import_cache:
-                if not cls._import_cache[mod]:
+        for string in requires:
+            has_module = True
+            match = MODULE_RE.match(string)
+            modname, _, version = match.groups()
+            if modname in cls._import_cache:
+                if not cls._import_cache[modname]:
                     return False
             try:
-                imp.find_module(mod)
+                module = importlib.import_module(modname)
+                if version and hasattr(module, '__version__'):
+                    if module.__version__ < version:
+                        has_module = False
             except ImportError:
-                cls._import_cache[mod] = False
-                return False
-            else:
-                cls._import_cache[mod] = True
+                has_module = False
+            finally:
+                cls._import_cache[modname] = has_module
+                if not has_module:
+                    return False
         return True
 
     @classmethod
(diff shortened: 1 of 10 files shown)


drdavella:
This is failing because a backwards-incompatible change was made to tag schemas in order to support a change in `astropy`. One of the tests on travis is against `astropy-1.2.1` which no longer works. This could be handled in a variety of ways but it seems to suggest that `asdf` standard versions need to be tied to specific versions of `astropy`. Will create a separate issue to address this.

drdavella:
The latest updates also now account for the additional changes to `astropy` frames that were pointed out in issue #259.

asdf/asdftypes.py
```diff
@@ -25,6 +26,9 @@
                           list(six.integer_types) +
                           [float, list, dict, tuple])
 
+# regex used to parse module name from optional version string
+MODULE_RE = re.compile(r'([a-zA-Z]+)(-(\d.\d.\d))?')
```
perrygreenfield: Not being a re expert, but isn't this pattern a little too narrow? What if version numbers get up to 10 or more?

asdf/tags/wcs/wcs.py
```diff
@@ -295,3 +314,47 @@ def assert_equal(cls, old, new):
         assert old.name == new.name
         for old_frame, new_frame in zip(old.frames, new.frames):
             helpers.assert_tree_match(old_frame, new_frame)
+
```
perrygreenfield: I may be missing something in a previous email, but why does ICRS get special treatment for frames here? None of the others have an explicit class definition as far as I understand.

asdf/asdftypes.py
drdavella: Yes, good catch. I was testing for a narrow case but I believe "\d+" for each of those is correct.

asdf/tags/wcs/wcs.py
drdavella: The reason `ICRSCoord` is now defined here is because it is the type of the new `galcen_coord` parameter in galactocentric frames. The `ICRSCoord` is not itself a frame, so it can't be defined using the same schema. It's possible that I missed another existing schema that would subsume `ICRSCoord`, in which case this new code would not be necessary. It's also possible that `ICRSCoord` should be defined elsewhere (e.g. in its own source file under `wcs`), but since it is only being used by galactocentric frames as far as I am aware, this is where I put it for now.

philhodge: Isn't ICRS the coordinate reference system for all JWST coordinates?

nden: @philhodge You are right. But this here is an update to the `Galactocentric`  frame in astropy.coordinates which is now initialized by an object `ICRSCoordinates`. So asdf needs to know how to serizlize it.

asdf/tags/unit/tests/test_quantity.py
```diff
@@ -0,0 +1,38 @@
...
+    quantity = units.Quantity(3.14159, unit=units.kg)
+    buff = helpers.yaml_to_asdf(yaml)
+    with asdf.AsdfFile.open(buff) as ff:
+        assert ff.tree['quantity'] == quantity
+        buff2 = io.BytesIO()
+        ff.write_to(buff2)
+
+    buff2.seek(0)
+    with asdf.AsdfFile.open(buff2) as ff:
+        assert ff.tree['quantity'] == quantity
```
nden: It may be good to add a test with a 2d numpy array value.

asdf/tags/wcs/wcs.py
drdavella: Based on an email from @nden it sounds like we need to have a discussion about coordinate serialization in general in ASDF.

Issue #243, opened 2017-06-07
Fix warnings from pytest


drdavella:
When testing with later versions of pytest (e.g. 3.1.1), numerous warnings appear. Some of these are related to deprecated features in pytest:

```
/tmp/asdf-test-umtx9xt8/lib.linux-x86_64-3.6/asdf/tests/test_schema.py
  yield tests are deprecated, and scheduled to be removed in pytest 4.0
```

Others appear to be real issues with asdf schemas:
```
lib.linux-x86_64-3.6/asdf/tests/test_schema.py::test_all_schema_examples::[30]
  /tmp/asdf-test-umtx9xt8/lib.linux-x86_64-3.6/asdf/asdftypes.py:231: UserWarning: 'tag:stsci.edu:asdf/transform/remap_axes' with version 1.1.0 found in file, but asdf only understands version 1.0.0.
    warnings.warn(warning_string)
```
```
lib.linux-x86_64-3.6/asdf/tests/test_schema.py::test_all_schema_examples::[31]
  /tmp/asdf-test-umtx9xt8/lib.linux-x86_64-3.6/asdf/asdftypes.py:231: UserWarning: 'tag:stsci.edu:asdf/transform/remap_axes' with version 1.1.0 found in file, but asdf only understands version 1.0.0.
    warnings.warn(warning_string)
```

By default astropy 1.3.3 runs tests with a canned version of pytest which appears to be pegged at 2.8.3. Asdf uses astropy's test runner. These warnings do not occur when running with this older version of pytest. However, it is possible to use a more recent version of pytest by running asdf's unit tests with the following command (assuming that a sufficiently up-to-date version of pytest is installed in your environment):
```
$ cd asdf
$ ASTROPY_USE_SYSTEM_PYTEST=1 ./setup.py test
```

- [x] Change unit tests to no longer use yield tests (see [pytest mailing list](https://mail.python.org/pipermail/pytest-dev/2016-October/003907.html) for discussion)
- [x] Fix asdf schema warnings

PR #248, opened 2017-06-13
Mitigate and/or suppress pytest warnings

drdavella:
There were quite a few issues that were causing warnings when running pytests (see issue #243). Some of these were due to deprecated functionality in pytests itself. Others indicate potential issues that need to be addressed, but can be suppressed for the time being.

This addresses issue

### DIFF
--- a/asdf/tests/test_schema.py
+++ b/asdf/tests/test_schema.py
@@ -16,12 +16,10 @@ else:
 from jsonschema import ValidationError
 
 import numpy as np
-
 import pytest
-
 import six
-
 import yaml
+import warnings
 
 from .. import asdf
 from .. import asdftypes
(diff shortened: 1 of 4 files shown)


asdf/tests/test_schema.py
```diff
@@ -105,74 +111,122 @@ def validate_schema(path):
...
+        b = block.Block(np.empty((1024*1024*8), dtype=np.uint8))
+        b._used = True
+        ff.blocks.add(b)
+    b._array_storage = "streamed"
+
+    try:
+        # Ignore warnings that result from examples from schemas that have
+        # versions higher than the current standard version.
+        # TODO: this should no longer be necessary once the library can
+        # actually account for higher versions
```
nden: When updating the schemas to 1.1.0 I left the check for the version as is but I suspect this is not what was intended originally. It would be good to actually know what the intention was.

PR #250, opened 2017-06-15
Make default behavior of AsdfFile.open to fail on ASDF in FITS ...

drdavella:
Originally the intent was to have AsdfFile.open transparently open both
ASDF files and FITS files with ASDF extensions. However, clients (namely
the jwst pipeline) expected AsdfFile.open to fail with a ValueError if any
kind of FITS file was provided. In order to preseve backwards compatibility
for now, the default behavior is to be strict about accepting only ASDF
files but clients can use the try_asdf_in_fits parameter to change this
behavior.


nden:
@drdavella I wrote in the wrong window. I meant this needs to be merged together with the PR in jwst.

PR #257, opened 2017-06-28
Fix changing the compression codec

vmarkovtsev:
See test_recompression() in asdf/tests/test_compression.py

So we should distinguish the input compression type from the output one. Otherwise, loading a file with zlib-ed arrays and saving it with bzip2-ed arrays fails, as the compression is overridden.

This involves careful observation of every `Block.compression` usage and deciding if it should be "input" or "output". Besides, i removed the ambiguous property `is_compressed`. So yeah, this breaks the public API.

Another notice is that opening a zlib file and saving it with `write_to` without specifying `all_array_compression` changed it's behavior. Before: file is saved with the same output compression codec as the input one. After: file is saved without any compression.

### DIFF
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -128,7 +128,7 @@ class BlockManager(object):
             block._array_storage = array_storage
             self.add(block)
             if array_storage == 'streamed':
-                block.compression = None
+                block.output_compression = None
 
     @property
     def blocks(self):
@@ -284,10 +284,12 @@ class BlockManager(object):
             should be after the tree.
         """
         for block in self.internal_blocks:
-            if block.is_compressed:
+            if block.output_compression:
                 block.offset = fd.tell()
                 block.write(fd)
             else:
+                if block.input_compression:
+                    block.update_size()
                 padding = util.calculate_padding(
                     block.size, pad_blocks, fd.block_size)
                 block.allocated = block._size + padding
@@ -550,9 +552,8 @@ class BlockManager(object):
         if all_array_storage:
             self.set_array_storage(block, all_array_storage)
 
-        all_array_compression = getattr(ctx, '_all_array_compression', None)
-        if all_array_compression:
-            block.compression = all_array_compression
+        all_array_compression = getattr(ctx, '_all_array_compression', 'input')
+        block.output_compression = all_array_compression
 
         auto_inline = getattr(ctx, '_auto_inline', None)
         if auto_inline:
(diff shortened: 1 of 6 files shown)


asdf/tests/test_compression.py
```diff
@@ -90,3 +90,18 @@ def test_bzp2(tmpdir):
     tree = _get_large_tree()
 
     _roundtrip(tmpdir, tree, 'bzp2')
+
+
+def test_recompression(tmpdir):
+    tree = _get_large_tree()
+    tmpfile = os.path.join(str(tmpdir), 'test1.asdf')
+    file = asdf.AsdfFile(tree)
```
drdavella: Just a nit but it would be preferred if this variable had a name that does not shadow a builtin.

asdf/compression.py
```diff
@@ -188,6 +188,9 @@ def get_compressed_size(data, compression, block_size=1 << 16):
     compression : str
         The type of compression to use.
 
+    block_size : int, optional
+        The size of blocks (in raw data) to process at a time.
```
drdavella: This comment isn't quite clear to me: I think what it's supposed to say is something like "Input data will be chunked into blocks of this size (in bytes) before compression"

drdavella:
Can you please try to add and/or modify a test so that test coverage does not decrease with this merge?

asdf/block.py
```diff
@@ -823,16 +824,27 @@ def array_storage(self):
...
 
     @property
-    def is_compressed(self):
-        return self._compression is not None
+    def output_compression(self):
+        """
+        The compression codec used to write the block.
+        :return:
+        """
+        return self._output_compression
```
drdavella: In order to account for the case where a file with compression is read, but we do not want to have to explicitly say `all_array_compression` on output, I think this function should look something like this:
```python
def output_compression(self):
    if self._output_compression is None and self._input_compression is not None:
        return self._input_compression
    return self._output_compression
```
I haven't looked at this as closely as you have, but does this make sense?

asdf/compression.py
vmarkovtsev: But... I copy-pasted it from line 169 :-D All right, will rephrase.

asdf/block.py
vmarkovtsev: The thing is, if we change the code this way, setting `all_array_compression=None` will not work - `self._output_compression` will be None and `self._input_compression` will take over. I actually looked into this and decided that it's better to be explicit.

If you think that having the output compression default to the input compression, we could add "input" compression type, which can be the default and mean exactly that.

vmarkovtsev:
@drdavella Done. Had to add some extra tests to make the coverage happy.

I added "input" compression type, which is the default. It means, the same compression for output as in input. Thus `None` all_array_compression forces the decompressed output.

PR #258, opened 2017-06-29
Add lz4 compression type

### LINKED ISSUE #254: Implement LZ4 compression of blobs
Hi! Thanks for this project, it is really cool and we use it to serialize our machine learning models (e.g., https://github.com/src-d/ast2vec). Brilliant idea to combine human-readable YAML and space-efficient blobs.

As I understood from the code and the format specification, ASDF supports zlib and bz2 binary compression. I've recently learned about an exciting compression codec, [lz4](https://github.com/lz4/lz4). As you see from their benchmarks, LZ4 HC is damn slow at compression, yields a similar to zlib compression ratio, but is an order of magnitude faster than zlib during decompression. Our use case is indifferent to the time it takes to generate an ASDF, but is sensitive to the loading lag. I believe, we are not the only ones.

I propose to implement lz4 as the third compression option and include it into some next ASDF specification. I am ready to add it myself, provided by you bless me.

vmarkovtsev:
Fixes #254

Using [python-lz4](https://github.com/python-lz4/python-lz4) here. I had to implement the custom stream wrapper as we read the file in blocks of fixed, small size and need to accumulate them. Unfortunately, `lz4` does not contain the stream implementation yet.

Additional notes:

* The default block size is in bytes now, not in numpy array rows (wtf? I guess it was a bug), and I set it to 4 MiB. Does not change anything for zlib and bzip2 (we use streams), but improves the efficiency of lz4.
* I increased the data size in test_defragment(), otherwise bzip2 does not pass - likely because of the headr overhead.
* I allowed compression objects to miss the `flush()` method.

### DIFF
--- a/asdf/tests/test_compression.py
+++ b/asdf/tests/test_compression.py
@@ -92,8 +92,8 @@ def test_bzp2(tmpdir):
     _roundtrip(tmpdir, tree, 'bzp2')
 
 
-@pytest.mark.skipif(not helpers.has_package('lz4'), reason='lz4 is not installed')
 def test_lz4(tmpdir):
+    pytest.importorskip('lz4')
     tree = _get_large_tree()
 
     _roundtrip(tmpdir, tree, 'lz4')
--- a/asdf/tests/test_compression.py
+++ b/asdf/tests/test_compression.py
@@ -90,3 +90,10 @@ def test_bzp2(tmpdir):
     tree = _get_large_tree()
 
     _roundtrip(tmpdir, tree, 'bzp2')
+
+
+@pytest.mark.skipif(not helpers.has_package('lz4'), reason='lz4 is not installed')
+def test_lz4(tmpdir):
+    tree = _get_large_tree()
+
+    _roundtrip(tmpdir, tree, 'lz4')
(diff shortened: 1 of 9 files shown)


asdf/tests/test_compression.py
```diff
@@ -90,3 +90,9 @@ def test_bzp2(tmpdir):
     tree = _get_large_tree()
 
     _roundtrip(tmpdir, tree, 'bzp2')
+
+
+def test_lz4(tmpdir):
```
drdavella: This introduces `lz4` as a test dependency. The tests are failing because this is not currently accounted for. Can you please update `setup.py` so that this dependency is resolved and the tests pass?

drdavella: This introduces `lz4` as a test dependency. Can you please update `setup.py` so that this dependency is resolved and the test passes?

vmarkovtsev:
@drdavella Done. I had to add lz4 to travis and appveyor configs. As for setup.py, added lz4 as the optional requirement in case "CI" env is defined (Travis). lz4 tests are skipped in case `lz4` pkg is not installed.

PR #263, opened 2017-07-07
Quantity update

drdavella:
Updated Quantity schema to use `ndarray`, along with additional unit tests and other relatively minor changes. This addresses issue #263.


asdf/tags/unit/quantity.py
```diff
@@ -37,4 +37,7 @@ def from_tree(cls, node, ctx):
             return node
 
         unit = UnitType.from_tree(node['unit'], ctx)
-        return Quantity(node['value'], unit=unit)
+        value = node['value']
+        if isinstance(value, NDArrayType):
+            value = value._make_array()
```
nden: I see now what you mean - `node['value']` is the memory mapped array.

Issue #252, opened 2017-06-16
Astropy version awareness


drdavella:
Issue #249 addressed the need to support backwards compatibility for tag schemas. These schemas will change over time for various reasons and it will probably be desirable to maintain some form of backwards compatibility.

Recently a change to `astropy` required changes to several of our schemas (this is discussed in issue #212 and is being addressed in PR #244). Since this change is tied to a particular implementation of `astropy`, once it is implemented, using previous versions of `astropy` will not work unless some form of backwards compatibility is provided. However it is not enough to simply provide backwards compatibility for multiple tag schemas: it is necessary to tie particular tag schema versions, and therefore particular versions of the ASDF standard, to specific `astropy` versions.

This makes an already complicated versioning schema even more complicated. It may be sufficient to have a high-level check in `asdf` that makes sure that the installed version of `astropy` is compatible with the currently supported `asdf` standard.

drdavella:
I believe this issue has been sufficiently addressed by PR #244 and #272.

Issue #249, opened 2017-06-14
Support backwards compatibility for tag schemas


drdavella:
There doesn't appear to be a mechanism to support backwards compatibility for tag schemas. For example, although different versions of the standard are allowed, the versions do not appear to have any effect on the tag schema versions that are used.

For example, maybe hypothetical version 1.5.0 of the standard provides support for version 1.1.0 of ``wcs/frame``, and version 1.6.0 of the standard provides support for version 2.0.0 of ``wcs/frame``. Even if the library provides support for both versions of the standard, there appears to be no way for version 1.1.0 of ``wcs/frame`` to be used (this example ignores the question of how schema and standard versions should be correlated).

A possibly relevant observation is that the library's ``version_map`` only appears to be used when writing new ``asdf`` files, but not when reading or creating new ``AsdfFile`` objects. It will probably be necessary to make use of this information in order to support backwards compatibility for tag schemas.

drdavella:
This has been resolved by PR #272.

PR #272, opened 2017-07-11
WIP: Schema backwards compatibility

drdavella:
As it stands, this is a fairly naive implementation of backwards compatibility for `CelestialFrame`s. More than anything it's meant to demonstrate how backwards compatibility might be implemented for future tag classes where it is actually required. It's also meant to expose functionality that needs to be added/modified in order to support backwards compatibility. I will continue to update this PR with improvements until it seems mature enough to merge.

This is an attempt to address issue #249, and also may be related to #252.

### DIFF
--- a/asdf/asdftypes.py
+++ b/asdf/asdftypes.py
@@ -391,16 +391,13 @@ class ExtensionTypeMeta(type):
                 raise TypeError("name must be string or list")
 
         if hasattr(cls, 'supported_versions'):
-            if isinstance(cls.supported_versions, six.string_types):
-                cls.supported_versions = set(cls.supported_versions)
-            elif isinstance(cls.supported_versions, tuple):
-                cls.supported_versions = \
-                    set(version_to_string(cls.supported_versions))
-            elif isinstance(cls.supported_versions, (list, set)):
-                supported_versions = set()
-                for ver in cls.supported_versions:
-                    supported_versions.add(version_to_string(ver))
-                cls.supported_versions = supported_versions
+            if not isinstance(cls.supported_versions, (list, set)):
+                raise TypeError(
+                    "supported_versions attribute must be list or set")
+            supported_versions = set()
+            for ver in cls.supported_versions:
+                supported_versions.add(version_to_string(ver))
+            cls.supported_versions = supported_versions
 
         return cls
 
--- a/asdf/asdftypes.py
+++ b/asdf/asdftypes.py
@@ -390,6 +390,18 @@ class ExtensionTypeMeta(type):
             elif cls.name is not None:
                 raise TypeError("name must be string or list")
 
+        if hasattr(cls, 'supported_versions'):
+            if isinstance(cls.supported_versions, six.string_types):
+                cls.supported_versions = set(cls.supported_versions)
+            elif isinstance(cls.supported_versions, tuple):
+                cls.supported_versions = \
+                    set(version_to_string(cls.supported_versions))
+            elif isinstance(cls.supported_versions, (list, set)):
+                supported_versions = set()
+                for ver in cls.supported_versions:
+                    supported_versions.add(version_to_string(ver))
+                cls.supported_versions = supported_versions
+
         return cls
 
 
@@ -452,6 +464,7 @@ class ExtensionType(object):
     organization = 'stsci.edu'
     standard = 'asdf'
     version = (1, 0, 0)
+    supported_versions = set()
     types = []
     handle_dynamic_subclasses = False
     validators = {}
@@ -508,7 +521,9 @@ class ExtensionType(object):
         Indicates whether this tag class knows how to convert the given version
         of the associated schema.
         """
-        # WIP: this will be updated in the near future
+        if cls.supported_versions:
+            if version_to_string(version) not in cls.supported_versions:
+                return False
         return True
 
 
--- a/asdf/asdftypes.py
+++ b/asdf/asdftypes.py
@@ -325,10 +325,9 @@ def _from_tree_tagged_missing_requirements(cls, tree, ctx):
     return tree
 
 
-class AsdfTypeMeta(type):
+class ExtensionTypeMeta(type):
     """
-    Keeps track of `AsdfType` subclasses that are created, and stores
-    them in `AsdfTypeIndex`.
+    Custom class constructor for extension types.
     """
     _import_cache = {}
 
@@ -380,7 +379,7 @@ class AsdfTypeMeta(type):
                 new_types.append(typ)
             attrs['types'] = new_types
 
-        cls = super(AsdfTypeMeta, mcls).__new__(mcls, name, bases, attrs)
+        cls = super(ExtensionTypeMeta, mcls).__new__(mcls, name, bases, attrs)
 
         if hasattr(cls, 'name'):
             if isinstance(cls.name, six.string_types):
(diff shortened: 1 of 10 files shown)


drdavella:
Here's a fairly unstructured list of thoughts, questions, considerations, and the like (mostly for my own benefit):
* Before these changes, no tag class that implemented v1.0.0 of `CelestialFrame` was available. If a file was read that had an older schema, a warning was issued but then v1.1.0 of `CelestialFrame` was used as the tag. This didn't work since the v1.1.0 class doesn't know anything about the older data type or how to convert data stored in v1.0.0 to the newest `CelestialFrame` implementation. It was necessary to make ```my_version < supported_version``` raise an exception instead of a warning. It was then necessary to implement a class associated with `CelestialFrame` that used `1.0.0` as its version.

perrygreenfield:
I think I will make comments in reverse order.

- I agree that anything that prevents the proliferation of classes is probably a very good thing. So it sounds like the proposed solution of supported versions is better. But this may also be addressed by the comment to the middle item, next.

- Would it be a problem to make backward compatibility a necessary requirement for any particular tag? New versions may implement new features, attributes, etc, but a library implementation should always be able to handle *all* older versions. If that is required, are the points made a problem still? For example, a version of the library that supports version Y runs into an older version X of the schema. It can handle that. Suppose an older version of the library that could handle version X but reads a file with version Y of the schema; what happens? Good question. For it to be able to handle minor differences that don't require changes in implementation (are there examples of this? If is a new version, one thinks some change in implementation is needed). In this case I would propose a warning ("you really should read this file with a newer version of the library"), and reading it as a basic Python data structure. But I may be missing a key element in this so give me an example where these approaches don't work.

- I'm not clear on the first item. OK, it previously didn't work. I'm unclear of what the final situation is though. So the latest version of the library has a CelestialFrame schema/version incompatible with the older schema. Is it possible to trap the exception and check for the special case ("I wonder if it uses the older form for the Frame, and I will try to construct it that way"). Yes, this violates the schema formalism on some level (but practicality beats purity sometimes). We would strive to avoid situations like this in the future if possible.  Again, I may be missing something.

drdavella:
I will respond in pseudorandom order:
* The point I failed to make in the first bullet is that ASDF did not recognize that it (at that time) did not support an older version of the `CelestialFrame` schema. It simply assumed that it could map the older, unsupported version to the newer tag class, which was not true in this case. The first step I took was to modify the logic that ASDF uses to recognize compatible versions. This caused a hard failure when reading the older schema, which is what I wanted. I was then able to implement a new tag class specifically for the older schema, so now everything "works".
* I think I agree that backwards compatibility, at least within the same major version series, should be a requirement for all tag classes. The question at issue, however, is how exactly to implement that backwards compatibility. So far in this branch, backwards compatibility is supported by providing multiple tag classes, with one corresponding to each supported version of the schema. However, I've argued, and it sounds like you agree, that this might not be the best solution in the long term. Again, the point I'm trying to make is that ASDF doesn't have a general mechanism or coherent strategy for handling backwards compatibility. There are some pieces in place, and I'm trying to sort out the rest in this PR. The `CelestialFrame` case is serving as a useful example.
* I might be saying the same thing over and over, but we need to figure out how ASDF **knows** that it supports a particular schema version. It's one thing to say that it **should** support backwards compatibility, but there still needs to be a mechanism in place for the library to recognize whether or not it supports a particular version.
* What happens if my library supports v1.1.0 of a particular schema, but I read a file with v1.2.0 of that schema? Should that cause a failure? A warning? I'm not really sure.

perrygreenfield:
> I think I agree that backwards compatibility, at least within the same major version series, should be a requirement for all tag classes. The question at issue, however, is how exactly to implement that backwards compatibility. So far in this branch, backwards compatibility is supported by providing multiple tag classes, with one corresponding to each supported version of the schema. However, I've argued, and it sounds like you agree, that this might not be the best solution in the long term. Again, the point I'm trying to make is that ASDF doesn't have a general mechanism or coherent strategy for handling backwards compatibility. There are some pieces in place, and I'm trying to sort out the rest in this PR. The CelestialFrame case is serving as a useful example.

Yes, I agree that this isn't the best long-term approach and that we will need to do work to avoid multiple tag classes.

> I might be saying the same thing over and over, but we need to figure out how ASDF knows that it supports a particular schema version. It's one thing to say that it should support backwards compatibility, but there still needs to be a mechanism in place for the library to recognize whether or not it supports a particular version.
> What happens if my library supports v1.1.0 of a particular schema, but I read a file with v1.2.0 of that schema? Should that cause a failure? A warning? I'm not really sure.

I think we should never have a hard failure (i.e., untrapped exception) reading an ASDF file so long as the yaml is well formed. In such a case we issue a warning (at most) and read the unhandled schema tag as a basic python data structure. If backward compatibility is always adhered to, then the main issue is how to handle versions ahead of what the library supports (right?). What I'm suggesting (I don't know if it is practical) is to try reading it with the latest schema that it supports, and if it works, fine (warning about the version), and if not, read it in as a python structure (though if nested items have supported tags, e.g., numpy arrays, they do get handled).

I'm not even sure I would agree that a major version change breaks backward compatibility. The basic principle here is that a tag means something that has an invariant aspect to it. Future versions may change how it may be represented, or enhanced. But if we can't represent the old version in the new (by conversion), it's a different entity and should be named differently (and not just by version numbers). I'm open to see good counterexamples (these may have to come from a wider circle)

drdavella:
> I think we should never have a hard failure (i.e., untrapped exception) reading an ASDF file so long as the yaml is well formed. In such a case we issue a warning (at most) and read the unhandled schema tag as a basic python data structure.

After some more investigation, it actually looks like this much is already supported (assuming it does not try and fail to map it to another version of the same type). I will add some unit tests to confirm.

> What I'm suggesting (I don't know if it is practical) is to try reading it with the latest schema that it supports, and if it works, fine (warning about the version), and if not, read it in as a python structure (though if nested items have supported tags, e.g., numpy arrays, they do get handled).

This seems like a good strategy in all cases. I'll do some more digging to see whether this works yet or not (I suspect it doesn't yet but I might be wrong).

Still tbd is the best way to accommodate explicit implementations of multiple versions of the same schema. I'll play around more with the ideas proposed above.

drdavella:
The latest commit addresses issue #276 by updating the extension type hierarchy. This will be a prerequisite for a clean and general way of determining schema versions that are supported by a particular tag class.

drdavella:
Extension classes can now define a `supported_versions` field. If this field is not explicitly defined, the tag class is assumed to be compatible with any schema version that it encounters. If the field is provided, it will be used to determine whether the class can be used to convert a given schema version. If a schema version is encountered that is not supported, the schema will simply be converted to a raw Python object representing the schema data.

asdf/asdftypes.py
```diff
@@ -417,6 +454,11 @@ class AsdfType(object):
     version : 3-tuple of int
         The version of the standard the type is defined in.
 
+    supported_versions : set
```
drdavella: The idea of supporting version ranges occurred to me as well, and I think it's probably the right thing to do. I also think it makes sense to address it in a separate issue after getting this PR merged.

Likewise, I think it makes sense to allow for indefinite ranges. None of this should be too difficult to support; it will just require some modifications to the way these things are represented internally. Again, I'd like to make a separate PR for these.

perrygreenfield: I agree it should be in a different PR

PR #284, opened 2017-07-26
Implement AsdfVersion class

### LINKED ISSUE #279: Implement AsdfVersion type
Currently versions can be represented as either tuples or strings:
```python
version = (1, 0, 0)
version = '1.1.0'
```

However, this causes issues when performing version comparisons, so generally all versions need to be converted to one format or the other. Instead, it might be helpful to have an `AsdfVersion` class that allows for transparent comparisons of all different version formats:

```python
version = AsdfVersion('1.0.0')
assert version == AsdfVersion((1,0,0))
assert version == AsdfVersion(1,0,0)
assert version == '1.0.0'
assert version == (1,0,0)
```

### LINKED ISSUE #280: Allow ranges for supported versions
ASDF tag types now provide a `supported_versions` attribute for explicitly declaring which schema versions are supported (and implicitly declaring those that are not). Right now a list of versions needs to be provided. However, it would probably be much more useful if these versions could be expressed as a range (e.g. `2.0.0-2.1.3`), or potentially as a wildcard (e.g. `2.*` or some such). It may also be useful to provide indefinite ranges (e.g. `1.3.3+`) to express that every version at or above a certain point is supported.

This work can probably go hand in hand with issue #279.

### LINKED ISSUE #282: Can we add semver as a real dependency?
Right now we package the `semver` module as an extern. However, it is available on [pypi](https://pypi.python.org/pypi/semver), so it seems like there's no reason we can't add it as a requirement to `setup.py` and use it that way. But maybe there's some other reason I'm not aware of for avoiding that.

### LINKED ISSUE #283: Replace semver with another package
There are several open issues related to versioning right now: issues #279 and #280 request new features for the versioning mechanism. After doing some research, I found a Python package called `semantic_version` that I think provides a better solution than the current `semver` package. I would like to add this package as a dependency from [pypi](https://pypi.python.org/pypi/semantic_version/) rather than including it as an extern, like the current `semver` module. 

This would make #282 obsolete, although the question still stands.

drdavella:
This PR directly resolves #279 by implementing a new type `AsdfVersion` that is used to represent all schema version information. It also implements an `AsdfSpec` class that will be used to represent ranges of versions, which was requested in #280.

This also directly resolves #283, in that it replaces the extern package `semver` with `semantic_version`, which is now a dependency that is expressed in `setup.py`. It indirectly closes #282 since we no longer want to add `semver` as a real dependency.

I intend to directly address #280 before closing this PR.

[edit]
Now fixes #280 too.


drdavella:
Remember to update `CHANGES.rst`

drdavella:
This now resolves #280 as well (as requested by @perrygreenfield) and allows supported versions to be expressed in terms of unbounded ranges (e.g. `>=X.Y.Z`).

drdavella:
Since this revision removes the extern package `semver`, is it also okay to remove the license file `licenses/LICENSE-semver.rst`? It seems like the answer should be yes but I'm not really sure.

PR #294, opened 2017-08-03
Fix schema mismatch warnings

### LINKED ISSUE #277: Address root cause of warnings in test_schema.py
There are a lot of warnings that occur when running `test_schema.py`:
> lib.linux-x86_64-3.6/asdf/tests/test_schema.py::test_schema_example[asdf/transform/subtract-1.1.0.yaml-example]
  /tmp/asdf-test-5u03i9dz/lib.linux-x86_64-3.6/asdf/asdftypes.py:260: UserWarning: 'tag:stsci.edu:asdf/transform/subtract' with version 1.1.0 found in file, but asdf only supports version 1.0.0
    warnings.warn(warning_string)
  /tmp/asdf-test-5u03i9dz/lib.linux-x86_64-3.6/asdf/asdftypes.py:238: UserWarning: 'tag:stsci.edu:asdf/transform/generic' with version 1.1.0 found in file, but asdf only supports version 1.0.0
    warnings.warn(warning_string)

They are currently being suppressed in `test_schema_example` in `test_schema.py`. We should address the underlying cause, especially now that the version compatibility logic is being updated in PR #272.

drdavella:
This resolves #277. It technically doesn't address the root cause of the warnings because the warnings are perfectly valid and correct. However, we expect them in certain unit tests, so they should be suppressed. This PR adds a user-facing mechanism for suppressing warnings due to mismatched schema versions, and uses that mechanism to suppress the warnings that we expect in the schema validation unit tests.

Note: this PR is based on #284 and should not be merged until after that has been merged.


nden:
@drdavella In general this and #284 look fine. 👍  to merge soon and see if any problems show up in the results of the nightly JWST tests.

PR #295, opened 2017-08-03
Issue warnings for unrecognized tags

### LINKED ISSUE #13: Warn when encountering unknown tags
Not sure if we do this by default, but it would be useful.

drdavella:
This fixes #13. It also fixes unit tests that were producing warnings once this feature was added.

It's possible that later we will want to add an option to suppress this warning.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,6 +46,7 @@
 - When possible, display name of ASDF file that caused version mismatch
   warning. [#306]
 
+- Issue a warning when an unrecognized tag is encountered. [#295]
 
 1.2.1(2016-11-07)
 -----------------
--- a/asdf/tags/core/__init__.py
+++ b/asdf/tags/core/__init__.py
@@ -18,6 +18,7 @@ class HistoryEntry(dict, AsdfType):
     name = 'core/history_entry'
 
 
+from .constant import ConstantType
 from .ndarray import NDArrayType
 from .complex import ComplexType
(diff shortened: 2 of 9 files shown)


drdavella:
Rebasing onto `master` revealed some more warnings that were missed. See issue #300 for more details. I'd like all tests to pass without warnings so this should be resolved before merging.

drdavella:
@nden I've added a workaround for testing that accounts for the fact that `label_mapper` and `regions_selector` are not defined within ASDF itself. This should fix #300, although we'll need to discuss a long-term solution at some point. In the meantime, I'd like to merge this PR if you think it looks okay.

drdavella:
@nden I'd like to merge this if you approve. It will probably cause some more warnings in the jwst pipeline but I can help track them down.

drdavella:
@nden I just wanted to update and say that I have tested this PR against the JWST pipeline and it appears that it does not cause any new warnings to occur.

nden:
@drdavella Do I read this correctly - this does not change behaviour (meaning the code already returns the raw data on reading when the type is not found), only issues a warning. There's no change on write - a `RepresenterError` is raised. If so this looks OK to me.

drdavella:
@nden, yes the only change is that a warning occurs where previously there was none. This exposed some latent issues in the ASDF test suites but again doesn't appear to have any effect on the jwst pipeline.

PR #297, opened 2017-08-06
asdf validation error

nden:
The purpose of this PR is to illustrate a failing example. The schema is included here although for the example to work it should be placed under `asdf-standard/schema/.../transform`. The error I get is
```
ValidationError: mismatched tags, wanted 'tag:stsci.edu:jwst_pipeline/test_fail-1.1.0', got 'tag:stsci.edu:asdf/transform/test_fail-1.1.0'
```
The example to trigger the error:
```
from astropy.modeling import models as astmodels
from asdf import AsdfFile
from asdf.tags.transform import test_fail
fa=AsdfFile()
m=test_fail.TestFail(astmodels.Shift(1))
fa.tree['model']=m
 fa.write_to('test_fail.asdf')
```

If I don't specify `version='1.1.0'` in TestFailType, it defaults to version `1.0.0` and the error is
```
ValidationError: mismatched tags, wanted 'tag:stsci.edu:jwst_pipeline/test_fail-1.1.0', got 'tag:stsci.edu:asdf/transform/test_fail-1.0.0'
``` 
This only happens when a specific transform is defined in the schema, like `shift-1.1.0` here. If instead the schema requires `transform-1.1.0`, it works. Since all schemas we have in JWST sofar need a general `transform` and not a specific one this error just surfaced now. 

This is a simplified version of the real use case. @drdavella Do you see anything wrong with the example? If not, is this sufficient information to debug it?

### DIFF
--- a/asdf/tags/transform/test_fail.py
+++ b/asdf/tags/transform/test_fail.py
@@ -1,0 +1,42 @@
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+from ... import yamlutil
+from .basic import TransformType
+from astropy.modeling.core import Model
+
+__all__ = ['TestFailType', 'TestFail']
+
+
+class TestFail(Model):
+    """
+    Class to compute the z coordinate through the NIRSPEC grating wheel.
+
+    """
+    separable = False
+
+    inputs = ("x",)
+    outputs = ("z",)
+
+    def __init__(self, tab, **kwargs):
+        self.tab = tab
+        super(TestFail, self).__init__(**kwargs)
+        
+    def evaluate(self, x):
+        return self.tab(x)
+
+    
+class TestFailType(TransformType):
+    name = "transform/test_fail"
+    types = [TestFail]
+    version = "1.1.0"
+
+
+    @classmethod
+    def from_tree_transform(cls, node, ctx):
+        tab = node['model']
+        return TestFail(tab)
+
+    @classmethod
+    def to_tree_transform(cls, model, ctx):
+        node = {'model': model.tab}
+        return yamlutil.custom_tree_to_tagged_tree(node, ctx)
--- a/asdf/tags/transform/__init__.py
+++ b/asdf/tags/transform/__init__.py
@@ -8,5 +8,6 @@ from .compound import *
 from .projections import *
 from .polynomial import *
 from .tabular import *
+from .test_fail import *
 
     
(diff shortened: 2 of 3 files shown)


drdavella:
@nden, I think the problem is that by default the `standard` field for an `AsdfType` object is `asdf` unless it is redefined in the child class. In this case, I believe that if you add `standard = 'jwst_pipeline'` to the definition of `TestFailType`, it should work.

asdf/tags/transform/test_fail.py
```diff
@@ -0,0 +1,42 @@
...
+        self.tab = tab
+        super(TestFail, self).__init__(**kwargs)
+        
+    def evaluate(self, x):
+        return self.tab(x)
+
+    
+class TestFailType(TransformType):
+    name = "transform/test_fail"
+    types = [TestFail]
```
drdavella: Try adding `standard = 'jwst_pipeline'` here

nden:
@drdavella I was trying to make a simple example showing the problem and forgot to change the standard. However, the example here is supposed to run/work within asdf and changing the standard in the schema definition to `asdf` does not help. I updated the schema to use `standard=asdf`. It only fails if a specific model is defined in the schema (in this example `shift`). If I specify `transform` instead it works. Any other concrete transform fails.

drdavella:
@nden, sorry I guess I skimmed the description of your issue a little too quickly. Is the error you get still the same? The error you pasted above seems to be related to the difference in values for `standard`, so I'm not sure if that was the original issue or just a side effect. 

The current version of the code fails because the tag is under `asdf/test_fail-1.1.0` but the relative schema path is to `../asdf/transform/shift-1.1.0`. which doesn't make sense. Is the tag supposed to be `asdf/transform/test_fail-1.10`? 

If I change the tag, then I get a different failure which is maybe what is actually wrong.

drdavella:
@nden I think I might actually know what the problem is. The `version` field of `TransformType` was never bumped to `1.1.0` when the associated schemas were updated. Since `ShiftType` inherits from `TransformType`, it is also pegged at the older version `1.0.0`, so the library can't find the right version of the schema.

This has been fixed on #294 and can be merged. That PR is based on #284 though, and that will need to be merged as well. I'm not sure if you want to review them quickly or whether I should just go ahead and pull them in.

nden:
@drdavella You are right. Bumping the `version` field of `TransformType` fixed it. I'll take a look at the PRs now.

drdavella:
@nden if it's easier we can just do a quick patch for the version now. I would like to get those PRs merged at some point though.

Issue #100, opened 2015-04-14
overwriting asdf files


nden:
This isn't urgent, I am simply reporting it so it's not forgotten.
When attempting to overwrite an asdf file on Windows I get an error:

```
In [17]: f1.write_to('foc2sky.asdf')
---------------------------------------------------------------------------
IOError                                   Traceback (most recent call last)
<ipython-input-17-1418e770a532> in <module>()
----> 1 f1.write_to('foc2sky.asdf')

C:\Anaconda\envs\gwcs\lib\site-packages\pyasdf-0.0.dev417-py2.7.egg\pyasdf\asdf.
pyc in write_to(self, fd, all_array_storage, all_array_compression, auto_inline,
 pad_blocks)
    586         original_fd = self._fd
    587
--> 588         self._fd = fd = generic_io.get_file(fd, mode='w')
    589
    590         self._pre_write(fd, all_array_storage, all_array_compression,

C:\Anaconda\envs\gwcs\lib\site-packages\pyasdf-0.0.dev417-py2.7.egg\pyasdf\gener
ic_io.pyc in get_file(init, mode, uri)
   1033             realpath = url2pathname(parsed.path)
   1034             return RealFile(
-> 1035                 open(realpath, realmode), mode, close=True,
   1036                 uri=uri)
   1037

IOError: [Errno 22] invalid mode ('wb') or filename: 'foc2sky.asdf'

```

On Linux the file is silently overwritten. Perhaps add a `clobber` argument to `write_to` similar to fits.

mdboom:
I'll see if I can reproduce this on Windows (but will have to reboot into Windows first).  I suspect it may because the file is already open under another handle (which is not allowed on Windows, but is on Unix).

I'm not crazy about adding a `clobber` option -- I think it's better for these things to behave the way the OS' underlying file routines behave so there's consistent expectations.  I've always thought the `clobber` in pyfits was a mistake, but I might be in the minority there.

embray:
I agree; I wish I could rip it out.  I don't want to deal with the fallout of that though.  People were mad enough when I made it raise a KeyError if you try to delete a keyword that doesn't exist.

mdboom:
I can't seem to reproduce this on Windows 8.  There must be something more to it than just overwriting a file.  Keeping the file handle open in the same process, or another process, and trying to write to it then works as well.  Can you provide a script to reproduce it?

nden:
[...]

PR #302, opened 2017-08-08
Add test to reproduce overwrite bug in issue #100

drdavella:
This attempts to reproduce the issue reported in #100. Since we are running on AppVeyor again, hopefully it will either reproduce the issue or demonstrate that it no longer exists.


drdavella:
@nden this PR adds a test for the overwrite bug on Windows that you reported in #100. Since the test passes on AppVeyor I'm inclined to close that issue unless you think it's still a problem.

PR #306, opened 2017-08-17
Display filename in supported version warning

### LINKED ISSUE #305: Display ASDF file name when warning about version mismatch
Requires modification to `fix_yaml_tag` in `asdftypes.py` and callers.

drdavella:
This fixes #305.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,9 @@
 
 - Added command to ``asdftool`` to display available tags. [#303]
 
+- When possible, display name of ASDF file that caused version mismatch
+  warning. [#306]
+
 
 1.2.1(2016-11-07)
 -----------------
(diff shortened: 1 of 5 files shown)


drdavella:
@nden currently this is just a quick-and-dirty fix and should probably be cleaned up a little bit.

drdavella:
Need to add change log entry.

drdavella:
@nden this is ready for merge if it looks good to you.

nden:
LGTM. Thanks.

drdavella:
It looks like the warnings in the JWST pipeline are coming from FITS files with embedded ASDF, and unfortunately this fix doesn't handle that case. I'll open a new issue.

nden:
Yes, having the warnings display a filename is useful when creating reference files. I think we will have to suppress the warnings in the pipeline unless you have a better idea.

drdavella:
These warnings maybe should not be enabled by default. Maybe we can introduce some kind of "strict validation" mode which would cause these warnings to appear, but otherwise they are handled silently. I'll have to think about the best way to allow users to configure that.

On the other hand, I think the warnings introduced in #295 are more likely to indicate real sources of problems, and I think they should **not** be silenced by default. I actually ran the JWST pipeline against that PR and I don't believe that there were any new warnings as a result.

PR #310, opened 2017-08-23
Validate ASDF trees as early as possible

### LINKED ISSUE #287: Intermediate tree validation
The documentation contains the following example:

> This validation happens only when a `AsdfFile` is instantiated, read
or saved, so it's still possible to get the tree into an invalid
intermediate state::

```python
>>> from asdf import AsdfFile
>>> ff = AsdfFile()
>>> ff.tree['data'] = 'Not an array'
>>> # The ASDF file is now invalid, but asdf will tell us when
>>> # we write it out.
>>> ff.write_to('test.asdf')  # doctest: +IGNORE_EXCEPTION_DETAIL
...
ValidationError: mismatched tags, wanted
'tag:stsci.edu:asdf/core/ndarray-1.0.0', got
'tag:yaml.org,2002:str'
...
```

It would be nice if errors like this could be caught immediately rather than deferring to write time, but there may be technical barriers that I'm not thinking about right now.

drdavella:
This change allows ASDF trees to be validated as early as possible. In many cases it will be possible to notice a schema validation issue at the time that an ASDF tree is assigned, rather than having to wait until the file is written.

However, it will not always be possible to catch validation errors that occur when updating existing trees. Enabling this kind of "deep" validation will require a lot of additional machinery that may not be worth the effort.

This fixes #287.

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -242,7 +242,7 @@ class AsdfFile(versioning.VersionedMixin):
 
     @tree.setter
     def tree(self, tree):
-        asdf_object = AsdfObject(tree)
+        asdf_object = AsdfObject(tree, validator=self._validate)
         self._validate(asdf_object)
         self._tree = asdf_object
 
@@ -254,8 +254,7 @@ class AsdfFile(versioning.VersionedMixin):
         return self._comments
 
     def _validate(self, tree):
-        tagged_tree = yamlutil.custom_tree_to_tagged_tree(
-            tree, self)
+        tagged_tree = yamlutil.custom_tree_to_tagged_tree(tree, self)
         schema.validate(tagged_tree, self)
 
     def validate(self):
@@ -852,8 +851,8 @@ class AsdfFile(versioning.VersionedMixin):
         try:
             with generic_io.get_file(fd, mode='w') as fd:
                 self._fd = fd
-                self._pre_write(fd, all_array_storage, all_array_compression,
-                                auto_inline)
+                self._pre_write(
+                    fd, all_array_storage, all_array_compression, auto_inline)
 
                 try:
                     self._serial_write(fd, pad_blocks, include_block_index)
--- a/asdf/tags/core/__init__.py
+++ b/asdf/tags/core/__init__.py
@@ -6,15 +6,30 @@ from __future__ import absolute_import, division, unicode_literals, print_functi
 from ...asdftypes import AsdfType
 
 
-class AsdfObject(dict, AsdfType):
+class ValidateDictMixin(dict):
+    def __init__(self, *args, **kwargs):
+        self._validator = kwargs.get('validator')
+        if self._validator:
+            kwargs.pop('validator')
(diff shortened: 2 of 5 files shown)


nden:
Just to make sure I understand - this only validates `tree` assignments, like `AsdfFile().tree = {...}`, and does not validate `AsdfFile().tree['data'] = 42`. In addition it only validates `data` and `history`,
Hm, my current dev version validates `AsdfFile().tree = {'data': 42}`

More importantly, what effect does this have on performance? It'd be interesting to time a large file with this PR. If there's a big performance hit I'd prefer not to validate at this level.

drdavella:
The example you gave of `AsdfFile().tree = {'data': 42}` already worked. This PR implements the other case: `AsdfFile().tree['data'] = 42`, which was not previously covered.

It should work for any top-level tree assignment, not just `data` and `history`. However, unfortunately it won't work for deeper tree assignments like the following: `AsdfFile().tree['foo']['bar']['baz'] = 'invalid'`.

Do you have a large file that I could use for testing performance?

nden:
Aren't `data` and `history` "reserved" keywords. How would it validate anything else?

As for performance testing, actually it doesn't have to be a large file, but one that is dominated by `data`. A miri imager reading gives:

this PR
timeit datamodels.ImageModel('miri_image.fits')
760 ms ± 3.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

dev master
 timeit datamodels.ImageModel('miri_image.fits')
   ...:
1 loop, best of 3: 491 ms per loop

On writing
this PR
timeit im.to_fits('todel.fits', overwrite=True)
647 ms ± 8.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

dev master
timeit im.to_fits('todel.fits', overwrite=True)
1 loop, best of 3: 430 ms per loop

I'm not sure how much this can be trusted though.

drdavella:
This PR just hooked the `__setitem__` method of the top-level `tree` object so that whenever a new item is added (e.g. `AsdfFile().tree['foo'] = 'bar'`), the yaml validator gets called first.

That makes more of a difference than I would have expected, but I think I might be able to make it more efficient. If you think it's not worth the potential performance cost, we can close the PR. I just like the idea of detecting potential errors as early as possible, but it's not critical.

nden:
JWST data models validate on assignment because each one has a schema. We have discussed a few times removing validation on assignment and adding a `validate` function or providing some other way to validate just once for performance reasons. I also like the idea of validating early but sometimes it's not practical. Perhaps that can be made configurable and off by default.

drdavella:
@nden, huh that's an interesting thought. So you mean having the validation optional even for the `AsdfFile().tree = {'data': ... }` case? There's already a method `AsdfFile().validate()` that should be possible to use for one-time validation.

One question is whether one-time validation should be the default or whether it should have to be added explicitly.

drdavella:
Closing for now since there's not a strong motivation. We should maybe revisit one-time validation later.

PR #312, opened 2017-08-23
Add FITS file name to version mismatch warning

### LINKED ISSUE #311: Display file name when warning from FITS files
Issue #305 requested the file name to be displayed when warning about version mismatches. PR #306 addressed this issue. However, this PR did not address the case when the warning is coming from a FITS file with embedded ASDF.

drdavella:
This fixes #311. It is just a follow-on to #306, so there is no need to update the change log.


nden:
LGTM.

PR #317, opened 2017-08-25
Fix reading json schema in Python 3.5

### LINKED ISSUE #242: Cannot load json files under Python 3
Asdf uses generic_io to open schema files, which explicitly opens the files in binary mode. Under Python 3 reading a file opened in binary mode gives you a byte array. But the json parsing library expects to receive a string and throws an error message if it gets any other type:

TypeError: the JSON object must be str, not 'bytes'

### LINKED ISSUE #314: Asdf fails to load json schema files
Asdf loads schema files using generic_io.get_file. It forces the file to be opened in binary mode. This returns the contents of the file as a byte array under Python 3. The json parser chokes on the byte array, saying it wants the str type.

Adding this issue for tracking, although if someone else wants to take a swing at it, feel free.

drdavella:
This fixes #314 (and fixes #242 which is a duplicate). The JSON module in Python 3.5 doesn't accept `bytearray` as input, but 3.6 does. This PR adds a test and a workaround.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,8 @@
 
 - Issue a warning when an unrecognized tag is encountered. [#295]
 
+- Fix bug with loading JSON schemas in Python 3.5. [#317]
+
 1.2.1(2016-11-07)
 -----------------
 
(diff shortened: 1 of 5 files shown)


bernie-simon:
The fix looks good to me.

PR #318, opened 2017-08-25
Update travis test matrix

### LINKED ISSUE #316: Update latest astropy version in travis config
The config currently has the latest version pegged at `1.3.2`, but the latest released version is `2.0.1`.

For investigation: travis provides a way to specify a `development` version, but can it also specify a `latest` version?

drdavella:
This resolves #316. Instead of hard-coding the latest stable versions of `numpy` and `astropy`, use the `stable` flag provided by [astropy/ci-helpers](https://github.com/astropy/ci-helpers).

Also add a new test to the matrix covering `astropy-1.3.2` since that is now the previous stable version, not the latest.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,7 +38,7 @@ environment:
 
       - PYTHON_VERSION: "3.6"
         ASTROPY_VERSION: "1.3.2"
-        NUMPY_VERSION: "1.11.1"
+        NUMPY_VERSION: "1.13.0"
         platform: x64
 
       - PYTHON_VERSION: "3.6"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,6 +26,7 @@ environment:
 
       - PYTHON_VERSION: "3.5"
         ASTROPY_VERSION: "1.3.2"
+        NUMPY_VERSION: "1.11.1"
         platform: x64
 
       - PYTHON_VERSION: "3.5"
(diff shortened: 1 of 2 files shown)


drdavella:
Updated the appveyor test matrix as well. This de-emphasizes Python 2.7 tests by only testing against stable versions of `astropy` and `numpy`.

appveyor.yml
```diff
@@ -7,23 +7,34 @@ environment:
...
         PYTHON_ARCH: "64"
 
       - PYTHON_VERSION: "3.4"
-        NUMPY_VERSION: "1.9.1"
+        ASTROPY_VERSION: "stable"
+        platform: x64
+        PYTHON_ARCH: "64"
+
+      - PYTHON_VERSION: "3.5"
+        ASTROPY_VERSION: "1.3.2" "stable" "development"
```
nden: Is this valid syntax?

nden:
Is appveyor running on PRs?

appveyor.yml
drdavella: I'm actually not sure. I thought I saw an example along these lines somewhere but this commit was kind of just an experiment.

drdavella:
It has been working on PRs since @jhunkeler reactivated it a week or two ago, but I might have broken something with this latest commit...

appveyor.yml
nden: I assume it should be valid yaml but there may be other things. I saw there's a validation tool here https://ci.appveyor.com/tools/validate-yaml

drdavella: The [appveyor validator](https://ci.appveyor.com/tools/validate-yaml) says no, so I guess it's back to the drawing board. I was hoping to avoid listing each configuration explicitly, but I'm not sure that will be possible.

PR #319, opened 2017-08-25
Configure warnings

### LINKED ISSUE #313: Add mechanism for configuring warning levels
Here are a few assertions that may or may not be reasonable:

- version mismatch warnings are **minor** and should be silenced by default
- tag conversion failure warnings and unsupported version warnings (introduced in #295) are **serious** and should be enabled by default
- it should be possible to enable version mismatch warnings ("strict" mode)
- it should be possible to suppress the serious warnings

I'm less confident about this one, but maybe it would be useful:
- it should be possible to treat warnings as errors

Here are some possible options for configuring warning behavior:
- module-level warning configuration function (e.g. `asdf.set_warning_level(...)`)
- new optional parameters to `asdf.AsdfFile()` and `asdf.open()`
- environment variables (in general I'm 👎  on this, but it could be useful for test environments)
- [INI-style](https://docs.python.org/3.4/library/configparser.html) configuration files (where would they go, though?)
- install-time configuration
- some combination of the above

At this moment I think I'm mostly in favor of a combination of a module-level configuration function along with optional parameters to the `AsdfFile` constructor and `asdf.open()`. The latter parameters would override any module-level configuration for that particular file.

drdavella:
This PR does two things:

1. It makes version mismatch warnings silent by default
2. It enables warnings for unrecognized tags to be silenced

@nden, this seems like the right way to handle the large number of version mismatch warnings in jwst in the long term. However, I think that unrecognized tag warnings should be enabled by default, and there are a few other more serious warnings which are still not possible to silence at all.

Also, while this resolves #313 for the most part, it does not add a module-level configuration mechanism. There are too many questions about precedence that make it a little too complicated to implement without sufficient motivation. Please let me know if you disagree.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,7 +46,10 @@
 - When possible, display name of ASDF file that caused version mismatch
   warning. [#306]
 
-- Issue a warning when an unrecognized tag is encountered. [#295]
+- Issue a warning when an unrecognized tag is encountered. [#295] This warning
+  is silenced by default, but can be enabled with a parameter to the
+  ``AsdfFile`` constructor, or to ``AsdfFile.open``. Also added an option for
+  ignoring warnings from unrecognized schema tags. [#319]
 
 - Fix bug with loading JSON schemas in Python 3.5. [#317]
 
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -42,7 +42,8 @@ class AsdfFile(versioning.VersionedMixin):
     """
     The main class that represents a ASDF file.
     """
-    def __init__(self, tree=None, uri=None, extensions=None, version=None):
+    def __init__(self, tree=None, uri=None, extensions=None, version=None,
+        ignore_version_mismatch=True, ignore_unrecognized_tag=False):
         """
         Parameters
         ----------
@@ -65,7 +66,16 @@ class AsdfFile(versioning.VersionedMixin):
             The ASDF version to use when writing out.  If not
             provided, it will write out in the latest version
             supported by asdf.
+
+        ignore_version_mismatch : bool, optional
+            When `True`, do not raise warnings for mismatched schema versions.
+            Set to `True` by default.
+
+        ignore_unrecognized_tag : bool, optional
+            When `True`, do not raise warnings for unrecognized tags. Set to
+            `False` by default.
         """
+
         if extensions is None or extensions == []:
             self._extensions = extension._builtin_extension_list
         else:
@@ -76,6 +86,9 @@ class AsdfFile(versioning.VersionedMixin):
                     extensions = [extensions]
                 extensions.insert(0, extension.BuiltinExtension())
                 self._extensions = extension.AsdfExtensionList(extensions)
+
+        self._ignore_version_mismatch = ignore_version_mismatch
+        self._ignore_unrecognized_tag = ignore_unrecognized_tag
 
         self._fd = None
         self._external_asdf_by_uri = {}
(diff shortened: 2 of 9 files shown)


drdavella:
@nden, despite the appearance of the diff, this is a pretty small change. I'd like to merge it if you have no objections. It should not cause any more warnings to be produced by the JWST pipeline.

nden:
@drdavella This looks good to me. I have one suggestion. 
Would it be better to update *Type classes to include the `supported_versions` attribute you added in #284. I've added `supported_versions` to transform types and I don't see warnings any more. I think it's good to explicitly state which versions are supported and then warnings will be issued for those that are not.

If so, the default of `ignore_version_mismatch` can be set to `False` and let application choose how to deal with warnings. The reasoning is that sometimes the warning may point to a real problem.

If you agree, just set the default of  `ignore_version_mismatch` to False in this PR. Since I've added `supported_versions` to experiment I can make a PR for that.

drdavella:
@nden I've been trying to clarify my thoughts on this issue. I think I need to add some documentation that describes these changes and the philosophy behind them. In the meantime, here are my current thoughts (which maybe can eventually be converted into documentation in some form):

There are really two different stages at which warnings can occur:
- schema validation, and
- tag conversion

Also, tag classes now have two different attributes related to versioning:
- `version`, and
- `supported_versions`

There are also now two levels of warning control:
- `ignore_version_mismatch` which is set to `True` by default, and
- `ignore_unrecognized_tag`, which is set to `False` by default

The `version` attribute of a tag class describes the current (i.e. latest) version of that tag. It should correspond to the version of the latest available schema for that tag. For example, if we have a schema that defines `foo-3.2.0`, then the `version` attribute of the `FooType` class will be `"3.2.0"`.

During schema validation, we detect mismatched tag versions. For example, if I read an ASDF file that references `foo-3.0.0` or `foo-3.3.0` but the latest schema is `foo-3.2.0`, then a version mismatch warning will occur. This has nothing to do with the tag conversion stage: this is strictly a statement about schema validation. This kind of warning is now controlled by `ignore_version_mismatch`, and it is currently ignored by default since it does not indicate a serious problem.

The `supported_versions` attribute is used to describe the versions of a tag that can successfully be translated by a particular tag class. By default, it is assumed that _all_ versions of a particular tag can be converted by the corresponding tag class. However, this is not always true. For example, if `foo-3.0.0` introduced backwards-incompatible changes, our tag class may only be able to convert version 3.0.0 and later. We can declare this by setting `supported_versions=AsdfSpec('>=3.0.0')`.

During tag conversion, we determine whether a particular schema can successfully converted to a custom Python data structure by a particular tag implementation. Continuing with the example above, if we encounter `foo-3.0.0`, our `supported_versions` attribute tells us we can successfully convert it, so there is no warning, even though the latest version we know of is `foo-3.2.0`. However, if we encounter `foo-2.7.1`, we **can't** convert it, so there will be a warning. This warning **can't** be ignored at all.

The `ignore_unrecognized_tag` option controls warnings that occur when no corresponding tag class can be found for a particular tag. For example, if I read an ASDF file that makes reference to `baz-1.2.0`, but I don't have a `BazType` class, then this warning will occur. These warnings are **not** suppressed by default.

**TL;DR**: It is useful and a good idea to explicitly define `supported_versions` for tag classes (including transform types), but it is not strictly necessary in most cases. However, since `supported_versions` has no bearing on schema mismatch warnings, and since these warnings indicate a relatively minor issue, it seems reasonable to ignore them by default.

drdavella:
@nden reported that the use of `supported_versions` affected the behavior of YAML validation warnings, which is not what I intended. This has now been fixed and I believe that it is ready to be merged as soon as the tests pass.

nden:
OK, so really the version_mismatch warning is of no interest to end users. I think this can merged. I'll open a separate issue about `supported_versions` as I see some potential issues.

drdavella:
I'm not sure it will never be of use to end users, but I think the warnings are of a sufficiently minor nature that they should be have to be explicitly enabled (maybe think of it like `-Wall` for a C compiler).

PR #333, opened 2017-09-06
Remove all remnants of support for Python 2.6

### LINKED ISSUE #329: Remove references to Python < 2.7
Python 2.7 will be deprecated in the next release. All earlier versions of Python will not be supported at all.

drdavella:
This fixes #329. Since Python 2.7 support will be removed entirely in v2.0.0, it is reasonable to completely remove support for Python 2.6 now. Even though #191 also addressed this issue, there were places in the code where Python 2.6 was still handled.


drdavella:
@nden if you have a chance for a quick look-over I think this one is ready for merge.

PR #343, opened 2017-09-27
Fix bug with astropy.time location, update time schema

### LINKED ISSUE #341: Failure to serialize astropy.time.Time objects with location attribute
Here is a simple case that reproduces the issue:
```python
from astropy.time import Time
from asdf import AsdfFile

t = Time([1,2], location=[[1,2], [3,4], [5,6]], format='cxcsec')
af = AsdfFile(tree={'time': t})
af.write_to('time.asdf')
```

Add a test case and fix it. Optionally, create a new tag to serialize `EarthLocationInfo` objects.

drdavella:
This fixes #341. Time objects from astropy are now serialized correctly when they have an `EarthLocation` attribute. The proper way to implement this fix was to update the time schema to `v1.1.0`, so this PR requires https://github.com/spacetelescope/asdf-standard/pull/147 to be merged first.

This change also supports backwards compatible writing with `time-1.0.0` in order to support tests for WCS to ensure backwards compatibility with the old frame format.

### DIFF
--- a/asdf/tags/time/tests/test_time.py
+++ b/asdf/tags/time/tests/test_time.py
@@ -104,7 +104,12 @@ def test_time(tmpdir):
 
 def test_time_with_location(tmpdir):
     # See https://github.com/spacetelescope/asdf/issues/341
-    t = time.Time([1,2], location=[[1,2], [3,4], [5,6]], format='cxcsec')
+    from astropy import units as u
+    from astropy.coordinates.earth import EarthLocation
+
+    location = EarthLocation(x=[1,2]*u.m, y=[3,4]*u.m, z=[5,6]*u.m)
+
+    t = time.Time([1,2], location=location, format='cxcsec')
 
     tree = {'time': t}
 
--- a/asdf/tags/time/tests/test_time.py
+++ b/asdf/tags/time/tests/test_time.py
@@ -7,7 +7,6 @@ import six
 import pytest
 import datetime
 from collections import OrderedDict
-from jsonschema import ValidationError
 
 astropy = pytest.importorskip('astropy')
 from astropy import time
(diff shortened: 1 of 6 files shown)


drdavella:
@nden I'll run this PR against the jwst pipeline to assess if there's any impact there.

drdavella:
This needs a changelog update. I really ought to implement a bot for this repo...

drdavella:
This PR will need to be rebased on PR #350 when it is merged in order for the tests to pass.

asdf/tags/time/tests/test_time.py
```diff
@@ -103,6 +102,15 @@ def test_time(tmpdir):
     helpers.assert_roundtrip_tree(tree, tmpdir)
 
 
+def test_time_with_location(tmpdir):
+    # See https://github.com/spacetelescope/asdf/issues/341
+    t = time.Time([1,2], location=[[1,2], [3,4], [5,6]], format='cxcsec')
```
nden: If the schema defines the `x`, `y`, `z` of `location` to be quantities, then why does this pass?

drdavella: Good question: it's because the list that gets passed to the constructor of `time.Time` gets converted into an `EarthLocation` object, which uses `Quantity` for each of the coordinate components. This test could have also explicitly created an `EarthLocation` object and passed it to the `Time` constructor, but since the conversion is done automatically there's no need.

drdavella:
I'm not sure why the OSX tests keep having errors, but all of the tests pass. This PR is ready for merge once spacetelescope/asdf-standard#147 is merged.

asdf/tags/time/tests/test_time.py
nden: I see. The example is a bit confusing because it's unclear that default units in coordinates are used. But it works.

nden:
👍 to merge.

asdf/tags/time/tests/test_time.py
drdavella: Good point. I updated it so hopefully it's more clear now.

PR #346, opened 2017-09-28
Remove astropy-1.1.2 from test matrix, add 1.3.3

### LINKED ISSUE #345: Add astropy v1.3.3 to the test matrix
This is a fairly significant oversight and I believe it will expose a few latent bugs (although hopefully not very serious ones).

drdavella:
This also removes a test against astropy-1.3.2 using Python 3.5 from the appveyor matrix since it seems redundant.

This resolves #345.


drdavella:
Merging even though it will cause test failures. Opening a new PR to address the issues.

PR #348, opened 2017-09-28
Fix WCS test failures in astropy v1.3.3

### LINKED ISSUE #347: Testing against astropy v1.3.3 causes WCS-related failures
Astropy v1.3.3 was added to the test matrix in PR #346. This exposed some (fairly minor) test cases that now fail. Addressing this should be relatively straightforward, but it likely involves making the WCS tag implementation aware of the installed version of astropy.

drdavella:
This addresses the bugs that were revealed when updating the test matrix in #346. It fixes #347.


drdavella:
Going ahead and merging this so that it doesn't break JWST nightly build.

PR #349, opened 2017-09-28
Fix mechanism used to handle missing requirements for tag classes

drdavella:
Instead of directly issuing a warning and passing the raw node back in this case, raise an exception that will be handled by `tagged_tree_to_custom_tree`, which will then use the existing warning mechanism to indicate a failure to convert the tree. This is much cleaner since it reuses mechanisms that were recently improved; and it avoids some issues that occurred when running tests against older versions of astropy.

This also helps address an issue that was uncovered by PR #343 when running against older versions of astropy.


drdavella:
This needs to be rebased on PR #348 in order to fix the test failures when running against `astropy-1.3.3`.

PR #350, opened 2017-09-29
Fix issues with the type index for writing

drdavella:
There were several issues with the type index used for writing ASDF file. These bugs were causing issues that were exposed in PR #343. This PR fixes those issues. 

One of the main problems was that items in the type index for earlier versions of the standard (i.e. `1.0.0`) were being overwritten by more recent version of the same type. This meant that files that used the earlier version of the standard were not being validated properly.

Also, this PR make sure that all of the tests of examples in schema files run against the appropriate version of the ASDF standard. It is not enough to assume in all cases that the examples can be tested against the latest version of the standard.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,11 @@
 
 - Remove all remnants of support for Python 2.6. [#333]
 
+- Fix issues with the type index used for writing out ASDF files. This ensures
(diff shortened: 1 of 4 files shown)


drdavella:
I believe that merging this should resolve the current test failure on `master`.

PR #355, opened 2017-10-06
Add copy option to asdf.open

### LINKED ISSUE #226: allow opening an asdf file without memory mapping of arrays
A boolean argument `copy` to `AsdfFile.open()` would be very useful.

drdavella:
This resolves #226 and allows callers to optionally specify that underlying data blocks should not be memory mapped. Memory mapping is not possible under all circumstances (e.g. when an array is compressed), but is enabled by default if it is possible. Using `copy=True` will ensure that the array is **not** memory mapped, even if it is possible.

This does not affect ASDF-in-FITS since apparently memory mapping of FITS files depends on the HDUs and keywords that are present, so it is out of our control.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -63,6 +63,9 @@
 - Update time schema to reflect changes in astropy. This fixes an outstanding
   bug. [#343]
 
+- Add ``copy`` option to ``asdf.open`` to control whether or not underlying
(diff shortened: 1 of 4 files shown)


drdavella:
@nden I'm actually thinking about changing the `copy` argument to `copy_arrays` or `copy_blocks` before merging, just to be a little more explicit.

Issue #265, opened 2017-07-07
Move astropy- and gwcs-specific tags to their respective libraries


drdavella:
There are several schemas and tag implementations that are used to serialize objects that are defined and implemented in either `astropy` or `gwcs`. These schemas and tag implementations don't really belong with ASDF proper on a philosophical level since they are application-specific extensions, not core ASDF components. Moving these schemas and tags closer to the implementation of the objects they serialize will allow more thorough testing of these schemas when  the underlying objects are changed by `astropy` and `gwcs`. Resolving this issue will require removing the tag implementations from this project, removing the schema definitions from [asdf-standard](https://github.com/spacetelescope/asdf-standard), and submitting PRs to `astropy` and `gwcs` with these schema definitions and tag implementations.

drdavella:
The goal is to implement these changes before the `astropy` users meeting in September.

drdavella:
This has been addressed by #359 and #417.

Issue #153, opened 2015-07-22
Remove astropy as hard dependency


mdboom:

mdboom:
Fixes #151.

mdboom:
Just a status update: pyasdf itself no longer requires astropy, but running the tests does.  Once the testing framework in astropy has been sufficiently refactored out, we can remove the dependency there as well.

drdavella:
Addressing issue #265 will be a big step towards this goal. We still need to separate the `astropy` testing framework from `astropy` proper, though.

drdavella:
There's also an `astropy` dependency for parsing FITS files, although this may already be a 'soft' dependency.

PR #359, opened 2017-10-23
WIP: Move astropy-specific tag implementations to astropy

### LINKED ISSUE #281: Figure out what to do with TransformType
With the division of the custom tag type class hierarchy into `AsdfType` for built-ins and `CustomType` for third-party tags, we need to figure out what to do with `TransformType`. This is used heavily by `gwcs`, and since it inherits from `AsdfType`, it means that all custom `gwcs` types are treated as built-ins, which is not good. See [this discussion](https://github.com/spacetelescope/asdf/wiki/Custom-Tag-Implementation) on the wiki for more details.

drdavella:
This represents the first step towards moving astropy-specific tag implementations out of ASDF and into astropy. At the moment, this PR needs to build against an unmerged branch in a fork of astropy, and so the CI scripts have been updated accordingly.

This partially addresses #265. It also is a good step in the direction of #153.

This PR also closes #281, since `TransformType` is now implemented in `astropy` and no longer inherits from `AsdfType`.


drdavella:
Note that this PR needs to be processed in coordination with changes to astropy. These changes will disallow backwards compatibility with earlier versions of astropy. Consequently, it does not make sense to test these changes against Py2, so these tests have been removed from the CI scripts.

PR #361, opened 2017-10-24
Relax requirement on semantic_version version to 2.3.1

### LINKED ISSUE #360: semantic_version version requirement too strong
The version of the "semantic_version" package is given as 2.6.0, which seems to be unnecessarily strong: There seems to be nothing in the [`ChangeLog`](https://github.com/rbarrois/python-semanticversion/blob/543299ebb0a73c9ee513ac1165ed3732aaa8c94f/ChangeLog) of semantic_version.
For the Debian build, we cannot just download newer versions of a package, and the semantic_version version in Debian is still 2.3.1. Building asdf with that version works however fine.

Could you consider reducing the version requirement for semantic_version?

drdavella:
This fixes #360. 

cc @olebole

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 1.3.1 (Unreleased)
 ------------------
 
+- Relax requirement on ``semantic_version`` version to 2.3.1. [#361]
+
 1.3.0 (2017-10-24)
 ------------------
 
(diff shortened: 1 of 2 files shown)


drdavella:
Current test failure is due to an upstream issue in `astropy` that was caused by the latest development version of `numpy`. It has been resolved and is awaiting merge: https://github.com/astropy/astropy/pull/6785

PR #362, opened 2017-10-24
CI script update

drdavella:
Start to scale back tests against Py2. Also run tests against development version of `numpy`. These tests are allowed to fail.


drdavella:
OSX builds are hanging, so cancelling the travis build and merging.

PR #381, opened 2017-11-14
WIP: Attempt to save inline object ndarray

Cadair:
This currently errors with a schema error. Putting this here now so @drdavella can have a look.


drdavella:
Thanks @Cadair! Closing this issue for now. We will reopen if we decide to move forward with #383.

PR #382, opened 2017-11-14
Test numpy array with dtype=object

drdavella:
@Cadair has pointed out that it is not currently possible to serialize a `numpy` array with `dtype=object`. Consider the following  code example, which does not currently work:

```python
import asdf
import numpy as np

data = np.empty((10, 10), dtype=object)
for i, _ in enumerate(data.flat):
    data.flat[i] = SomeObject(i, foo, 42)

af = asdf.AsdfFile({'data': data})
af.write_to('data.asdf')
```

In principle, it seems possible to support inline `ndarray`s that contain arbitrary object data. We should discuss whether this should be supported. In the meantime, this PR introduces unit tests that demonstrate the kind of functionality that would be expected. They are currently marked with `pytest.mark.xfail` since they will not pass.

See [this PR](https://github.com/spacetelescope/asdf-standard/pull/148) in `asdf-standard` for the kind of change that might be required to the `ndarray` schema.

### DIFF
--- a/asdf/tags/core/tests/test_ndarray.py
+++ b/asdf/tags/core/tests/test_ndarray.py
@@ -786,3 +786,26 @@ def test_inline_shape_mismatch():
     with pytest.raises(ValueError):
         with asdf.AsdfFile.open(buff) as ff:
             pass
+
+
+@pytest.mark.xfail(
+    reason="NDArrays with dtype=object are not currently supported")
+def test_simple_object_array(tmpdir):
+    dictdata = np.empty((3, 3), dtype=object)
+    for i, _ in enumerate(dictdata.flat):
+        dictdata.flat[i] = {'foo': i*42, 'bar': i**2}
+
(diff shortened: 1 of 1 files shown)


perrygreenfield:
I may be missing something but support for arbitrary objects seems like a very big deal and not necessarily desirable, if only for security reasons. ASDF shouldn't promise to be able to reconstitute any Python object automatically. Perhaps a scheme for saving all reasonable info about an object may make sense, but without any code or such. For example, what if an object was a file handle or something equivalent for which serialization makes little sense? Is the implication here that someone will supply schemas and code support for all the supported objects so that they can be serialized and reconstituted? Or is this completely open ended?

Cadair:
For me it was arrays of objects that have asdf tags, not any arbitrary objects.

perrygreenfield:
Ah, much better :-) In that case I don't have an objection.

drdavella:
Sorry, my example here is misleading. `SomeObject` is intended as a stand-in for a tagged class (or a basic Python type). I'll update the example to be clearer.

PR #384, opened 2017-11-14
Install and load extensions using setuptools entry points

### LINKED ISSUE #378: Create ASDF extension registry interface
Right now packages that provide custom schemas and/or implement ASDF tags need to pass an extension object to each call to `asdf.AsdfFile` or `asdf.open`. This is cumbersome and not at all ideal. We should provide a mechanism for registering extensions either on a per-process basis (e.g. a function that is called at the beginning of a program), or an a per-package basis (allowing extensions to be registered when a package is installed).

@Cadair suggested using entry points in `setup.py` to allow packages to register custom extensions. Packages would define their extensions something like this:
```python
# setup.py
from setuptools import setup

entry_points = {
    'asdf.extensions': ['foobar=foobar.extensions:FoobarExtension']
}

setup(entry_points=entry_points)
```

This seems like a really nice solution. It uses an existing mechanism provided by `setuptools`, and it avoids the more awkward solution of providing a top-level function like `asdf.register_extensions`.

drdavella:
This closes #378. It allows ASDF extensions to be installed and loaded using `setuptools` entry points. ASDF's own `BuiltinExtension` is now installed this way. Other packages, such as `gwcs` or `astropy` that provide ASDF extensions should use entry points to install their own extensions.

cc @Cadair

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ env:
         - NUMPY_VERSION=stable
         - CONDA_DEPENDENCIES='semantic_version jsonschema pyyaml six lz4'
         - GWCS_GIT='git+git://github.com/spacetelescope/gwcs.git#egg=gwcs'
-        - PIP_DEPENDENCIES="$GWCS_GIT pytest-astropy"
         - SETUP_CMD='test --remote-data'
 
     matrix:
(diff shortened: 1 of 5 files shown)


drdavella:
One consequence of this change is that the package must be installed (`develop` mode is okay too) before testing. This required a modification to our CI scripts.

nden:
So, the old way to include extensions is still available, right? 
I think it makes sense to use entry points for gwcs and jwst.
Should this be documented (I expect other packages may want to use it).

drdavella:
@nden, yes it is still possible to pass extensions to `asdf.open` or `asdf.AsdfFile`. Right now the behavior when doing so is to append the given extension to the list of extensions that were installed as entry points. Maybe there should be an option to ignore installed extensions as well.

I need to figure out what's going on with the Windows tests. I don't think it's a problem with the code changes; I think there's probably something weird in our appveyor script.

PR #389, opened 2017-11-17
WIP: refactor AsdfExtension base classes

### LINKED ISSUE #386: Implement abstract base class for ASDF extensions
Right now `AsdfExtension` implements `__subclasshook__` so that it can recognize foreign extensions even if the foreign extension doesn't inherit directly from `AsdfExtension`. In most cases, it is not appropriate for foreign extensions to inherit from `AsdfExtension`.

However, using an abstract base class that can be used as a parent for both `AsdfExtension` and other extensions would make this much cleaner. It would allow invalid extension definitions to be checked much earlier than they are right now.

cc @Cadair

Cadair:
This is an attempt to fix #386.

I can't really work out why it is regarded as bad to subclass `AsdfExtension` is there a registry of those classes somewhere I should know about?

### DIFF
--- a/asdf/extension.py
+++ b/asdf/extension.py
@@ -12,13 +12,13 @@ from . import resolver
 
 
 @six.add_metaclass(abc.ABCMeta)
-class AsdfExtension(object):
+class AsdfExtensionBase(object):
     """
     Subclass to define an extension to ASDF.
     """
     @classmethod
     def __subclasshook__(cls, C):
-        if cls is AsdfExtension:
+        if cls is AsdfExtensionBase:
             return (hasattr(C, 'types') and
                     hasattr(C, 'tag_mapping') and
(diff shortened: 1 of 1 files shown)


drdavella:
Here's the rationale for this change: in general, when an external package implements a custom extension, it is going to need to implement custom values for all three properties of an extension class: `types`, `tag_mapping`, and `url_mapping`.

When external packages inherit `AsdfExtension`, they can accidentally forget to override any or all of the required properties. There will be no error if they make this mistake since any omitted properties in the child class are inherited from `AsdfExtension`. This could lead to bugs that would be difficult to diagnose.

However, if external packages inherit an abstract base class for extension types, then they will be forced to implement all three properties.

drdavella:
Thanks @Cadair! Closing this issue, but thanks for following me down the rabbit hole.

PR #390, opened 2017-11-17
Deprecate not inheriting from AsdfExtension

Cadair:
Take two!

### DIFF
--- a/asdf/extension.py
+++ b/asdf/extension.py
@@ -118,8 +118,8 @@ class AsdfExtensionList(object):
         for extension in extensions:
             if not isinstance(extension, AsdfExtension):
                 raise TypeError(
-                    "Extension must implement the asdf.extension.AsdfExtension "
-                    "interface")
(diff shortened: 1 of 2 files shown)


asdf/extension.py
```diff
@@ -113,8 +114,12 @@ def __init__(self, extensions):
         for extension in extensions:
             if not isinstance(extension, AsdfExtension):
                 raise TypeError(
-                    "Extension must implement asdftypes.AsdfExtension "
+                    "Extension must implement the asdf.extension.AsdfExtension "
                     "interface")
+            elif not issubclass(type(extension), AsdfExtension):
+                warnings.warn("Extensions not inheriting from asdf.extension.AsdfExtension"
```
drdavella: I think it would be helpful to have the name of the extension in the warning message if possible.

drdavella:
@Cadair I think this was misguided (my idea, my fault). I'm not sure it's even possible to do this check properly in Python. I'm closing the PR for now.

PR #393, opened 2017-11-21
Do not automatically load Astropy extension

drdavella:
Now that extensions can be installed using `setuptools` entry points, we won't make a special case of loading Astropy extensions. Instead, Astropy will install its extensions using entry points, which will be used as long as Astropy is installed.

**NOTE**: This branch updates the CI scripts to build off of a feature branch in Astropy. This will need to be reverted before merging.


.travis.yml
```diff
@@ -22,7 +22,8 @@ env:
         # overidden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
         - PYTHON_VERSION=3.6
-        - ASTROPY_VERSION=development
+        #- ASTROPY_VERSION=development
+        - ASTROPY_GIT='git+git://github.com/drdavella/astropy@asdf-entry-point#egg=astropy'
```
nden: Does astropy not use entry points to install the extension yet? Just wondering why you are pointing to your github repo.

.travis.yml
```diff
@@ -62,6 +63,7 @@ matrix:
 install:
     - git clone git://github.com/astropy/ci-helpers.git
     - source ci-helpers/travis/setup_conda.sh
+    - python -m pip install $ASTROPY_GIT
```
nden: What is the advantage of using `python -m` instead of `PIP_DEPENDENCIES`?

.travis.yml
drdavella: No, not yet. I'm going to submit that PR today also and then merge both of these at the same time.

.travis.yml
drdavella: I actually should have used `PIP_DEPENDENCIES` here. The reason for using it below is that `gwcs` needs to be installed with `--no-deps`, but `astropy-helpers` needs to install the dependencies. I don't think I can do this with `PIP_DEPENDENCIES` since it will use the same `pip` flags in both cases, which doesn't work. This is what was causing me so much trouble on the CI runs on a previous PR.

nden: Ah, good to know. Thanks!

PR #399, opened 2017-11-27
Allow foreign tags to be resolved in schemas and files

### LINKED ISSUE #392: Tag mapping does not get used when resolving schemas

drdavella:
This fixes #392. Previously it was not possible to resolve foreign tag references that were included in ASDF schemas or files. For example, if a schema in an external package made reference to `tag:stsci.edu:asdf/core/ndarray-1.0.0`, it would not be resolved properly. 

This PR combines tag mappings and url mappings into a single resolver, so that both tags and urls used as IDs can be resolved to real urls. 

@nden this change will enable schemas to move to Astropy and GWCS as appropriate.

**[edit]**
To be clear, this change applies to *fully specified* tag references like `tag:stsci.edu:asdf/core/ndarray-1.0.0`. Relative tag references within the same organization (such as `stsci.edu` worked (e.g. `!core/ndarray-1.0.0`), but these are not sufficient for use by packages that define schemas in different organizations (e.g. `astropy.org`).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@
 
 - Automatically initialize ``asdf-standard`` submodule in ``setup.py``. [#398]
 
+- Allow foreign tags to be resolved in schemas and files. Deprecate
+  ``tag_to_schema_resolver`` property for ``AsdfFile`` and
+  ``AsdfExtensionList``. [#399]
+
 1.3.2 (unreleased)
 ------------------
 
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -193,6 +193,9 @@ class AsdfFile(versioning.VersionedMixin):
     def url_mapping(self):
         return self._extensions.url_mapping
 
+    def resolver(self, uri):
+        return self.url_mapping(self.tag_to_schema_resolver(uri))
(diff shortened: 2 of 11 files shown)


nden:
@drdavella  I think the case you described worked before too. we have schemas and tags in `jwst` which reference schemas/tags in `asdf`. Do you mean that  now tags with a different organization (like astropy) can be resolved?
I wish there were docstrings in some of the private functions to help with understanding the code. In any case this looks good! 👍 to merge.

drdavella:
@nden I don't think any of the tag references in `jwst` were fully specified: I think they were all relative references like `!core/ndarray-1.0.0` rather than `tag:stsci.edu:asdf/core/ndarray-1.0.0`. That works for `jwst` since it shares the same organization of `stsci.edu`, but it does not work when other external packages try to use `tag:stsci.edu:asdf/*`. It also would not work if another package tried to refer to something like `tag:astropy.org/astropy/*`, so this change was necessary to support both directions.

And yes, I do need to improve documentation both in the code itself and in the standalone documentation.

nden:
@drdavella OK. Thanks for the clarification.

PR #400, opened 2017-11-28
Add an external array reference type

### LINKED ISSUE #369: Support referencing data (and metadata) in external files that are not ASDF
ASDF currently supports references to data in external ASDF files and expolding ASDF files into multiple files. I would like to be able to reference data (and maybe metadata) in FITS files (preferably in a library independent manner).

What I would want to use this for is to have a collection of FITS files (representing a multi-dim dataset) where the ASDF file provides metadata (WCS information and other things) but the actual data (in 2D frames) is stored in FITS files.

Cadair:
This is a first attempt at adding a general tag for external array references. This needs testing with things other than FITS and much more work, but it's put here for feedback.

see also spacetelescope/asdf-standard#149, fixes #369

### DIFF
--- a/asdf/external_reference.py
+++ b/asdf/external_reference.py
@@ -1,0 +1,54 @@
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+from .asdftypes import AsdfType
+
+
+class ExternalArrayReference(AsdfType):
+    """
+    Store a reference to an array in an external File.
+
+    Parameters
+    ----------
+
+    filepath: `str`
+        The path to the path to be referenced. Can be relative to the file
+        containing the reference.
+
+    target: `object`
+        Some internal target to the data in the file. Examples may include a HDU
+        index, a HDF path or an asdf fragment.
+
+    dtype: `str`
+        The (numpy) dtype of the contained array.
+
+    shape: `tuple`
+        The shape of the array to be loaded.
+    """
+    name = "core/externalarray"
+
+    def __init__(self, filepath, target, dtype, shape):
+        self.filepath = str(filepath)
+        self.target = target
+        self.dtype = dtype
+        self.shape = tuple(shape)
+
+    def __repr__(self):
+        return "<External array reference in {0} at {1} shape: {2} dtype: {3}>".format(
+            self.filepath, self.target, self.shape, self.dtype)
+
+    def __str__(self):
+        return repr(self)
+
+    @classmethod
+    def to_tree(self, data, ctx):
+        node = {}
+        node['filepath'] = data.filepath
+        node['target'] = data.target
+        node['datatype'] = data.dtype
+        node['shape'] = data.shape
+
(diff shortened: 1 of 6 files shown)


asdf/external_reference.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import absolute_import, division, unicode_literals, print_function
```
drdavella: Minor comment but we'll be removing Py2 support in the next major release, so no need to include this here.

asdf/external_reference.py
```diff
@@ -0,0 +1,54 @@
...
+
+class ExternalArrayReference(AsdfType):
+    """
+    Store a reference to an array in an external File.
+
+    Parameters
+    ----------
+
+    filepath: `str`
+        The path to the path to be referenced. Can be relative to the file
```
drdavella: Maybe this should actually be a URI since this might be useful for referencing files that don't necessarily live on the host file system.

asdf/external_reference.py
```diff
@@ -0,0 +1,54 @@
...
+        index, a HDF path or an asdf fragment.
+
+    dtype: `str`
+        The (numpy) dtype of the contained array.
+
+    shape: `tuple`
+        The shape of the array to be loaded.
+    """
+    name = "core/externalarray"
+
```
drdavella: We need an explicit `version` field here as well. For now it should just be `1.0.0`.

Cadair:
@drdavella I think this is rapidly approaching feature complete. It's very simple in reality as it's only supposed to be a standard way of referencing things.

asdf/tags/core/external_reference.py
```diff
@@ -0,0 +1,64 @@
+from ...asdftypes import AsdfType
+
+
+class ExternalArrayReference(AsdfType):
```
drdavella: It's a mouthful, but just wondering whether this should be `ExternalArrayReferenceType` for consistency.

asdf/tags/core/external_reference.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Store a reference to an array in an external File.
+
+    This class is a simple way of referring to an array in another file. It
+    provides no way to resolve these references, that is left to the user.
```
drdavella: Might want to add here that it doesn't even check whether the externally referenced file exists or not.

asdf/tags/core/external_reference.py
drdavella: Actually maybe not since this is both the tag type and the type itself, isn't it?

Cadair:
@drdavella I think I have tided this up, I only added more docs to the docstring, I didn't know where to put any other docs. I am open to more suggestions.

drdavella:
All tests pass except for OSX, which is backlogged. I'm going to go ahead and merge anyway.

PR #401, opened 2017-11-28
ICRSCoord tag implementation is moving to astropy

drdavella:


drdavella:
@nden yes, actually this PR needs a little more work. I'm looking into it now and I can let you know when the tests pass.

drdavella:
**NOTE**: https://github.com/spacetelescope/asdf-standard/pull/150 will need to be merged and the `asdf-standard` submodule reference will need to be updated here before merging this PR.

PR #410, opened 2017-12-12
Add test of setup.py install to CI matrices

drdavella:


drdavella:
All tests have passed except OSX, which is still queued, so merging.

PR #411, opened 2017-12-13
Fix bug that caused serialized FITS tables to be duplicated

### LINKED ISSUE #404: fits_embed inlines binary tables
Datamodels calls the code in fits_embed.py to serialize the asdf tree it creates so that it can be written to the asdf extension. Binary tables in the asdf tree are inlined in the serialization. This results in saved fits files having two copies of the binary table. A patch has been added to datamodels as a temporary fix, but the issue should be resolved in the asdf code.

drdavella:
This fixes #404. The bug occurred specifically in the case where the ASDF file was embedded inside an HDU in a FITS file.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,9 @@
   ``tag_to_schema_resolver`` property for ``AsdfFile`` and
   ``AsdfExtensionList``. [#399]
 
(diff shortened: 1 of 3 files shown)


bernie-simon:
The change to fits_embed.py fixes the binary table duplication problem in my test case.

PR #412, opened 2017-12-14
Use new non-standard FITS extension for ASDF-in-FITS

### LINKED ISSUE #336: Use FOREIGN extension instead of IMAGE for ASDF embedded in FITS
This was reported as an issue on the [JWST pipeline Jira](https://jira.stsci.edu/browse/JP-177).

Astropy does not currently provide a `ForeignHDU`, but it's possible that `NonStandardExtHDU` would serve these purposes. It's possible that another solution would work as well. More investigation is required.

drdavella:
The motivation for this update was a problem report from the archive. They pointed out that storing ASDF metadata in an `IMAGE` HDU goes against convention and would likely lead to incorrect handling of the data when the resulting file is processed by FITS readers.

It was originally proposed to instead use the `FOREIGN` extension. However, this convention [has already been defined](https://arxiv.org/abs/1201.1829), and it seemed too restrictive to have to conform to this convention, especially since there is not already an implementation in `astropy`.

Instead, it was agreed that we would create a new `ASDF` extension for storing ASDF metadata in FITS files. This PR contains the implementation of the corresponding HDU, which is nearly identical to the `FitsHDU` in `astropy.io.fits`. In this case, the `XTENSION` keyword is `ASDF`. The ASDF library writes and expects to read extensions with an `EXTNAME` of `ASDF`.

The `AsdfInFits` implementation has been updated to write new files using the new convention by default. However, it is still capable of reading older files using the `IMAGE` extension, and it is still possible to optionally write files using the `IMAGE` extension as well.

I expect that these changes will break a lot of tests in the JWST pipeline, so I will make sure that everyone is on board before hitting the merge button.

This closes #336.

### DIFF
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -20,13 +20,80 @@ from . import generic_io
 
 try:
     from astropy.io import fits
+    from astropy.io.fits.file import _File
+    from astropy.io.fits.header import Header, _pad_length
+    from astropy.utils import lazyproperty
 except ImportError:
     raise ImportError("AsdfInFits requires astropy")
 
 
 ASDF_EXTENSION_NAME = 'ASDF'
-
 FITS_SOURCE_PREFIX = 'fits:'
+
+
+class AsdfHDU(fits.hdu.base.NonstandardExtHDU):
+    """
+    A non-standard extension HDU for encapsulating an entire ASDF file within a
+    single HDU of a container FITS file.  These HDUs have an extension (that is
+    an XTENSION keyword) of ASDF.
+    """
+
+    _extension = ASDF_EXTENSION_NAME
+
+    @classmethod
+    def from_buff(cls, buff, compress=False, **kwargs):
+        """
+        Creates a new AsdfHDU from a given AsdfFile object.
+
+        Parameters
+        ----------
+        buff : io.BytesIO
+            A buffer containing an ASDF metadata tree
+        compress : bool, optional
+            Gzip compress the contents of the ASDF HDU
+        """
+
+        if compress:
+            buff = gzip.GzipFile(fileobj=buff, mode='wb')
+
+        # A proper HDU should still be padded out to a multiple of 2880
+        # technically speaking
+        data_length = buff.tell()
+        padding = (_pad_length(data_length) * cls._padding_byte).encode('ascii')
+        buff.write(padding)
+
+        buff.seek(0)
+
+        cards = [
+            ('XTENSION', cls._extension, 'ASDF extension'),
+            ('BITPIX', 8, 'array data type'),
+            ('NAXIS', 1, 'number of array dimensions'),
+            ('NAXIS1', len(buff.getvalue()), 'Axis length'),
+            ('PCOUNT', 0, 'number of parameters'),
+            ('GCOUNT', 1, 'number of groups'),
+            ('COMPRESS', compress, 'Uses gzip compression'),
+            ('EXTNAME', cls._extension, 'Name of ASDF extension'),
+        ]
+
+        header = Header(cards)
+        return cls._readfrom_internal(_File(buff), header=header)
+
+
+    @classmethod
+    def match_header(cls, header):
+        card = header.cards[0]
+        if card.keyword != 'XTENSION':
+            return False
+        xtension = card.value
+        if isinstance(xtension, str):
+            xtension = xtension.rstrip()
+        return xtension == cls._extension
+
+    # TODO: Add header verification
+
+    def _summary(self):
+        # TODO: Perhaps make this more descriptive...
+        return (self.name, self.ver, self.__class__.__name__, len(self._header))
 
 
 class _FitsBlock(object):
@@ -222,9 +289,17 @@ class AsdfInFits(asdf.AsdfFile):
         return cls._open_asdf(self, buff, uri=uri, mode='r',
                               validate_checksums=validate_checksums)
 
+    def _create_hdu(self, buff, use_image_hdu):
+        # Allow writing to old-style ImageHDU for backwards compatibility
+        if use_image_hdu:
+            array = np.frombuffer(buff.getvalue(), np.uint8)
+            return fits.ImageHDU(array, name=ASDF_EXTENSION_NAME)
+        else:
+            return AsdfHDU.from_buff(buff)
+
     def _update_asdf_extension(self, all_array_storage=None,
-                               all_array_compression=None,
-                               auto_inline=None, pad_blocks=False):
+                               all_array_compression=None, auto_inline=None,
+                               pad_blocks=False, use_image_hdu=False):
         if self.blocks.streamed_block is not None:
             raise ValueError(
                 "Can not save streamed data to ASDF-in-FITS file.")
@@ -235,22 +310,19 @@ class AsdfInFits(asdf.AsdfFile):
             all_array_compression=all_array_compression,
             auto_inline=auto_inline, pad_blocks=pad_blocks,
             include_block_index=False)
-        array = np.frombuffer(buff.getvalue(), np.uint8)
 
-        try:
-            asdf_extension = self._hdulist[ASDF_EXTENSION_NAME]
-        except (KeyError, IndexError, AttributeError):
-            self._hdulist.append(fits.ImageHDU(array, name=ASDF_EXTENSION_NAME))
-        else:
-            asdf_extension.data = array
+        if ASDF_EXTENSION_NAME in self._hdulist:
+            del self._hdulist[ASDF_EXTENSION_NAME]
+        self._hdulist.append(self._create_hdu(buff, use_image_hdu))
 
     def write_to(self, filename, all_array_storage=None,
                  all_array_compression=None, auto_inline=None,
-                 pad_blocks=False, *args, **kwargs):
+                 pad_blocks=False, use_image_hdu=False, *args, **kwargs):
         self._update_asdf_extension(
             all_array_storage=all_array_storage,
             all_array_compression=all_array_compression,
-            auto_inline=auto_inline, pad_blocks=pad_blocks)
+            auto_inline=auto_inline, pad_blocks=pad_blocks,
+            use_image_hdu=use_image_hdu)
 
         self._hdulist.writeto(filename, *args, **kwargs)
 
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -22,7 +22,6 @@ try:
     from astropy.io import fits
     from astropy.io.fits.file import _File
     from astropy.io.fits.header import Header, _pad_length
-    from astropy.utils import lazyproperty
 except ImportError:
     raise ImportError("AsdfInFits requires astropy")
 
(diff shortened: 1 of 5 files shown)


asdf/fits_embed.py
```diff
@@ -235,22 +310,22 @@ def _update_asdf_extension(self, all_array_storage=None,
...
             auto_inline=auto_inline, pad_blocks=pad_blocks,
             include_block_index=False)
-        array = np.frombuffer(buff.getvalue(), np.uint8)
 
         try:
             asdf_extension = self._hdulist[ASDF_EXTENSION_NAME]
         except (KeyError, IndexError, AttributeError):
-            self._hdulist.append(fits.ImageHDU(array, name=ASDF_EXTENSION_NAME))
+            self._hdulist.append(self._create_hdu(buff, use_image_hdu))
         else:
```
jdavies-st: `array` is not defined here.

asdf/fits_embed.py
```diff
@@ -222,9 +289,17 @@ def open(cls, fd, uri=None, validate_checksums=False, extensions=None,
         return cls._open_asdf(self, buff, uri=uri, mode='r',
                               validate_checksums=validate_checksums)
 
+    def _create_hdu(self, buff, use_image_hdu):
+        # Allow writing to old-style ImageHDU for backwards compatibility
+        if use_image_hdu:
+            array = np.frombuffer(buff.getvalue(), np.uint8)
+            return fits.ImageHDU(array, name=ASDF_EXTENSION_NAME)
```
hbushouse: Is the use of the ASDF_EXTENSION_NAME value here going to cause the XTENSION keyword to still have a value of "ASDF", even though in this case we want it to be "IMAGE"? Or does "name" in this case set the value of the EXTNAME keyword?

drdavella: It's the latter: the `name` keyword argument will make `EXTNAME` be `ASDF`, but the `XTENSION` will still be `IMAGE` since we're using `ImageHDU`.

However, this is the backwards compatible case. In the new case, both `XTENSION` and `EXTNAME` will be `ASDF`.

asdf/fits_embed.py
drdavella: Good catch. This was an oversight. I'm not sure it will be quite as simple as replacing the deleted line, but I'll add a unit test to exercise this case and go from there.

asdf/fits_embed.py
hbushouse: 👍

jdavies-st: Howard, here's an example of a current DMS NIRISS rate file read in and written out under the new asdf code.  Compare the last HDU.
```
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     244   ()      
  1  SCI           1 ImageHDU        48   (2048, 2048)   float32   
  2  ERR           1 ImageHDU        10   (2048, 2048)   float32   
  3  DQ            1 ImageHDU        11   (2048, 2048)   int32 (rescales to uint32)   
  4  VAR_POISSON    1 ImageHDU         9   (2048, 2048)   float32   
  5  VAR_RNOISE    1 ImageHDU         9   (2048, 2048)   float32   
  6  ASDF          1 ImageHDU         7   (5998,)   uint8   
---------------------------------------------------------------
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     244   ()      
  1  SCI           1 ImageHDU        48   (2048, 2048)   float32   
  2  ERR           1 ImageHDU        10   (2048, 2048)   float32   
  3  DQ            1 ImageHDU        11   (2048, 2048)   int32 (rescales to uint32)   
  4  VAR_POISSON    1 ImageHDU         9   (2048, 2048)   float32   
  5  VAR_RNOISE    1 ImageHDU         9   (2048, 2048)   float32   
  6  ASDF          1 NonstandardExtHDU      8   ()      
```

jdavies-st:
And it looks like there's really only one change to a test in `jwst` package that needs to be done to be compatible with this PR, and that's only because the test explicitly checks to see if the ASDF extension is an ImageHDU, something that is an ASDF implementation detail, and really shouldn't be tested on the JWST side anyway.  So that particular `assert` in the test can be removed now to pave the way for this change.

drdavella:
I just want to mention that this PR also explicitly "un-implements" the `.update` method of `AsdfInFits` objects. I don't think it was working previously, but I have explicitly added a `NotImplementedError`. I'm not sure this is worth supporting unless I get strong pushback from the JWST folks.

eteq:
I was trying to test this out and I ran into a couple of oddities while doing so.  It's not clear that it's related to these changes, but it's preventing me from testing this.  Specifically, I follow the example from http://asdf.readthedocs.io/en/latest/asdf/examples.html#saving-asdf-in-fits (which looks to be unchanged in this PR) and I get a ``ValueError: Don't know how to write out ASDF version 1.2.0``.  Is that expected?  I also tried doing the penultimate line as ``ff = fits_embed.AsdfInFits(hdulist, tree, version='1.1')`` (or any other version) with no luck...

(What I was actually trying to do was look at the generated fits files *without* asdf installed to see how things behave.  But can't do that without an example.)

asdf/fits_embed.py
hbushouse: I assume the lack of Dimensions and Format info for the new ASDF extension is due to the fact that fits.info no longer knows how to handle this type of extension?

drdavella:
@eteq ASDF can't be used from the source directory. Try installing the package and doing the same thing somewhere else. I can try to add a better error message for this for the user but I don't think we're going to support this for a variety of reasons that can be discussed elsewhere.

drdavella:
@hbushouse I just pushed an update that will allow `astropy.io.fits` to recognize the custom extension, so I think the dimension and format information would be available if this is run again with the latest updates.

cc @jdavies-st

eteq:
FWIW, it turns out the problem I had above was that I hadn't updated `git submodule update`ed... so not related.

Now that I got it working, I have an observation of the behavior that I think might be a bit surprising.  Supposing you have a file `embedded_asdf.fits` that's got asdf in the final HDU, here's the behavior:
```python
>>> from astropy.io import fits
>>> fits.open('embedded_asdf.fits')[-1]
<astropy.io.fits.hdu.base.NonstandardExtHDU at 0x10cd6df98>
>>> import asdf
>>> fits.open('embedded_asdf.fits')[-1]
<astropy.io.fits.hdu.base.NonstandardExtHDU at 0x111788780>
>>> from asdf import fits_embed
>>> fits.open('embedded_asdf.fits')[-1]
<asdf.fits_embed._AsdfHDU at 0x111793860>
```

So the class that comes out depends in whether `asdf.fits_embed` is imported, and I think that would surprise most users. So perhaps the registration should be in the root of `asdf` instead of in `fits_embed`?  To go even further one could imagine having a note (a `COMMENT`?) somewhere in the header that explicitly says "if you're using python, you should import `asdf`" or something like that?  Otherwise a novice user might see the first case and just be lost from the beginning.

drdavella:
@eteq very good point, will push an update shortly.

asdf/fits_embed.py
```diff
@@ -20,15 +20,87 @@
...
+
+        cards = [
+            ('XTENSION', cls._extension, 'ASDF extension'),
+            ('BITPIX', 8, 'array data type'),
+            ('NAXIS', 1, 'number of array dimensions'),
+            ('NAXIS1', len(buff.getvalue()), 'Axis length'),
+            ('PCOUNT', 0, 'number of parameters'),
+            ('GCOUNT', 1, 'number of groups'),
+            ('COMPRESS', compress, 'Uses gzip compression'),
+            ('EXTNAME', cls._extension, 'Name of ASDF extension'),
```
eteq: If I read this right, the *name* is still hard-coded as `ASDF`.  Is that desirable?  I would have thought this should allow multiple ASDF-type extensions, with arbitrary names.  Or is this necessary for backwards compatibility?

drdavella: Admittedly this is not the most generic design. However, since the `AsdfInFits` class is expected to be the only user of this HDU class, and since `astropy.io.fits` keys HDUs in an `HDUList` based on `EXTNAME`, it just made sense to set `EXTNAME` to `ASDF` here. In other words, this is a matter of convenience and an implementation detail of `AsdfInFits`, but not necessarily part of the abstract ASDF HDU convention itself.

hbushouse: Multiple ASDF extensions could be distinguished via unique EXTVER values, assuming that keyword makes sense in an ASDF extension?

drdavella: It might be possible, but I would need to investigate more. Currently `AsdfInFits` expects only a single ASDF extension to be present in a FITS file. Until the JWST pipeline has a use case for multiple ASDF extensions in the same file, I don't think it makes sense to support this.

eteq: Ah, well, I'm not necessarily attached to this *specific* line as the spot to change it.  Are you just thinking that if I am a user and want to have an ASDF-in-FITS HDU with the name "WCS_MODEL", I'd be expected to do something like
```
fits_embed.AsdfInFits(hdulist, tree)
hdulist['ASDF'].header['EXTNAME] = 'WCS_MODEL'
```
?  I'm really just saying I think it would be useful to have the option to do something like ``fits_embed.AsdfInFits(hdulist, tree, name='WCS_MODEL')``, as the typical user is likely to see that as the main place where they set information about the hdu.

eteq: Oh, I think I just realized my mis-conception.  I thought it was possible to have *multiple* asdf trees in one file. I see now that the "reader" mode only supports a single one.  I guess that makes this use case less important.  Probably I should make a *separate* issue about that, but this concern about the naming could wait until then...

drdavella: Just to be clear: the ASDF HDU is used solely for storing a serialized version of the ASDF metadata tree. `AsdfInFits` currently expects a one-to-one mapping between FITS files and ASDF trees. The actual data arrays, including any used for WCS models or any other purposes, are actually stored in conventional FITS HDUs. The details of this are controlled completely by the user.

So it doesn't really make sense to use anything other than `ASDF` as the `EXTNAME` for this HDU because the only thing that will ever actually process it is our `AsdfInFits` code. Its just an implementation detail and anyone who reads the pure FITS file does not need to be concerned with it. Unless there becomes a compelling reason to have more than one ASDF metadata tree in the same FITS file, I don't think it makes sense to change anything.

eteq: Gotcha, that clears things up.  I withdraw my request 😉

drdavella:
@hbushouse, @jdavies-st, @bernie-simon, @nden I'll wait for your call on when this is okay to merge. Based on discussions with @jdavies-st it doesn't sound like this change is necessarily going to be catastrophic for the JWST regression suite, but I want to make sure everyone is ready for it when it gets merged.

drdavella:
According to @nden there is a DMS delivery this week, so we will wait to merge until next week.

drdavella:
@nden just wondering if merging this would be a good way to ring in the new year?

hbushouse:
@drdavella I would say go for it. Everything is pretty stable right now and has been since we did the tag/branch for DMS just before Christmas. No other major changes have been merged lately.

PR #417, opened 2017-12-20
move wcs tags to gwcs

nden:
Adjusted tag versions after moving WCS schemas to gwcs.
spacetelescope/gwcs#122 and spacetelescope/asdf-standard#151 are related.


drdavella:
@nden I'm seeing some weird issues here so I'm going to have to dig deeper before we can move ahead with the merge.

nden:
@drdavella That's fine. I'm sitll working on tests on the other PR.

Issue #414, opened 2017-12-15
Developer usability concerns


drdavella:
There are a few issues that bite new developers (and, speaking from experience, even some more experienced ones) when trying to get started with ASDF and building from source:
* making sure the `asdf-standard` submodule is initialized and up-to-date
* the fact that it is not possible to import and use `asdf` from the source directory

It would be nice to address these issues somehow so that new developers are not put off by the project.

Cadair:
What get's me is the fact that doing `pip install -e .` doesn't work, i.e. even trival changes to the python source are not reflected in the imported version. This is probably the same as being able to import it from the source directory.

drdavella:
@Cadair this should be fixed on the latest master. The problem appears to have been the use of `use_2to3` in `setup.py`, which was removed in #409.

drdavella:
Closing since the major issues here have been addressed. A solution for making sure `asdf-standard` is always up to date exists in #419, but I am hesitant to merge this due to potential security issues.

PR #420, opened 2018-01-02
Allow import and use of module from source directory and development build

drdavella:
This is a relatively simple fix that allows the package to be imported and used from the source directory. It also is necessary to support builds in development mode, which was only recently supported in #409.

This partially addresses #414.


drdavella:
Test failure on travis is unrelated.

PR #424, opened 2018-01-08
Add a way of customising the comparison used in assert_roundtrip_tree

Cadair:
Happy new year!

### DIFF
--- a/asdf/tests/helpers.py
+++ b/asdf/tests/helpers.py
@@ -117,7 +117,7 @@ def assert_tree_match(old_tree, new_tree, ctx=None,
     recurse(old_tree, new_tree)
 
 
-def assert_roundtrip_tree(tree, tmpdir, asdf_check_func=None,
+def assert_roundtrip_tree(tree, tmpdir, *, asdf_check_func=None,
                           raw_yaml_check_func=None, write_options={}, extensions=None,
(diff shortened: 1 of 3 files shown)


asdf/tests/helpers.py
```diff
@@ -113,8 +118,8 @@ def recurse(old, new):
 
 
 def assert_roundtrip_tree(
-        tree, tmpdir, asdf_check_func=None, raw_yaml_check_func=None,
-        write_options={}, extensions=None):
+        tree, tmpdir, asdf_check_func=None, tree_match_func='assert_equal',
```
Cadair: I feel that `asdf_check_func` and `tree_match_func` are a little redundant here. Do you think it makes sense to remove `asdf_check_func`?

drdavella: It does seem redundant. We'll have to go through and update all of the test cases that currently make use of it.

Cadair:
@drdavella I tried to remove the `asdf_check_func` argument, but the key difference is that it takes a asdf file object not the trees. So I think we need both. I did go through and change all uses of `assert_rountrip_tree` to use kwargs. I can remove the `*,` if you don't want to enforce that, but I thought it was a better idea in the code.

drdavella:
Thanks @Cadair!

PR #425, opened 2018-01-10
Create pytest plugin for testing schema files

### LINKED ISSUE #423: Provide API for allowing other packages to validate schemas and examples
Now that other packages are providing their own ASDF schemas (e.g. `astropy` and `gwcs`), it would be useful if they had the ability to test those schemas in the same way that ASDF currently tests its own schemas. This should include making sure that the schemas themselves are valid, as well as testing any examples provided in the schemas.

drdavella:
This PR introduces a `pytest` plugin that is used for testing schema files. The intent is for the plugin to be used by other packages as well for testing any ASDF schemas that they might provide.

This resolves #423.


nden:
@drdavella  Is there an example how to use this from other packages?

drdavella:
@nden, no I should add this to the documentation. In the meantime, it should just be necessary to add this to your `conftest.py`:
```python
pytest_plugins = [
    'asdf.tests.schema_tester'
]
```

PR #427, opened 2018-01-15
TemporalFrame and FrameType uses astropy tag

Cadair:
So this is a quick hack to see if I could make my solar gwcs seralise with my changes in astropy/astropy#7079. I appreciate that schema moves are going on and such, I don't expect this to get merged.

I am opening it because I have some questions that I will put inline.

### DIFF
--- a/asdf/tags/wcs/wcs.py
+++ b/asdf/tags/wcs/wcs.py
@@ -65,106 +65,9 @@ class StepType(dict, AsdfType):
 
 class FrameType(AsdfType):
     name = "wcs/frame"
-    requires = ['gwcs', 'astropy-1.3.3']
+    requires = ['gwcs']
     types = ['gwcs.Frame2D']
     version = '1.2.0'
-
-    import astropy
-    _astropy_version = astropy.__version__
-    # This indicates that Cartesian Differential is not available
-    _old_astropy = astropy.__version__ <= '1.3.3'
-
-    @classmethod
-    def _get_reference_frame_mapping(cls):
-        if hasattr(cls, '_reference_frame_mapping'):
-            return cls._reference_frame_mapping
-
-        from astropy.coordinates import builtin_frames
-
-        cls._reference_frame_mapping = {
-            'ICRS': builtin_frames.ICRS,
-            'FK5': builtin_frames.FK5,
-            'FK4': builtin_frames.FK4,
-            'FK4_noeterms': builtin_frames.FK4NoETerms,
-            'galactic': builtin_frames.Galactic,
-            'galactocentric': builtin_frames.Galactocentric,
-            'GCRS': builtin_frames.GCRS,
-            'CIRS': builtin_frames.CIRS,
-            'ITRS': builtin_frames.ITRS,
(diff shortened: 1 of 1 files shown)


asdf/tags/wcs/wcs.py
```diff
@@ -65,107 +65,10 @@ class StepType(dict, AsdfType):
...
-        reference_frame_mapping = cls._get_reference_frame_mapping()
-
-        cls._inverse_reference_frame_mapping = {}
-        for key, val in reference_frame_mapping.items():
-            cls._inverse_reference_frame_mapping[val] = key
-
-        return cls._inverse_reference_frame_mapping
-
-    @classmethod
-    def _reference_frame_from_tree(cls, node, ctx):
```
Cadair: So, as far as my limited understanding of all this can tell, all of this *should* be able to be deleted once astropy/astropy#7079 is in? (With the priviso of needing to support all the representation types in that PR?) is that base assumption correct @nden ?

asdf/tags/wcs/wcs.py
```diff
@@ -343,54 +273,54 @@ def assert_equal(cls, old, new):
         for old_frame, new_frame in zip(old.frames, new.frames):
             helpers.assert_tree_match(old_frame, new_frame)
 
-class ICRSCoord(AsdfType):
```
Cadair: This conflicts with the latest frame generalisation changes in astropy/astropy#7079, in that asdf will use this and not the `BaseCoordinateFrame` schema in astropy. I don't know how to fix that other than remove it from here?

drdavella: Ideally I'd like to keep this tag in asdf for backwards compatibility, although it's not critical in this case.

A possible (but maybe not ideal) solution would be to make a special case of `ICRSCoord` and explicitly provide an implementation in Astropy as well. This should guarantee that it gets overridden.

Otherwise, we might need to provide some kind of API that allows existing tags to be overridden by the kind of subclass implementation you are working on in Astropy. But I suspect the logic necessary to support that might get kind of hairy.

asdf/tags/wcs/wcs.py
drdavella: There's a question about whether we want to retain this here for backwards compatibility. In any case, these things are moving to the `gwcs` package, so it would be a question as to whether `gwcs` needs to continue to maintain them, or whether they can be dropped there.

(@nden please chime in if I'm wrong about this)

asdf/tags/wcs/wcs.py
Cadair: Can we just drop it for 2.0 😆

asdf/tags/wcs/wcs.py
nden: As @drdavella said this needs to stay for backwards compatibility (though asdf versioning needs to be defined better).

gwcs uses the low level builtin frames in coordinates and I don't see them implemented in the astropy PR but may be this is coming @Cadair? In the future I think gwcs will use astropy's schemas for frames and coordinates.

Cadair:
An alternate version of this is opened against gwcs.

PR #428, opened 2018-01-16
fix tag creation in schema tester

nden:
@drdavella This problem surfaced when using the schemas tester in gwcs. To find the tag the tester splits the path to the schema assuming 2 levels of nested directories (like asdf's schemas). With this change the tag is created dynamically.

### DIFF
--- a/asdf/tests/schema_tester.py
+++ b/asdf/tests/schema_tester.py
@@ -105,6 +105,7 @@ class AsdfSchemaExampleItem(pytest.Item):
     def _find_standard_version(self):
         filename = self.filename
         components = filename[filename.find('schemas') + 1:].split(os.path.sep)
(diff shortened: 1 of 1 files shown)


asdf/tests/schema_tester.py
```diff
@@ -105,6 +105,7 @@ def __init__(self, schema_path, parent, example):
     def _find_standard_version(self):
         filename = self.filename
         components = filename[filename.find('schemas') + 1:].split(os.path.sep)
+        tag = "tag:" + ":".join(["{}" for i in range(len(components[1:]))])
         tag = 'tag:{}:{}'.format(components[1], '/'.join(components[2:]))
```
drdavella: Should this line be deleted?

nden: Actually, this is not correct either. I'm going to close this.

PR #433, opened 2018-01-25
WIP: Fix for weird non-deterministic bug

### LINKED ISSUE #430: Intermittent failure loading external schemas/tags
This is a fun one. It's almost certainly due to expectations of dictionary ordering somewhere in the guts of the type index machinery since it only seems to be reproducible on Python 3.5.

The failures look something like the following:
https://travis-ci.org/astropy/astropy/jobs/332871390#L5044

This affects the movement of tags to Astropy and GWCS.

cc @Cadair

drdavella:
This seems to fix #430. I think it makes sense as a solution, but I need to do a bit more investigation/experimentation.

@Cadair, if you want to give this a spin on your own code that would be very helpful.

### DIFF
--- a/asdf/asdftypes.py
+++ b/asdf/asdftypes.py
@@ -111,7 +111,7 @@ class _AsdfWriteTypeIndex(object):
                     "Don't know how to write out ASDF version {0}".format(
                         self._version))
 
-            for name, _version in version_map['tags'].items():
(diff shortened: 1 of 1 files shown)


drdavella:
Closing since the correct solution has been pushed in #434.

PR #434, opened 2018-01-26
Fix bug that incorrectly overwrites subclasses in the write type index

### LINKED ISSUE #430: Intermittent failure loading external schemas/tags

drdavella:
**TL; DR**: This fixes #430.

### Background
Each ASDF tag implementation names the types that it serializes. The ASDF type index automatically uses each tag type to serialize not only the types explicitly named, but also any subclasses of those types. This is all well and good.

It is possible for multiple types within the same type hierarchy to each be serialized by different tags. For example, the class `Animal` might be serialized by `AnimalType`, but the class `Dog`, which is a subclass of `Animal`, should be serialized by `DogType`.

### The Problem
ASDF was not careful about the way that it overrode tags associated with a particular subclass. The result ended up being dependent on the order in which the tags were added to the index, which is non-deterministic on certain versions of Python. In concrete terms, sometimes `DogType` correctly was used to serialize `Dog`, but other times `AnimalType` was incorrectly used. It entirely depended on the order in which `AnimalType` and `DogType` were added to the type index, which is potentially non-deterministic since the tags are stored in dictionaries.

### The Solution
Make sure that ASDF does not overwrite tags associated with a given subclass if the tag serializes a class that is higher in the type hierarchy than the given subclass. In other words, don't use `AnimalType` to serialize `Dog` if there is already a `DogType` that serializes `Dog`.

### Real-World Example
In the particular case reported in #430, the problem was that `astropy.coordinates.distances.Distance` is a subclass of `astropy.units.Quantity`, which itself is a subclass of `numpy.ndarray`. The correct case is that `QuantityType` should be used to serialize `Distance`, but in some cases `NDArrayType` was used, incorrectly. This is what led to intermittent random failures, but only on Python 3.5. 

The fact that the tests were consistent on Python 3.6 is expected since dictionaries retain their order in CPython. However, the fact that the test always passed on Python 3.6 was just a coincidence. It could have just as easily been the case that the test always failed, in which case this probably would have been much easier to debug.

cc @Cadair

### DIFF
--- a/asdf/asdftypes.py
+++ b/asdf/asdftypes.py
@@ -80,14 +80,25 @@ class _AsdfWriteTypeIndex(object):
         self._type_by_cls = {}
         self._type_by_name = {}
         self._type_by_subclasses = {}
+        self._class_by_subclass = {}
         self._types_with_dynamic_subclasses = {}
+
+        def add_subclasses(typ, asdftype):
+            for subclass in util.iter_subclasses(typ):
+                # Do not overwrite the tag type for an existing subclass if the
+                # new tag serializes a class that is higher in the type
+                # hierarchy than the existing subclass.
+                if subclass in self._class_by_subclass:
+                    if issubclass(self._class_by_subclass[subclass], typ):
+                        continue
+                self._class_by_subclass[subclass] = typ
+                self._type_by_subclasses[subclass] = asdftype
 
         def add_type(asdftype):
             self._type_by_cls[asdftype] = asdftype
             for typ in asdftype.types:
                 self._type_by_cls[typ] = asdftype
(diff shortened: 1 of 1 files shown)


drdavella:
Test timeout/failure on OSX seems to be unrelated and can be addressed in a separate issue.

Issue #451, opened 2018-02-15
FITS hdu.info() no longer reports size information about ASDF extension


drdavella:
Previously when the ASDF metadata was stored in an `IMAGE` extension, running `fits.info()` would provide information about the size of the extension.

Now that we are using a non-standard FITS extension to store ASDF data, it seems like this information is not available. Presumably this is because Astropy's `io.fits` does not know how to interpret the extension, so it doesn't know how to extract length information.

It would be useful if there was some way to allow Astropy to determine the length of extension.

cc @hbushouse

philhodge:
It looks to me as if you can still use the BITPIX and NAXIS1 keywords to get the length of the data block.  I just ran a test with a file that had:

XTENSION= 'ASDF    '           / ASDF extension

NAXIS1 was 5760, with BITPIX = 8.  There were only 9 header keywords, so the header took one block, and 5760 is exactly two blocks, so the total size of the ASDF extension should be 3 * 2880 bytes.  I added up the sizes of the other HDUs (a primary HDU and one IMAGE extension) and added 3 * 2880 to that, and the total did agree with the actual file size of 50365440 bytes.

drdavella:
@philhodge thanks, that is useful information. I think the main issue comes down to whether we can convince `astropy.io.fits` to report this length information in a useful way (for example when using `.info()`. I'm a bit pessimistic since from Astropy's perspective the `ASDF` extension is non-standard, and so it doesn't know how to interpret the various keywords. My guess is that the only solution would be to ask Astropy to explicitly support the ASDF extension.

philhodge:
That's why I did the arithmetic, to check that the NAXIS1 keyword was correct, i.e. that the sum of the numbers of bytes agreed with `ls -l`.  The BITPIX and NAXIS* keywords have to be populated correctly so that a FITS reader can skip over an extension without having to read and interpret its contents.  I realize that the ASDF extension is the last one, but nothing prevents me from appending another extension of a different type.  `astropy.io.fits` must get the size of each extension by checking the number of header keywords and the values of the BITPIX and NAXIS* keywords, regardless of the type of the extension.  Why it doesn't do so for 'ASDF' I don't know, but there's no reason that it couldn't, even without supporting that extension type in any way other than assuming it follows the FITS standard for required keywords.

drdavella:
[...]

PR #452, opened 2018-02-15
NAXIS1 in AsdfHDU should reflect data length, not padded length

drdavella:
This partially addresses #451.


perrygreenfield:
Looks good, though I'm not sure what self.data is in this context (e.g., is it the shape of the asdf buffer or the implied ndarray object?

drdavella:
It is a field that comes from higher in the HDU hierarchy in Astropy. You are correct that it is the ndarray object representing the underlying data. This is the same way that `ImageHDU` and other HDU objects in Astropy represent the field in question in the summary.

PR #454, opened 2018-02-16
Test against Astropy v3.0 by default

drdavella:
Now that `astropy-3.0` is available, run CI tests against the stable version by default. Also added test case for development version.


drdavella:
Actually tests won't pass until https://github.com/astropy/ci-helpers/pull/271 is merged.

PR #470, opened 2018-03-05
Update CI config to account for astropy and gwcs dev versions

drdavella:


drdavella:
Merging despite failing tests. Test failures are caused by recent updates to `astropy` and `gwcs` and will be addressed in a separate PR.

PR #473, opened 2018-03-09
Update asdf-standard: support for units in transform schemas

drdavella:
See https://github.com/spacetelescope/asdf-standard/pull/156.

cc @Cadair


Cadair:
:tada:

PR #480, opened 2018-03-27
Add command to asdftool for removing ASDF extension from FITS file

drdavella:
Specifically for @hbushouse. Given a file `asdf.fits` that contains an ASDF extension, here's an example of how to use it:

```
$ asdftool remove-hdu asdf.fits no_asdf.fits
```
The file `no_asdf.fits` will be created (assuming it doesn't already exist) as a duplicate of the input file but without the ASDF extension.

For usage information:
```
$ asdftool -h
$ asdftool remove-hdu -h
```


drdavella:
This will need to be rebased once #482 is merged.

PR #483, opened 2018-03-29
Use XTENSION=FOREIGN for ASDF-in-FITS

drdavella:
Use the `FOREIGN` keyword in ASDF-in-FITS HDUs purely in an effort to pass `fitsverify`. 

The new HDU does not actually conform to the `FOREIGN` convention as described [here](https://fits.gsfc.nasa.gov/registry/foreign.html). However, the `fitsverify` program provided by NASA's FITS office appears to validate HDUs with the `FOREIGN` keyword if they otherwise conform to the `IMAGE` standard. 

To make things even more confusing, `fitsverify` does NOT appear to validate an HDU that actually conforms to the `FOREIGN` convention. It also does not validate an HDU that uses `XTENSION=ASDF`. We hope to eventually sort things out with the FITS office, but in the meantime, we want users of `fitsverify` to be able to validate our files.

If you need to convince yourself that this actually works now, open a Python interpreter and try the following:
```python
from asdf.tests.test_fits_embed import create_asdf_in_fits
asdf_in_fits = create_asdf_in_fits()
asdf_in_fits.write_to('asdf.fits')
```

You can open the new file `asdf.fits` to confirm that there is in fact an HDU with `XTENSION=FOREIGN`, and it holds the contents of the ASDF file. You can also run `fitsverify` against this file to confirm that it passes validation.

cc @stscicrawford

### DIFF
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -37,7 +37,17 @@ class _AsdfHDU(fits.hdu.base.NonstandardExtHDU):
     an XTENSION keyword) of ASDF.
     """
 
-    _extension = ASDF_EXTENSION_NAME
+    # Use the FOREIGN keyword here purely in an effort to pass fitsverify.
+    # This HDU does not actually conform to the FOREIGN convention as described
+    # at https://fits.gsfc.nasa.gov/registry/foreign.html. However, the
+    # fitsverify program provided by NASA's FITS office appears to validate
+    # HDUs with the FOREIGN keyword if they otherwise conform to the IMAGE
+    # standard. To make things even more confusing, fitsverify does NOT appear
+    # to validate an HDU that actually conforms to the FOREIGN convention. It
+    # also does not validate an HDU that usese XTENSION=ASDF. We hope to
+    # eventually sort things out with the FITS office, but in the meantime, we
+    # want users of fitsverify to be able to validate our files.
+    _extension = 'FOREIGN'
 
     @classmethod
     def from_buff(cls, buff, compress=False, **kwargs):
@@ -71,7 +81,7 @@ class _AsdfHDU(fits.hdu.base.NonstandardExtHDU):
             ('PCOUNT', 0, 'number of parameters'),
             ('GCOUNT', 1, 'number of groups'),
             ('COMPRESS', compress, 'Uses gzip compression'),
-            ('EXTNAME', cls._extension, 'Name of ASDF extension'),
+            ('EXTNAME', 'ASDF', 'Name of ASDF extension'),
         ]
 
         header = Header(cards)


drdavella:
If we move forward with this change we will need to update the change log.

stscicrawford:
Thanks @drdavella -- this looks good to me especially with the extensive note in the documentation.  If this is merged, an issue should be created to track this as well.

drdavella:
@stscicrawford, @hbushouse after several weeks and several emails with no response from NASA's FITS office, should I go ahead and merge this?

hbushouse:
Is it the case that this change is still creating an extension that does NOT actually conform to the official `FOREIGN` convention, but instead creates one that will pass `fitsverify`? If so, I'm not comfortable with forcing our code to do something wrong just so it matches something else that's equally wrong. We should be doing the right thing and telling folks that `fitsverify` is not to be used, because it's wrong.

stscicrawford:
I agree that I do not like this as a long term solution either.   An issue open about this definitely needs to be kept open if this is accepted and it should be regularly revisited, but considering how widely used CFITSIO is, I do not think it is a realistic option to ask people to stop using it. 

As completely alternative thought, would it be possible to store the ASDF specific stuff inside a FITS table?   This might be a little awkward but it would get around this problem of using the Foreign extension.

perrygreenfield:
Like how putting into a table? If just stuffing it as one big binary item, we have the same issue as using the image format don't we? It's not really a table then (or whatever kind of column item it becomes).

stscicrawford:
@perrygreenfield Yes it  was just a thought -- I didn't know if it was stored as a table (even a 1x1 table with a big binary blob) that might be a more 'FITS-like' solution.   Then again, I also don't remember from the FITS standards if binary blobs can be stored in FITS tables, so this might all moot.

perrygreenfield:
Has anyone tried contacting Bill Pence directly? (He's shown as emeritus, so he may be still active with FITSIO) William.D.Pence@nasa.gov

perrygreenfield:
@stscicrawford You are probably right expect weirder things from tables than images. Column name of ASDF_CONTENT or something like that provides some clue.

stscicrawford:
Okay, I'm gong to recommend merging this so it fixes the issues faced by the other groups and we have a mostly compliant FITS file and then opening a follow-up issue for how to handle the third version of this.

drdavella:
@perrygreenfield I have not contacted him directly: I just tried using the FITS email alias. I can give him a try though, unless you want to.

perrygreenfield:
@drdavella which email address did you use specifically? I ask since getting cfitsio to adopt the changes is more important than the FITS office specifically.

drdavella:
I got fits@fits.gsfc.nasa.gov from https://fits.gsfc.nasa.gov/.

perrygreenfield:
try ccfits@heasarc.gsfc.nasa.gov

drdavella:
Okay I'll give it a try. @stscicrawford should I still merge or hold off to see if I get a response?

stscicrawford:
Okay, give it another couple of days to see if you hear a response from them.

drdavella:
Closing since this does not seem to be the right solution. Discussion will continue at https://github.com/STScI-JWST/jwst/issues/1807.

PR #489, opened 2018-04-12
Update asdf-standard to reflect new multiply schema

drdavella:
See https://github.com/spacetelescope/asdf-standard/pull/160.


drdavella:
Tests won't pass until https://github.com/astropy/astropy/pull/7210 is merged.

PR #490, opened 2018-04-15
Make the build reproducible

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed that `asdf` could not be built reproducibly.

This is due to it including nondeterminstically random output as well as the current build time in the documentation.

This was originally filed in Debian as #895737 [1].

 [0] https://reproducible-builds.org/
 [1] https://bugs.debian.org/895737

Signed-off-by: Chris Lamb <lamby@debian.org>

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -3,6 +3,8 @@
 
 
 import io
+import os
+import time
 import re
 import copy
 import datetime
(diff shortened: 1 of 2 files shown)


drdavella:
Thanks @lamby!

PR #493, opened 2018-04-18
Use BINTABLE HDU for storing ASDF-in-FITS

drdavella:
This is intended to resolve https://github.com/STScI-JWST/jwst/issues/1807. After contacting NASA's FITS office, it seems like we should avoid the use of `FOREIGN` since it is an old convention and is not well supported. In fact, it does not seem to be properly validated by `fitsverify`.

cc @stscicrawford, @hbushouse, @perrygreenfield

### DIFF
--- a/asdf/__init__.py
+++ b/asdf/__init__.py
@@ -46,12 +46,4 @@ from jsonschema import ValidationError
 class ValidationError(ValidationError):
     pass
 
-try:
-    from astropy.io import fits
-except ImportError:
-    pass
-else:
-    from .fits_embed import _AsdfHDU
-    fits.register_hdu(_AsdfHDU)
-
 open = AsdfFile.open
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -17,8 +17,7 @@ from . import generic_io
 
 try:
     from astropy.io import fits
-    from astropy.io.fits.file import _File
-    from astropy.io.fits.header import Header, _pad_length
+    from astropy.io.fits import Column, BinTableHDU
 except ImportError:
     raise ImportError("AsdfInFits requires astropy")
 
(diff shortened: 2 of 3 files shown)


stscicrawford:
This looks good to me -- have you checked how fitsverify and astropy.verify handle this file?   Also, any chance send an example file to those who might be interested to see if they have any comments?

drdavella:
Github won't let me attach a FITS file here, but here's a recipe for creating a simple example (you'll need to check out this branch in order for it to work):
```
from asdf.tests.test_fits_embed import create_asdf_in_fits

aif = create_asdf_in_fits()
aif.write_to('example.fits')
```

The resulting file will contain a `BINTABLE` HDU with the ASDF metadata. It will verify correctly with `fitsverify` (find the online version [here](https://fits.gsfc.nasa.gov/fits_verify.html)). It will also open fine using `astropy.io.fits`.

Strangely enough, it won't pass verification with `fitscheck` from Astropy. However, I tested `fitscheck` against some other random test files (including some provided as test data by Astropy), and it seems to be having problems with them as well.

stscicrawford:
What error are you getting?  `Fitscheck` is for verifying and updating the checksum which is different than verifying the file.   So I'm not sure if that is the appropriate test unless the pipeline is using fitscheck, and if that is a requirement, then adding a checksum should be part of writing out the file. 

If I run verify on the file in astropy, it seems to pass:

    >>> from astropy.io import fits
    >>> h = fits.open('example.fits')
    >>> h.verify('exception')

That does not seem to raise any problems, so this does look to be working okay.   It might be useful to add that as a test (although I know that might add astropy as a testing dependency) so that we do not run into this problem again in the future with any updates.

drdavella:
Ah okay, my mistake. It's no problem to add a test: we already have a testing dependency on `astropy`.

PR #496, opened 2018-04-19
Documentation improvements based on feedback

drdavella:
@perrygreenfield I haven't yet found anything about making the code blocks larger, but I'll keep digging.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,9 @@
+div.highlight {
+    transition: width 0.5s;
+    -webkit-transition: width 0.5s;
+    overflow: hidden;
+}
+
+div.highlight:hover {
+    width: 150%
+}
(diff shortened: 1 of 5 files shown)


perrygreenfield:
That's weird. One would think that skinny code blocks would be an issue for many. Is it a sphinx thing?

perrygreenfield:
On further experimentation, they do get wider, but only in proportion to the width of the whole page. And there is a whole lot of empty space on the right in doing so. Is there a way to get rid of that empty margin, or make it a lot smaller?

perrygreenfield:
Unless there is some TeX'y way of giving the box and text all the elasticity for width changes (i.e., assign all increases to this item and not sidebar or margin). Otherwise that will have to do.

drdavella:
I'm going to look into it a little more, but that will be my fallback. I think I can provide custom CSS settings, but I'm not sure how much control that will really give me.

drdavella:
@perrygreenfield I found a cool CSS hack that will allow the code blocks to automatically expand when you hover over them. I like this a bit better than universally making the margins wider because I think it helps preserve the readability of the rest of the text. Let me know what you think.

PR #501, opened 2018-05-06
Add API function for retrieving history entries

### LINKED ISSUE #500: Abstract history reader interface
On read-the-docs there's a clear example of how to add/write ASDF history.   I could not see how to read one back.  Is there any kind of API for reading history,  say into an ordered list of tuples similar to:

[ (date, ... description ...),
  ...
]

Having an abstract API could insulate callers from changes in the underlying tree format.   Does an API exist?

drdavella:
This fixes #500.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 2.1.0 (unreleased)
 ------------------
 
+- Add API function for retrieving history entries. [#501]
+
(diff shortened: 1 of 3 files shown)


bernie-simon:
I looked at get_history_entries and it had the problem that if the history was not yet instantiated, it returned a new empty list. Appending to a new list will not add the appended value to the history.

Issue #499, opened 2018-04-21
[2.0.0] AttributeErrors during build time tests


olebole:
Additionally to #498, there are a number of `AttributeError`s:

```________________________ test_missing_extension_warning ________________________

    def test_missing_extension_warning():
    
        yaml = """
    history:
      extensions:
        - !core/extension_metadata-1.0.0
          extension_class: foo.bar.FooBar
          software: !core/software-1.0.0
            name: foo
            version: 1.2.3
        """
    
        buff = yaml_to_asdf(yaml)
        with pytest.warns(None) as warnings:
>           with asdf.open(buff) as af:

asdf/tags/core/tests/test_history.py:136: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
asdf/asdf.py:766: in open
    ignore_missing_extensions=ignore_missing_extensions)
asdf/asdf.py:678: in _open_impl
    ignore_missing_extensions=ignore_missing_extensions)
asdf/asdf.py:638: in _open_asdf
    self._check_extensions(tree, strict=strict_extension_check)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <asdf.asdf.AsdfFile object at 0x7f3748917358>
tree = {'history': {'extensions': [{'extension_class': 'foo.bar.FooBar', 'software': {'name': 'foo', 'version': '1.2.3'}}]}}
strict = False

    def _check_extensions(self, tree, strict=False):
        if 'history' not in tree or not isinstance(tree['history'], dict) or \
                'extensions' not in tree['history']:
            return
    
        for extension in tree['history']['extensions']:
            filename = "'{}' ".format(self._fname) if self._fname else ''
>           if extension.extension_class not in self._extension_metadata:
E           AttributeError: 'TaggedDict' object has no attribute 'extension_class'

asdf/asdf.py:151: AttributeError
```

Another one:

```
_____________________________ test_mask_arbitrary ______________________________

    def test_mask_arbitrary():
        content = """
        arr: !core/ndarray-1.0.0
          data: [[1, 2, 3, 1234], [5, 6, 7, 8]]
          mask: 1234
        """
    
        buff = helpers.yaml_to_asdf(content)
[...]

Issue #498, opened 2018-04-21
[2.0.0] Tests fail with RepresenterError during build


olebole:
Hi,

When running the tests during the build of the package, many of them fail with a `RepresenterError`:

```
   dh_auto_test -O--buildsystem=pybuild
I: pybuild base:217: cd /build/python-asdf-2.0.0/.pybuild/cpython3_3.6_asdf/build; python3.6 -m pytest 
============================= test session starts ==============================
platform linux -- Python 3.6.5, pytest-3.3.2, py-1.5.3, pluggy-0.6.0
rootdir: /build/python-asdf-2.0.0, inifile: setup.cfg
plugins: remotedata-0.2.1, openfiles-0.3.0, doctestplus-0.1.3, arraydiff-0.2
collected 272 items / 2 skipped

asdf/asdf.py .                                                           [  0%]
asdf/stream.py F                                                         [  0%]
asdf/util.py ...                                                         [  1%]
asdf/commands/tests/test_defragment.py FFs                               [  2%]
[...]
=================================== FAILURES ===================================
_________________________ [doctest] asdf.stream.Stream _________________________
012     Examples
013     --------
014     Save a double-precision array with 1024 columns, one row at a
015     time::
016 
017          >>> from asdf import AsdfFile, Stream
018          >>> import numpy as np
019          >>> ff = AsdfFile()
020          >>> ff.tree['streamed'] = Stream([1024], np.float64)
021          >>> with open('test.asdf', 'wb') as fd:
UNEXPECTED EXCEPTION: RepresenterError("cannot represent an object: {'streamed': Stream([1024], float64, strides=None), 'history': {'extensions': []}, 'asdf_library': {'
Traceback (most recent call last):

  File "/usr/lib/python3.6/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)

  File "<doctest asdf.stream.Stream[4]>", line 2, in <module>

  File "/build/python-asdf-2.0.0/.pybuild/cpython3_3.6_asdf/build/asdf/asdf.py", line 1059, in write_to
    self._serial_write(fd, pad_blocks, include_block_index)

  File "/build/python-asdf-2.0.0/.pybuild/cpython3_3.6_asdf/build/asdf/asdf.py", line 820, in _serial_write
    self._write_tree(self._tree, fd, pad_blocks)

  File "/build/python-asdf-2.0.0/.pybuild/cpython3_3.6_asdf/build/asdf/asdf.py", line 781, in _write_tree
[...]

PR #502, opened 2018-05-06
Allow test suite to run even when package is not installed

drdavella:
This allows the package to be tested in environments where it can't first be installed (e.g. the Debian build environment).

In order to make this work, simply define `ASDF_TEST_BUILD` before running `pytest`. 

@olebole this should solve the issue you reported in #498 and #499. I hope to release this in bugfix v2.0.1 sometime this week.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@
 
 - Add API function for retrieving history entries. [#501]
 
+2.0.1 (unreleased)
+------------------
+
+- Allow test suite to run even when package is not installed. [#502]
+
 2.0.0 (2018-04-19)
 ------------------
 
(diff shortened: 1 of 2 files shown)


olebole:
@drdavella You are great! Thank you very much for your efforts! This will help me a lot! :+1:

drdavella:
@olebole I just want to mention that the schema tests do not run when running the tests this way, but I don't think it's a big deal since those are mostly interesting to developers.

drdavella:
@olebole I just released v2.0.1. It looks like I was wrong about the schema tests: they should work for you. Just to be clear, you will need to set the environment variable `ASDF_TEST_BUILD` to some value like `1` or `True` in order for it to work:

`$ ASDF_TEST_BUILD=1 pytest`

Also, you will need to have `gwcs` installed in order for all of the tests to pass.

PR #507, opened 2018-05-18
Allow serialization of broadcasted ndarrays

### LINKED ISSUE #506: Saving arrays from np.broadcast_arrays fails
This:
```python
def test_broadcasting(tmpdir):
   attrs = np.broadcast_arrays(np.array([10,20]), np.array(10), np.array(10))
   tree = {'one': attrs[0],
           'two': attrs[1],
           'three': attrs[2]}
   assert_roundtrip_tree(tree, tmpdir)
```

Fails with a 
```
E           jsonschema.exceptions.ValidationError: {'shape': [2], 'source': 1, 'datatype': 'int64', 'byteorder': 'little', 'strides': [0]} is not valid under any of the given schemas
```

(full traceback: https://gist.github.com/Cadair/9ff970a06c4f90bf409a86b4022505f8)

I am assuming it's a problem with 'strides'

drdavella:
The solution in this PR is simply to create a contiguous copy of the broadcasted array and save it to disk. While this uses more memory on disk than strictly required by `numpy`, we need to account for the fact that broadcasting is a concept that is specific to `numpy` and may not translate to other implementations of ASDF. For this reason, it seems to make the most sense to create a copy of the broadcasted array so that it is available for other implementations.

This fixes #506.

cc @Cadair

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@
 
 - Add API function for retrieving history entries. [#501]
 
+2.0.2 (unreleased)
+------------------
+
+- Allow serialization of broadcasted ``numpy`` arrays. [#507]
+
 2.0.1 (2018-05-08)
 ------------------
 
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -372,7 +372,7 @@ class GenericFile(object):
     """
 
     def write_array(self, array):
-        _array_tofile(None, self.write, array)
+        _array_tofile(None, self.write, np.ascontiguousarray(array))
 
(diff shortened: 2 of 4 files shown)


Cadair:
Are you checking for c contiguous because that's how you are saving it to disk? I mean vs f contiguous?

drdavella:
@Cadair I believe the answer is yes, but in truth I'm checking for `c_contiguous` here because that's what was being checked for prior to this change as well. This should actually probably be defined by the standard somewhere, or encoded in the schema, though.

PR #508, opened 2018-05-24
Fix wcs tests with incompatible version of Astropy and GWCS

drdavella:
These tests appear to fail when using the latest release of GWCS (`0.9.0`) and the latest release of Astropy (`3.0.2`). It appears to work fine with the latest development version of Astropy, however.

This PR simply skips the failing tests when the incompatible versions are present.


nden:
That's unfortunate. Let me try to find a working version.

drdavella:
I'm just going to merge this for now as a workaround. I'm not sure these tests even belong here anymore since this functionality has moved into other packages. I mostly was keeping it around to test backwards compatibility.

PR #526, opened 2018-08-03
Update complex number schema and add tests

drdavella:
This PR addresses several outstanding problems with the schema for complex numbers. It adds much more thorough tests of the schema and ensures that improperly formatted complex numbers are detected by a schema validation error (which was not previously the case).

In a way, this is a fairly substantial change since it imposes much more stringent requirements on the format of complex numbers. However, since these requirements were implicit to begin with and should have been enforced in the first place, it seems unlikely that anyone is using files that are in conflict with the new rules.

This is closely related to https://github.com/spacetelescope/asdf-standard/pull/171 and integrates the updates to the schema that were introduced there.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,12 @@
 
 - Store ASDF-in-FITS data inside a 1x1 BINTABLE HDU. [#519]
 
+2.0.3 (unreleased)
+------------------
+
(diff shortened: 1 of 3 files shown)


drdavella:
Added a change log entry since this change has the potential (however minimal) to affect the behavior of `asdf` when reading files with ill-formed complex numbers.

PR #533, opened 2018-08-16
Ensure when closing, to explicitly close data blocks

stscieisenhamer:
This is to close out references to memory-mapped arrays that,
in the current context, not associated with a file. The `Block._memmapped`
is `False` but the `.data` is a memory-mapped array.

WIP spacetelescope/jwst#2282

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 - Update asdf-standard to reflect more stringent (and, consequently, more
   correct) requirements on the formatting of complex numbers. [#526]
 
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
A bit of test code can be found in @stscieisenhamer [gist asdf_blocks.py](https://gist.github.com/stscieisenhamer/5afa786626cd9000bcf8fd6f9eab0634#file-asdf_blocks-py).

If run without this patch, the end result is 1 open file handle, which has no direct way of being closed.

With the patch, no file handles are left.

PR #534, opened 2018-08-22
Allow typing.NamedTuple to be serialized

vmarkovtsev:
This allows to serialize `typing.NamedTuple`-s and other crazy inheritors from `list` and `tuple` which do not preserve the original constructor signature.

We catch `TypeError` in the constructor and use the auxiliary list.

### DIFF
--- a/asdf/tests/test_yaml.py
+++ b/asdf/tests/test_yaml.py
@@ -96,8 +96,12 @@ def run_tuple_test(tree, tmpdir):
     def check_raw_yaml(content):
         assert b'tuple' not in content
 
+    # Ignore these warnings for the tests that don't actually test the warning
+    init_options = dict(ignore_implicit_conversion=True)
+
     helpers.assert_roundtrip_tree(tree, tmpdir, asdf_check_func=check_asdf,
-                                  raw_yaml_check_func=check_raw_yaml)
+                                  raw_yaml_check_func=check_raw_yaml,
+                                  init_options=init_options)
 
 
 def test_python_tuple(tmpdir):
@@ -145,7 +149,9 @@ def test_named_tuple_collections_recursive(tmpdir):
     def check_asdf(asdf):
         assert (asdf.tree['val'][2] == np.ones(3)).all()
 
-    helpers.assert_roundtrip_tree(tree, tmpdir, asdf_check_func=check_asdf)
+    init_options = dict(ignore_implicit_conversion=True)
+    helpers.assert_roundtrip_tree(tree, tmpdir, asdf_check_func=check_asdf,
+                                  init_options=init_options)
 
 
 def test_named_tuple_typing_recursive(tmpdir):
@@ -159,7 +165,25 @@ def test_named_tuple_typing_recursive(tmpdir):
     def check_asdf(asdf):
         assert (asdf.tree['val'][2] == np.ones(3)).all()
 
-    helpers.assert_roundtrip_tree(tree, tmpdir, asdf_check_func=check_asdf)
+    init_options = dict(ignore_implicit_conversion=True)
+    helpers.assert_roundtrip_tree(tree, tmpdir, asdf_check_func=check_asdf,
+                                  init_options=init_options)
+
+
+def test_implicit_conversion_warning():
+    nt = namedtuple("TestTupleWarning", ("one", "two", "three"))
+
+    tree = {
+        "val": nt(1, 2, np.ones(3))
+    }
+
+    with pytest.warns(UserWarning, match="Failed to serialize instance"):
+        with asdf.AsdfFile(tree) as af:
+            pass
+
+    with pytest.warns(None) as w:
+        with asdf.AsdfFile(tree, ignore_implicit_conversion=True) as af:
+            assert len(w) == 0
 
 
 def test_tags_removed_after_load(tmpdir):
--- a/asdf/tests/test_yaml.py
+++ b/asdf/tests/test_yaml.py
@@ -2,7 +2,8 @@
 # -*- coding: utf-8 -*-
 
 import io
-from collections import OrderedDict
+from collections import namedtuple, OrderedDict
+from typing import NamedTuple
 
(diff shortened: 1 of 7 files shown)


asdf/tests/test_yaml.py
```diff
@@ -97,14 +109,30 @@ def test_python_tuple(tmpdir):
         "val": (1, 2, 3)
     }
 
-    def check_asdf(asdf):
-        assert isinstance(asdf.tree['val'], list)
+    run_tuple_test(tree, tmpdir)
+                                  
+
```
drdavella: There is extraneous whitespace on these two lines.

asdf/tests/test_yaml.py
```diff
@@ -97,14 +109,30 @@ def test_python_tuple(tmpdir):
...
-                                  raw_yaml_check_func=check_raw_yaml)
+    tree = {
+        "val": nt(1, 2, 3)
+    }
+
+    run_tuple_test(tree, tmpdir)
+
+def test_named_tuple_typing(tmpdir):
+    # Ensure that we are able to serialize a typing.NamedTuple.
+
```
drdavella: There is extraneous whitespace on this line as well.

asdf/tests/test_yaml.py
```diff
@@ -97,14 +109,30 @@ def test_python_tuple(tmpdir):
         "val": (1, 2, 3)
     }
 
-    def check_asdf(asdf):
-        assert isinstance(asdf.tree['val'], list)
+    run_tuple_test(tree, tmpdir)
+                                  
+                                  
+def test_named_tuple_collections(tmpdir):
```
drdavella: Can you please add a test showing that recursive serialization of `namedtuple` members is handled correctly? For example, you could create a `namedtuple` where one of the members is an `np.ndarray`.

vmarkovtsev:
Added more tests and removed whitespaces.

vmarkovtsev:
@drdavella ping

drdavella:
@vmarkovtsev I'd like some input from @perrygreenfield before merging. I have some lingering concerns over whether this is okay since it does not preserve true round-trip behavior, although maybe it doesn't matter too much.

Just for the sake of argument: what would prevent a user from performing this conversion explicitly before passing to ASDF? I realize it's probably less convenient and/or elegant, but it would put the burden of information loss on the end user rather than requiring ASDF to strip away information implicitly under the hood.

vmarkovtsev:
Yep, I see the point. On the other side, the exact subtypes of `namedtuple` are always defined in the source code (`namedtuple` and `NamedTuple` are abstract) and it is impossible to recover them without saving the source code - thus we arrive at pickles. But the ultimate parent of those is `tuple` and it is possible to recover the value after serialization because the order of elements persists.

ASDF already strips items with `None` values btw.

drdavella:
> the exact subtypes of namedtuple are always defined in the source code (namedtuple and NamedTuple are abstract) and it is impossible to recover them without saving the source code

This seems like a design flaw to me, but there's not much we can do about it. Unfortunately it means we can't write a tag for these types, which would be the ideal solution.

> ASDF already strips items with None values btw.

Interesting point, although I think it's really a separate issue (e.g. #536 😉).

perrygreenfield:
I’ll take a look at this today (though I have github authentication issue due to getting a new phone…)

drdavella:
@vmarkovtsev maybe as a compromise for now we could add a top-level option like `implicit_type_conversion` to enable this behavior (defaults to `False`)? Also it would be useful to emit a warning when this occurs.

setup.cfg
```diff
@@ -17,7 +17,7 @@ open_files_ignore = test.fits asdf.fits
 # Account for both the astropy test runner case and the native pytest case
 asdf_schema_root = asdf-standard/schemas asdf/schemas
 asdf_schema_skip_names = asdf-schema-1.0.0 draft-01
-addopts = --doctest-rst
+#addopts = --doctest-rst
```
drdavella: I believe this was changed by mistake.

vmarkovtsev: Yep

vmarkovtsev:
How about `implicit_lists`? It is only used for those types which inherit from `list` or `tuple` and do not preserve the `__init__` signature. -1 to warnings because when the user sets this flag to `True` they know what they are doing.

perrygreenfield:
If we support that, why not implicit_dicts as well? I tend to agree with not requiring the warning. These mechanisms would be advertised as an easy way to save python info, but only in one direction (without the user reconstructing the objects outside of ASDF).

vmarkovtsev:
Do we have an example in the wild when this code fails in `treeutil.py`:

```python
if isinstance(tree, dict):
    result = tree.__class__()
    seen.add(id_tree)
    for key, val in tree.items():
        val = recurse(val)
        if val is not None:
            result[key] = val
    seen.remove(id_tree)
```

All dicts have `items()` and `__setitem__`, otherwise something is really screwed up on the user's side.
I guess there will be no problems afterwards...

drdavella:
Sorry to waffle, but here's one more proposal: perform the conversion by default, but with a warning, and add a top-level option to suppress the warnings like `ignore_implicit_conversion=False` or some such. 

In this case I like the generality of "implicit conversion" vs "implicit list" or "implicit dict" since it lets us overlook the `dict` case and other possible cases for now (although I agree with @vmarkovtsev that `namedtuple` is probably fairly unique).

This also parallels existing top-level options `ignore_version_mismatch` and `ignore_unrecognized_tag`. (Btw, eventually it might be nice to have some other configuration mechanism for these kinds of options).

vmarkovtsev:
I like this proposal.

drdavella:
@vmarkovtsev cool. I can finish up this PR if you want, as long as I'm able to push to your branch.

vmarkovtsev:
I sent you the push permission invitation just in case.

drdavella:
Thanks @vmarkovtsev!

PR #543, opened 2018-09-19
Fix bug when using HDUs with underscore in name with ASDF-in-FITS

drdavella:


drdavella:
The FITS standard actually says that `EXTNAME` can contain any printable ASCII character, so to be completely correct the regex should probably accommodate that range instead.

PR #550, opened 2018-10-04
Add sanity check for extension override

drdavella:
This PR adds a test to make sure that ASDF-provided tags are properly overridden if/when an external package provides an extension with the same type. The test case used here is `gwcs`, although `astropy` could also be used.

cc @nden


drdavella:
@nden, @perrygreenfield I wonder at what point we can say that `asdf` no longer provides WCS tags for backwards compatibility, and if users encounter an older file using these tags, they need to install an older version of `asdf`.

Since presumably all of the users who would care are internal, it seems like this wouldn't be such a big deal.

nden:
I think it's fine to remove the gwcs tags from asdf now.

PR #557, opened 2018-10-09
Automatically store small numeric arrays inline

### LINKED ISSUE #556: Add automatic size threshold for putting array data inline
@perrygreenfield has suggested that arrays below a certain size threshold should be automatically put inline rather than creating data blocks.

I'm not sure what the default threshold should be, but somewhere between 20-100 items seems reasonable. The threshold value should be customizable. Setting the threshold to 0 will mean that no arrays are inlined automatically.

Users already have control over whether particular data arrays are inline or not, so a mechanism is already possible to override the thresholding behavior for individual cases.

Since this is purely an implementation detail, there is no need to modify the standard for this change. However, it may be worth mentioning in the standard that implementations are free to take advantage of this flexibility.

drdavella:
This resolves #556.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 2.2.0 (unreleased)
 ------------------
 
+- Small numeric arrays are now automatically stored inline. This behavior can
+  be overridden using the new ``inline_threshold`` argument to the ``AsdfFile``
+  constructor. It can also be controlled with the existing
+  ``set_array_storage`` method of ``AsdfFile`` and the ``all_array_storage``
+  argument to ``AsdfFile.write_to``. [#557]
 
 2.1.1 (unreleased)
 ------------------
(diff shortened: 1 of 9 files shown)


drdavella:
@perrygreenfield good point. I'll run the JWST test suite locally to see what happens, and I'll also open an issue there before merging this.

I still need to add some tests and documentation here anyway before merging.

Do you have any opinions about the default threshold size? It's currently 50, which is pretty arbitrary, but it did allow more unit tests to pass without modification (many of the tests that expect internal arrays use an array size of 64).

drdavella:
Note: once this is merged, we need to open a PR in the JWST pipeline to make sure that datamodels set `inline_threshold=0` by default.

perrygreenfield:
@drdavella  50 seems reasonable. I think we adjust it based on what people think is best. Eventually, I would like to see the datamodels use another threshold, but they should be in charge of the transition given the impact.

PR #561, opened 2018-10-15
Do not automatically inline small arrays for ASDF-in-FITS

drdavella:


nden:
👎 on this PR. I think the option to change how arrays are stored is sufficient and allows enough flexibility for projects to decide how to handle arrays. I think the current default is fine.

PR #563, opened 2018-10-16
Use decorators to serialize subclasses of known custom types

### LINKED ISSUE #446: Consider the use of decorators for tagging subclasses
This was an idea proposed by @nden: given a tagged class `A`, would it be possible to serialize the subclass `B`, along with any additional properties it might have, using decorators? For that matter, would it be possible to implement all tags using decorators? This probably wouldn't entirely replace the existing tag mechanism, but it could exist as an alternative way to define tags.

cc @Cadair

### LINKED ISSUE #555: ASDF extension behavior appears to be sensitive to order of imports
cc @nden

drdavella:
This resolves #446 and also resolves #555.

More work is required, and there also needs to be some more discussion in order to determine whether this would better be supported by some updates to the standard. Also, we need to evaluate current examples in GWCS, Astropy, and possibly Sunpy to evaluate how this can be improved.

cc @nden, @Cadair

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@
   default. Add configuration setting and command line option to enable schema
   tests. [#676]
 
+- Enable handling of subclasses of known custom types by using decorators for
+  convenience. [#563]
+
 2.3.4 (unreleased)
 ------------------
 
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -72,6 +72,7 @@ def _type_to_tag(type_):
 
 
 def validate_tag(validator, tagname, instance, schema):
+
     if hasattr(instance, '_tag'):
         instance_tag = instance._tag
     else:
--- a/asdf/tags/core/__init__.py
+++ b/asdf/tags/core/__init__.py
@@ -41,6 +41,11 @@ class ExtensionMetadata(AsdfType):
         return tree
 
 
+class SubclassMetadata(dict, AsdfType):
+    name = 'core/subclass_metadata'
+    version = '1.0.0'
+
+
 from .constant import ConstantType
 from .ndarray import NDArrayType
 from .complex import ComplexType
--- a/asdf/tests/__init__.py
+++ b/asdf/tests/__init__.py
@@ -5,7 +5,8 @@ This packages contains affiliated package tests.
 
 import numpy as np
 
-from .. import CustomType
+from .. import CustomType, util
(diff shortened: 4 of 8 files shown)


asdf/asdftypes.py
```diff
@@ -696,7 +703,27 @@ def to_tree_tagged(cls, node, ctx):
             An instance of `asdf.tagged.Tagged`.
         """
         obj = cls.to_tree(node, ctx)
-        return tagged.tag_object(cls.yaml_tag, obj, ctx=ctx)
+        yaml_tag = cls.yaml_tag
+
+        node_cls = type(node)
+
+        if node_cls.__name__ in cls._subclass_map and isinstance(obj, dict):
+            obj['subclass'] = node_cls.__name__
```
drdavella: While this works, to me it seems brittle and inelegant. Ideally we could build this into the standard itself somehow. 

One way to do this would be to use URI fragments, which are allowed by the YAML standard. For example, if the base tag is `tag:nowhere.org/custom/foo-1.0.0`, then a subclass could be tagged with something like `tag:nowhere.org/custom/foo-1.0.0#subclass=FooSubclass`.

This would require a few changes to the validation machinery and also to reference resolution, but it might be worth the effort. The main question is how other ASDF implementations would handle this. Also it would be important to make sure that subclasses are namespaced properly.

drdavella:
@nden sorry for ignoring this for the past week, although I suspect you've been busy with other things too 😉.

I'd like to move forward with this if it's still useful to you. But I think we need to figure out how to support this kind of feature in the standard itself. I'm uncomfortable with having our implementation that adds `subclass` attributes willy-nilly.

We could consider modifying the tags themselves in the way I described above, but ultimately I think that's pretty heavy-handed and might be difficult to replicate in future implementations of ASDF.

I wonder if we can just make the `subclass` attribute a special word during validation? It would have a similar status to existing attributes like `tag` and `description`. We can encode this in the standard and update our validators to handle it properly.

drdavella:
The current implementation of this requires no changes to the standard. I have added important caveats to the documentation, so it should be clear that this is not necessarily a stable feature. Introducing it now should allow other packages to experiment with it. Eventually we may want to have a discussion about whether we can make changes to the standard to better support this.

drdavella:
I'm going to go ahead and merge this so it gets pulled in by repos that test against `master`. It shouldn't affect any existing functionality, but it's always good to confirm.

PR #564, opened 2018-10-17
Format README header to the hipster style

vmarkovtsev:
E.g. like in https://github.com/sharkdp/bat/blob/master/README.md


vmarkovtsev:
@drdavella I can add comments at the top - they will be ignored by the renderer but perfectly visible by humans.

vmarkovtsev:
Done

PR #566, opened 2018-10-17
Add support for arbitrary precision integers

### LINKED ISSUE #552: Provide arbitrary precision integer support
This probably doesn't come up in most of our internal use cases, but there should be a better way of representing integer values larger than 2<sup>52</sup>-1, which is the limit imposed by the standard for literal integers in the tree.

drdavella:
This resolves #552.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,12 @@
   is still effective and the active memory maps may still require the file
   to stay open in case ``copy_arrays`` is ``False``. [#573]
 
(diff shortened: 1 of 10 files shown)


drdavella:
@Commander-Firestrike this PR allows files with large integer literals to be read in, albeit with a warning. It also provides a way to write arbitrarily large integers.

drdavella:
This requires https://github.com/spacetelescope/asdf-standard/pull/176 to be merged before merging here.

PR #571, opened 2018-10-22
Allow tree to be created even when schema is missing

drdavella:
Previously an error occurred in the case where a node was tagged with a type from a known extension, but the type itself was unknown. This could happen, for example, in the case where the type is provided by a more recent version of a package, but only the older version of the package is installed.

Now, just a warning occurs, and the affected node is returned as a basic Python tree.


drdavella:
CI hooks are currently busted 😭

PR #573, opened 2018-10-24
Add lazy_load parameter to open()

vmarkovtsev:
This allows loading ASDF files, then closing the file
and still retaining full access to the tree objects.

Signed-off-by: Vadim Markovtsev <vadim@sourced.tech>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,13 +6,13 @@
   constructor. It can also be controlled with the existing
   ``set_array_storage`` method of ``AsdfFile`` and the ``all_array_storage``
   argument to ``AsdfFile.write_to``. [#557]
-- Add new parameter `lazy_load` to ``AsdfFile.open``. It is ``True`` by
+- Add new parameter ``lazy_load`` to ``AsdfFile.open``. It is ``True`` by
   default and preserves the default behavior. ``False`` detaches the
   loaded tree from the underlying file: all blocks are fully read and
   numpy arrays are materialized. Thus it becomes safe to close the file
-  and continue using ``AsdfFile.tree``. However, `copy_arrays` parameter
+  and continue using ``AsdfFile.tree``. However, ``copy_arrays`` parameter
   is still effective and the active memory maps may still require the file
-  to stay open in case `copy_arrays` is ``False``. [#573]
+  to stay open in case ``copy_arrays`` is ``False``. [#573]
 
 2.1.1 (unreleased)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,13 @@
   constructor. It can also be controlled with the existing
   ``set_array_storage`` method of ``AsdfFile`` and the ``all_array_storage``
   argument to ``AsdfFile.write_to``. [#557]
+- Add new parameter `lazy_load` to ``AsdfFile.open``. It is ``True`` by
+  default and preserves the default behavior. ``False`` detaches the
+  loaded tree from the underlying file: all blocks are fully read and
+  numpy arrays are materialized. Thus it becomes safe to close the file
+  and continue using ``AsdfFile.tree``. However, `copy_arrays` parameter
+  is still effective and the active memory maps may still require the file
+  to stay open in case `copy_arrays` is ``False``. [#573]
 
 2.1.1 (unreleased)
 ------------------
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -90,13 +90,18 @@ class AsdfFile(versioning.VersionedMixin):
 
         copy_arrays : bool, optional
             When `False`, when reading files, attempt to memmap underlying data
-            arrays when possible.
+            arrays when possible. Please note that the file must be opened
+            in "rw" mode; "r" will result a SIGSEGV.
 
         lazy_load : bool, optional
-            When `True` and the read file is seekable, underlying data arrays
-            will be materialized when they are used for the first time.
-            This implies that the file should stay open during the lifetime
-            of the tree.
+            When `True` and the underlying file handle is seekable, data
+            arrays will only be loaded lazily: i.e. when they are accessed
+            for the first time. In this case the underlying file must stay
+            open during the lifetime of the tree. Setting to False causes
+            all data arrays to be loaded up front, which means that they
+            can be accessed even after the underlying file is closed.
+            Note: even if `lazy_load` is `False`, `copy_arrays` is still taken
+            into account.
 
         custom_schema : str, optional
             Path to a custom schema file that will be used for a secondary
@@ -755,13 +760,18 @@ class AsdfFile(versioning.VersionedMixin):
 
         copy_arrays : bool, optional
             When `False`, when reading files, attempt to memmap underlying data
-            arrays when possible.
+            arrays when possible. Please note that the file must be opened
+            in "rw" mode; "r" will result a SIGSEGV.
 
         lazy_load : bool, optional
-            When `True` and the read file is seekable, underlying data arrays
-            will be materialized when they are used for the first time.
-            This implies that the file should stay open during the lifetime
-            of the tree.
+            When `True` and the underlying file handle is seekable, data
+            arrays will only be loaded lazily: i.e. when they are accessed
+            for the first time. In this case the underlying file must stay
+            open during the lifetime of the tree. Setting to False causes
+            all data arrays to be loaded up front, which means that they
+            can be accessed even after the underlying file is closed.
+            Note: even if `lazy_load` is `False`, `copy_arrays` is still taken
+            into account.
 
         custom_schema : str, optional
             Path to a custom schema file that will be used for a secondary
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -90,8 +90,7 @@ class AsdfFile(versioning.VersionedMixin):
 
         copy_arrays : bool, optional
             When `False`, when reading files, attempt to memmap underlying data
-            arrays when possible. Please note that the file must be opened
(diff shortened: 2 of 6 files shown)


drdavella:
Hi @vmarkovtsev thanks for the PR!

We explicitly disallowed access to the tree for a closed ASDF file handle here: https://github.com/spacetelescope/asdf/pull/407.

But it seems like this is intended to apply to the case where you have opened an ASDF file using another open file handle, is that correct? The ASDF file handle remains open.

This PR does seem to fill a gap: you want to retain memory mapped access to an open file handle that existed outside the scope of the ASDF file handle. I assume that using the existing `copy_arrays` option would be too expensive.

It's probably really none of my business, but what's wrong with keeping the file handle open? My concern here is that ASDF may be enabling resource leakage since the underlying file is not actually entirely closed: the ASDF object now holds potentially many memory mapped handles to the original file. I assume garbage collection can handle this properly, but I just want to make sure we're not introducing a problem.

I'm just thinking out loud and trying to make sure I understand the intent of the PR and any side effects it might present.

vmarkovtsev:
We store machine learning models in ASDF. We love ASDF because it is way more efficient than pickles, way more flexible than hdf5 and is not nailed to Python *in theory*.

drdavella:
> So this PR allows to load all the tree directly into memory so that it no longer requires the file backup at all.

Is there a reason the `copy_arrays` argument doesn't meet your needs, then?

> We love ASDF because it is way more efficient than pickles, way more flexible than hdf5 and is not nailed to Python in theory.

Maybe we should add this to the user-testimonial section of our documentation 😉

vmarkovtsev:
I can write a long success story about ASDF fwiw 😄 

The `copy_arrays` argument does a different thing: it enables or disables `mmap()`-ing of the arrays. So if `copy_arrays` is `False`, we `mmap()` and never load the whole array into the memory (cool feature btw). If it is `True`, we read arrays from disk into memory as soon as it is required - that is, lazily.

perrygreenfield:
@vmarkovtsev Please do if you have the time. We are in the process of trying to spread its adoption within astronomy and any support would be quite useful in that regard. Thanks very much for the words of encouragement.

drdavella:
That makes sense. In light of that, this seems like a reasonable change. Just let me review a bit more and then we should be able to move ahead with it.

asdf/asdf.py
```diff
@@ -92,6 +92,12 @@ def __init__(self, tree=None, uri=None, extensions=None, version=None,
             When `False`, when reading files, attempt to memmap underlying data
             arrays when possible.
 
+        lazy_load : bool, optional
+            When `True` and the read file is seekable, underlying data arrays
+            will be materialized when they are used for the first time.
+            This implies that the file should stay open during the lifetime
+            of the tree.
```
drdavella: Sorry to nitpick, but I find this a little confusing. Instead I might say the following:

"When `True` and the underlying file handle is seekable, data arrays will only be loaded lazily: i.e. when they are accessed for the first time. In this case the underlying file must stay open during the lifetime of the tree. Setting to `False` causes all data arrays to be loaded up front, which means that they can be accessed even after the underlying file is closed."

I'm happy to make this change myself, BTW, but just want to make sure it makes sense to you.

vmarkovtsev: So much better!

asdf/block.py
```diff
@@ -262,22 +267,23 @@ def finish_reading_internal_blocks(self):
...
-            if (last_block._fd is not None and
-                last_block._fd.seekable()):
-                last_block._fd.seek(last_block.end_offset)
-                while True:
-                    last_block = self._read_next_internal_block(
-                        last_block._fd, False)
-                    if last_block is None:
-                        break
+        if not self._internal_blocks:
+            return
```
drdavella: Thanks for this, it's much cleaner.

asdf/tags/core/ndarray.py
```diff
@@ -240,6 +240,8 @@ def __init__(self, source, shape, dtype, offset, strides,
         self._offset = offset
         self._strides = strides
         self._order = order
+        if not asdffile.blocks._lazy_load:
+            self._make_array()
```
drdavella: Very minor, but I wonder if there should be an `@property` on the `AsdfFile` itself for this just to make the API a bit cleaner?

asdf/tags/core/tests/test_ndarray.py
```diff
@@ -96,14 +96,16 @@ def test_byteorder(tmpdir):
         }
 
     def check_asdf(asdf):
-        tree = asdf.tree
+        my_tree = asdf.tree
+        for endian in ('bigendian', 'little'):
+            assert my_tree[endian].dtype == tree[endian].dtype
```
drdavella: I assume this is unrelated but just for a bit more coverage?

asdf/tags/core/ndarray.py
vmarkovtsev: Makes sense. I will add that and also fix the docstring.

drdavella: Thanks, the docstring needs to be fixed in both places.

asdf/tags/core/tests/test_ndarray.py
vmarkovtsev: The old code shadowed `tree` from the outer scope, also that test failed when I debugged the test suite because the dtype-s did not match completely, so I spent some fun time here.

asdf/tests/helpers.py
```diff
@@ -233,6 +233,15 @@ def assert_roundtrip_tree(tree, tmpdir, *, asdf_check_func=None,
         finally:
             server.finalize()
 
+    # Now don't be lazy and check that nothing breaks
+    with io.BytesIO() as buff:
+        AsdfFile(tree, extensions=extensions, **init_options).write_to(buff, **write_options)
+        buff.seek(0)
+        ff = AsdfFile.open(buff, extensions=extensions, copy_arrays=True, lazy_load=False)
```
drdavella: I wonder if we also want a test with `copy_arrays=False`? Unless you think that `lazy_load=False` always implies `copy_arrays=True`, in which case we may need to make a change to the top-level API.

vmarkovtsev: Yeah, I will add the test.

vmarkovtsev:
@drdavella All the review comments have been addressed, waiting for the CI to pass.

vmarkovtsev:
Done

asdf/asdf.py
```diff
@@ -90,7 +90,18 @@ def __init__(self, tree=None, uri=None, extensions=None, version=None,
 
         copy_arrays : bool, optional
             When `False`, when reading files, attempt to memmap underlying data
-            arrays when possible.
+            arrays when possible. Please note that the file must be opened
+            in "rw" mode; "r" will result a SIGSEGV.
```
drdavella: Not for this PR, but it seems like we should have better error handling for this case?

vmarkovtsev: Definitely. I had to debug segmentation faults otherwise.

CHANGES.rst
```diff
@@ -6,6 +6,13 @@
   constructor. It can also be controlled with the existing
   ``set_array_storage`` method of ``AsdfFile`` and the ``all_array_storage``
   argument to ``AsdfFile.write_to``. [#557]
+- Add new parameter `lazy_load` to ``AsdfFile.open``. It is ``True`` by
+  default and preserves the default behavior. ``False`` detaches the
+  loaded tree from the underlying file: all blocks are fully read and
+  numpy arrays are materialized. Thus it becomes safe to close the file
+  and continue using ``AsdfFile.tree``. However, `copy_arrays` parameter
+  is still effective and the active memory maps may still require the file
+  to stay open in case `copy_arrays` is ``False``. [#573]
```
drdavella: Sorry for another nit, but we should have double backticks around `lazy_load` and `copy_arrays` here so they render properly on github.

vmarkovtsev: Sure

asdf/asdf.py
vmarkovtsev: This is the current behavior:

```
asdf.open("/path", mode="r")  # SIGSEGV
asdf.open("/path", mode="rw")  # OK
with open("/path", mode="r") as file:
    asdf.open(file)  # SIGSEGV
with open("/path", mode="rw") as file:
    asdf.open(file)  # OK
```

So the comment is always effective as long as the option itself is effective (that is, no streaming and non-seekable files).

vmarkovtsev: Actually, I am surprised that this has not been reported yet. In my case I always use compressed blocks so never had that SIGSEGV before.

drdavella: Ah okay thanks. I've opened an issue to track this. We can leave the comment as-is here.

vmarkovtsev: So yeah, I should have added that the "rw" requirement is only for files with uncompressed arrays.

drdavella: Yes, I think I'm confused by this and will need to investigate further. Your example above suggests that opening any file without explicitly given a mode of `'rw'` leads to segfault, which doesn't make sense since this is most common use case, both in the test suite, and in our internal code base.

drdavella: Even with uncompressed arrays, that's the most common use case since it's the default.

drdavella: Sorry, but can we leave this comment out of this PR? I'm tracking it in #574 but I don't want to hold this PR up while I figure out what the issue is here.

vmarkovtsev: Sure

drdavella:
Thanks @vmarkovtsev!

Issue #574, opened 2018-10-24
Address bug when using memory maps with open file handle in readonly mode


drdavella:
This was raised in a discussion with @vmarkovtsev in #573. There should be a test case and better error handling, as well as better documentation.

drdavella:
See https://github.com/spacetelescope/asdf/pull/573#discussion_r227866944 for related discussion and a useful test case.

vmarkovtsev:
This is relevant: https://github.com/spacetelescope/asdf/blob/master/asdf/tests/helpers.py#L204

drdavella:
Good catch, thanks.

stscijgbot:
This ticket is now being tracked at [AL-120](https://jira.stsci.edu/browse/AL-120)

PR #575, opened 2018-10-24
Attempt to reproduce segfault reported in #574

drdavella:
@vmarkovtsev I'm trying to reproduce the error you reported that we documented in #574. This seems to work fine for me, though, so I must be missing something.


drdavella:
🤔

PR #576, opened 2018-10-25
Reproduce SIGSEGV (#574)

vmarkovtsev:


vmarkovtsev:
@drdavella For example, https://travis-ci.org/spacetelescope/asdf/jobs/446063632#L2365

drdavella:
Thanks, I'll try to track this down today. I added `pytest-faulthandler` to the travis jobs a while ago, which is very useful for tracing these kinds of failures.

drdavella:
The only test that fails with this change is this one:
https://github.com/spacetelescope/asdf/blob/32abe89ce024ec93b838bdf684a74662a83ee0f0/asdf/tags/core/tests/test_ndarray.py#L62-L88

Specifically, the failure is caused by this line:
https://github.com/spacetelescope/asdf/blob/32abe89ce024ec93b838bdf684a74662a83ee0f0/asdf/tags/core/tests/test_ndarray.py#L82

Presumably this is because it attempts to write to a memory map that has been opened as read-only. I can reproduce this reliably, but I'm still not entirely certain about the root cause at this point.

My proposal to fix this bug is to deprecate the `mode=` argument to `asdf.open`, since I don't think it serves any real purpose and only can lead to issues like this one. See #578 for discussion.

PR #579, opened 2018-10-26
Refactor, better mode handling, deprecate asdf.AsdfFile.open

### LINKED ISSUE #574: Address bug when using memory maps with open file handle in readonly mode

### LINKED ISSUE #578: Deprecate mode argument to asdf.open
As far as I can tell, this argument does not serve any useful purpose, but only allows for buggy behavior related to memory maps (see the discussion in #575 and #576).

For these reasons, I propose to remove it entirely. For the next few releases, using this argument will result in a warning. By 3.0, it should be removed entirely.

cc @vmarkovtsev

drdavella:
This resolves #578 (albeit indirectly) and resolves #574.

It refactors the top-level `asdf.open` function into a standalone factory function in order to enable better decisions about handling file modes, particularly with respect to memory mapping. It also deprecates the use of `asdf.AsdfFile.open`, since this was not capable of handling such decisions.

There's still a bit more work:
- [ ] ~Add a warning when explicitly giving `mode='r'` with `copy_arrays=False`~ (no longer applies, see below)
- [x] Fix any internal usage that results in the above warning
- [x] Update the change log

Even though this is technically a bug fix, it affects the API and is a big enough change that it should go in 2.2.0.

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -725,6 +725,12 @@ class AsdfFile(versioning.VersionedMixin):
              custom_schema=None,
              strict_extension_check=False,
              ignore_missing_extensions=False):
+        """
+        Open an existing ASDF file.
+
+        .. deprecated:: 2.2
+            Use `asdf.open` instead.
+        """
 
         warnings.warn(
             "The method AsdfFile.open has been deprecated and will be removed "
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -589,8 +589,33 @@ class AsdfFile(versioning.VersionedMixin):
 
         return None
 
+    def _check_and_set_mode(self, fileobj, asdf_mode):
+
+        memmap = self._blocks.memmap
+
+        if asdf_mode is not None and asdf_mode not in ['r', 'rw']:
+            msg = "Unrecognized asdf mode '{}'. Must be either 'r' or 'rw'"
+            raise ValueError(msg.format(asdf_mode))
+
+        if asdf_mode is None:
+            if isinstance(fileobj, str):
+                parsed = generic_io.urlparse(fileobj)
+                if parsed.scheme == 'http':
+                    return 'r'
+                return 'rw' if memmap else 'r'
+            if isinstance(fileobj, io.IOBase):
+                return 'rw' if fileobj.writable() and memmap else 'r'
+
+            raise ValueError("Unknown file object type, can't guess mode")
+
+        # It is not safe to open files with memory maps in readonly mode
+        elif asdf_mode == 'r' and memmap:
+            raise ValueError("Can't open file as readonly without copy_arrays=True") 
+
+        return asdf_mode
+
     @classmethod
-    def _open_asdf(cls, self, fd, uri=None, mode='r',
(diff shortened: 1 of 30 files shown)


asdf/asdf.py
```diff
@@ -1287,6 +1222,132 @@ def get_history_entries(self):
...
+def _check_and_set_mode(fileobj, asdf_mode):
+
+    if asdf_mode is not None and asdf_mode not in ['r', 'rw']:
+        msg = "Unrecognized asdf mode '{}'. Must be either 'r' or 'rw'"
+        raise ValueError(msg.format(asdf_mode))
+
+    if asdf_mode is None:
+        if isinstance(fileobj, str):
+            parsed = generic_io.urlparse.urlparse(fileobj)
+            if parsed.scheme == 'http':
```
vmarkovtsev: What if the scheme is https

asdf/asdf.py
```diff
@@ -1287,6 +1222,132 @@ def get_history_entries(self):
...
+    asdffile : AsdfFile
+        The new AsdfFile object.
+    """
+
+    # For now retain backwards compatibility with the old API behavior,
+    # specifically when being called from AsdfFile.open
+    if not _compat:
+        mode = _check_and_set_mode(fd, mode)
+        if mode == 'r' and not copy_arrays:
+            copy_arrays = True
```
vmarkovtsev: Should we print a warning here?

asdf/asdf.py
```diff
@@ -1287,6 +1222,132 @@ def get_history_entries(self):
 AsdfFile.keys.__doc__ = dict.keys.__doc__
 
 
+def _check_and_set_mode(fileobj, asdf_mode):
+
+    if asdf_mode is not None and asdf_mode not in ['r', 'rw']:
```
vmarkovtsev: Can we have r+ or w+ modes?

asdf/asdf.py
drdavella: Yes, but I just pushed a new set of changes that makes this irrelevant. I think the update contains a better solution to the issue you reported.

asdf/asdf.py
drdavella: Do you just mean that we should add these and make them equivalent to the current `'rw'`? Although I think to be consistent with Python's `open` API, `'w+'` would have to overwrite an existing file. Maybe we should support that, though?

asdf/tags/core/ndarray.py
```diff
@@ -344,6 +346,16 @@ def __getattr__(self, attr):
...
 
+    def __setitem__(self, *args):
+        # This workaround appears to be necessary in order to avoid a segfault
+        # in the case that array assignment causes an exception. The segfault
+        # originates from the call to __repr__ inside the traceback report.
+        try:
+            self._make_array().__setitem__(*args)
+        except Exception:
+            self._array = None
+            raise
```
vmarkovtsev: Since we've cut with 2.7:

```python
except Exception as e:
    self._array = None
    raise e from None
```

seems better - it does not pollute the error trace.

PR #591, opened 2018-11-02
Do not allow builtin extension to override subclasses

drdavella:
In particular, those that were introced by an extension from an external package.

This fixes a regression that was introduced in `2.1.1` and was first discovered in the Astropy test suite (see [here](https://travis-ci.org/astropy/astropy/jobs/449635171)).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,13 @@
 - Add readonly protection to memory mapped arrays when the underlying file
   handle is readonly. [#579]
 
+2.1.2 (unreleased)
+------------------
+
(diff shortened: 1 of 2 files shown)


nden:
Is this related to #560?

drdavella:
Yes, and apparently the fix I introduced here is not correct, so I'll need to dig a bit more.

This bug is actually non-deterministic and only appears on Py35, which makes me think there's somewhere where we're expecting ordered dictionary behavior. This has led to other bugs in the past and it's always a pain.

drdavella:
Closing since I don't think this is the real problem.

Issue #592, opened 2018-11-03
Investigate test failure from latest Astropy update


drdavella:
The failure can be seen [here](https://travis-ci.org/spacetelescope/asdf/jobs/450179102#L2134) and is apparently due to a recent update on `astropy/master`.

nden:
Caused by the removal of the `DomainType`?

drdavella:
I don't think we should have to keep it in Astropy.

I think it makes sense to have a version bump on the `transform` schema that removes the `domain` field. This would affect v1.3.0 of the standard, but the new transform schema would be v1.2.0 under the convention we're using.

I think it also makes sense to remove the `domain` schema in v1.3.0 of the standard.

The fiddly part is our dependency on Astropy: is this change going into 3.1 or are we going to have to wait for another release cycle?

drdavella:
Just adding for future reference that the relevant PR in Astropy is https://github.com/astropy/astropy/pull/8052.

stscijgbot:
This ticket is now being tracked at [AL-112](https://jira.stsci.edu/browse/AL-112)

PR #593, opened 2018-11-04
Remove gwcs tags

### LINKED ISSUE #554: Deprecate/remove WCS tags from ASDF core
These are now provided by [gwcs](https://gwcs.readthedocs.io/en/latest).

nden:
Resolves #554 

It's not clear to me whether these should be deprecated or removed. AFAIK there's no code that references them. However, this probably means the schemas should be removed as well or testing them fails. Note that schemas and tags are now in gwcs.
This PR remove the tags and https://github.com/spacetelescope/asdf-standard/pull/179 removes the schemas. 
Feel free to close both if this is not the right approach.


drdavella:
In theory a deprecation period would be nice, and I think we should aspire to that for most other parts of the API.

However, in this case, who would we really be helping? In all likelihood the only users of these tags are internal, and they should already be using the ones in GWCS anyway. So I think I'm okay with removing both tags and schemas in the next release (2.2.0). Removal of the schemas would affect v1.3.0 of the standard.

Does this also fix #592?

drdavella:
@nden I just added an option to allow us to skip testing the examples in the `frame` schemas. This means that even if we decide to keep the schemas in the standard for now, they will still be validated by the test suite, but the examples won't run.

PR #595, opened 2018-11-05
Do not automatically add subclasses of ndarray to type index

drdavella:
This resolves an issue that was first reported in Astropy in https://github.com/astropy/astropy/pull/8064.


drdavella:
Actually I don't think this is the right solution either, and in fact is likely to have bad side effects for some users.

PR #596, opened 2018-11-05
Fix bug in logic for overriding subclasses

drdavella:
This fixes a bug first exposed by https://github.com/astropy/astropy/pull/8064.


drdavella:
This doesn't actually seem to fix the bug.

PR #606, opened 2018-11-15
remove non-ascii characters from readme

### LINKED ISSUE #605: README in released 2.2.0 breaks (some?) pip installs
I'm not sure exactly how broad the impact of this is... but note the following build:
https://travis-ci.org/astropy/astropy/jobs/455319988

That seems to be failing due to astropy trying to install asdf 2.2.0, which is then failing due to the non-ascii characters in the README (specifically, the bullets in the "header" of the README)... 

The blunt-force fix would be to just remove the bullets - I'll issue a PR to do just that momentarily (#606).  But if there's some better solution that somehow avoids this reading step I'm all ears...

eteq:
This fixes #605

Note that there *may* be some way to instead convince the pip step that's causing #605 to interpret the encoding differently... But it wasn't clear to me how.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,10 +17,10 @@
      <a href="http://www.astropy.org/"><img src="http://img.shields.io/badge/powered%20by-AstroPy-orange.svg?style=flat" alt="astropy"></a>
    </p>
    <p align="center">
-     <a href="#overview">Overview</a> •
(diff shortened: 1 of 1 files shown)


drdavella:
Closing this in favor of #607.

PR #607, opened 2018-11-15
Fix readme render

### LINKED ISSUE #605: README in released 2.2.0 breaks (some?) pip installs

drdavella:
This is intended to fix #605 and also to fix rendering of the README on pypi. It seems nearly impossible to test without another release, unfortunately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ env:
         # Make sure that installation does not fail
         - SETUP_CMD='install'
         # Make sure README will display properly on pypi
-        - PIP_DEPENDENCIES='collective.checkdocs pygments' SETUP_CMD='checkdocs'
+        - PIP_DEPENDENCIES='twine' TWINE_CHECK=1
         - PYTHON_VERSION=3.5 SETUP_CMD='test'
         - PYTHON_VERSION=3.6 SETUP_CMD='test'
         - PYTHON_VERSION=3.7 PYTEST_VERSION=3.8 SETUP_CMD='test'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -89,7 +89,12 @@ install:
     - python setup.py install
 
 script:
-   - $MAIN_CMD $SETUP_CMD
+    - if [[ $TWINE_CHECK ]]; then
+        python setup.py build sdist;
+        twine check dist/*;
```
eteq: Oh, this is diabolical (in a good way)

setup_helpers.py
```diff
@@ -84,9 +85,16 @@ def read_metadata(config_filename):
 
 
 def read_readme(readme_filename):
-    with open(readme_filename) as ff:
+    with codecs.open(readme_filename, encoding='utf8') as ff:
         lines = ff.read().splitlines()
 
+    # Skip lines that contain raw HTML markup
+    lines = lines[:4] + lines[26:]
```
eteq: This seems a bit dangerous - what if someone else makes the README and doesn't realize they need to update this file?

Maybe instead look for the `.. raw:: html` marker, and then skip all consecutive following lines that are indented (or empty)?

drdavella: Yes, you're absolutely right. We should do something smarter like that in the long term but for the sake of this bug fix I'm going to leave it alone.

drdavella: Also the point of the `twine check` above is that it should catch any such changes that lead to invalid output, but we're in agreement.

eteq:
Good points re: the comment @drdavella - I'll make an issue right now to track the "make the html thingie work better", but doesn't have to go in this PR.

.travis.yml
drdavella: 😈

Issue #599, opened 2018-11-12
auto_inline not working as expected


nden:
There must be a subtle bug in `auto_inline` - it seems the behavior depends on previously run commands. Is something setting a state?
```
f = AsdfFile({'data': np.arange(6)})
f.write_to('test.asdf') # correctly writes the data inline

f.write_to('test.asdf', auto_inline=3) # still writes it inline

# force it to write to a block
f.write_to('test.asdf', all_array_storage='internal') 

f.write_to('test.asdf', auto_inline=3) # writes to a block as expected

f.write_to('test.asdf', auto_inline=10) # writes it inline

f.write_to('test.asdf', auto_inline=3) # still writes it inline

```

drdavella:
Thanks for the report. I've noticed other places where `write_to` modifies the state of the `AsdfFile` object, which is not okay, in my opinion. I wish there was a better way to enforce this in Python (I'm thinking of `const` methods in C++, which can't update the associated object).

A related discussion came up in #597 regarding the use/implementation of `inline_threshold`. I think I'm going to have to rework that API and somehow combine it with `auto_inline` since they serve the same purpose. Since the `inline_threshold` feature is still in an unreleased version of ASDF, we have no concerns about making further API changes.

drdavella:
This is still a real issue, even after #610. Unfortunately, however, it's one that is unlikely to be resolved without a major redesign. I'm considering some major API changes for 3.0, which would hopefully address this kind of issue.

stscijgbot:
This ticket is now being tracked at [AL-110](https://jira.stsci.edu/browse/AL-110)

PR #610, opened 2018-11-16
Revert inline threshold option

drdavella:
This feature has proved to be too problematic to be supported in an official release. The semantics are unfortunately rather complicated due to the fact that the same `AsdfFile` object can be used both for reading and for writing. It also turns out that there already exists an option to `write_to` called `auto_inline` that serves a similar purpose, although I believe it may not be working properly in all cases (see #599).

Also, to complicate matters further, neither `inline_threshold` nor `auto_inline` are idempotent, meaning that both modify the underlying `AsdfFile` object in undesirable ways. This has exposed some significant design warts in the `asdf` API and implementation that need to be addressed, although that will be difficult and will probably lead to significant user-facing changes.

I hope to revisit this and address the issues in an upcoming PR, but I have begun to consider whether we need to make some significant design changes for 3.0.

cc @jdavies-st

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,6 @@
 2.3.0 (unreleased)
 ------------------
 
-- Small numeric arrays are now automatically stored inline. This behavior can
-  be overridden using the new ``inline_threshold`` argument to the ``AsdfFile``
-  constructor. It can also be controlled with the existing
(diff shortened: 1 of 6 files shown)


drdavella:
This PR retains the tests that were added and also a helper function for deciding when to inline since those will be useful for any future work and I don't want to lose them.

PR #617, opened 2018-11-21
Support pathlib with asdf.open and AsdfFile.write_to

### LINKED ISSUE #616: Accept pathlib.Path objects
Specifically in `asdf.open` but also everywhere else you can specify a path-like thing.

drdavella:
This resolves #616.

cc @Cadair

### DIFF
--- a/asdf/tests/test_api.py
+++ b/asdf/tests/test_api.py
@@ -2,6 +2,7 @@
 
 import os
 import io
+import pathlib
 
 import numpy as np
 from numpy.testing import assert_array_equal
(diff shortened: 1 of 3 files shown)


asdf/tests/test_api.py
```diff
@@ -234,6 +235,20 @@ def test_context_handler_resolve_and_inline(tmpdir):
         newf.tree['random'][0]
 
 
+def test_open_pathlib_path(tmpdir):
+
+    filename = str(tmpdir.join('pathlib.asdf'))
```
Cadair: Probably not relevant to this test, but `tmpdir` here is actually `pathlib.Path` like.

drdavella: Ah yeah, good point. I think I'm going to leave as-is to be explicit.

PR #641, opened 2019-01-17
Fix extension version check

drdavella:
This fixes a bug (and adds a test) for an issue that was reported by @jdavies-st.


jdavies-st:
Thanks @drdavella!

PR #649, opened 2019-02-25
Fix fake URLs to point to non-existent site...

drdavella:
This is kind of funny, but apparently someone bought the domain that we were using to test fake URLs, which caused test failures. So updating for now.

### DIFF
--- a/asdf/tags/core/tests/test_history.py
+++ b/asdf/tags/core/tests/test_history.py
@@ -46,7 +46,7 @@ def test_history():
     assert 'history' not in ff.tree
     ff.add_history_entry('This happened',
                          {'name': 'my_tool',
-                          'homepage': 'http://nowhere.com',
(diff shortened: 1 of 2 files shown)


asdf/tags/core/tests/test_history.py
```diff
@@ -69,7 +69,7 @@ def test_history_to_file(tmpdir):
     with asdf.AsdfFile() as ff:
         ff.add_history_entry('This happened',
                              {'name': 'my_tool',
-                              'homepage': 'http://nowhere.com',
+                              'homepage': 'http://nowhere.org',
```
pllim: This only works until they also buy nowhere.org... :thinking:

drdavella: This isn't validated anywhere. The only one that actually matters for testing purposes is the one in the diff below.

PR #654, opened 2019-03-06
Use $schema keyword to determine metaschema to use for schema tests

drdavella:
Prior to this change the YAML meta-schema ID was hard-coded into the `check_schema` routine which is used by the schema tester. This ID is still used as a fallback if the `$schema` keyword is not present, however.


drdavella:
Tested this against the schemas in `astropy` and `gwcs` and it seems like the impact is minimal (no changes required).

PR #664, opened 2019-03-22
POC: Automatically add description from schema when available

drdavella:
This is a prototype of an idea that @perrygreenfield proposed for automatically adding description fields from schemas to the associated nodes of ASDF files when they are written. It should not be merged as-is, but does demonstrate proof-of-concept.

I think we might want to add a class attribute to `asdf.types.ExtensionType` that controls whether the description field should be written or not. It can be set to `False` by default but child tag implementations can override it.

This could also make it possible to populate descriptions contained in an ASDF file into the in-memory representation of the tree, where it could then be accessed by tools.

### DIFF
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -102,6 +102,14 @@ def validate_propertyOrder(validator, order, instance, schema):
     instance.property_order = order
 
 
+def validate_additionalProperties(validator, additionalProperties, instance, schema):
+
+    if isinstance(instance, dict):
+        instance.pop('_description', None)
+    return mvalidators.Draft4Validator.VALIDATORS['additionalProperties'](
+        validator, additionalProperties, instance, schema)
+
+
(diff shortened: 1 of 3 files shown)


drdavella:
If we intend to support this, we need to add something to the standard to codify this convention. I propose adding a reserved attribute like `$description` (in the same vein as `$ref`) in order to make sure we don't collide with other potential attributes.

perrygreenfield:
It would be good to illustrate this with an example showing when it would be automatically used to add a description field. It certainly would be important to be able to indicate whether or not it should be propagated to the asdf file. I can also see a need for two different description fields in the schema. One more detailed on for those reading the schema, and one as a comment to be used in the asdf file. I'm not sure if this would support that.

WilliamJamieson:
Closing. This has become very stale

PR #670, opened 2019-04-04
Packaging overhaul

### LINKED ISSUE #667: no module asdf.version after pip install of directory in clean environment
I created a new virtual environment, did `pip install .` in the `asdf` directory and then tried to run `pytest`. It came back with an error:
```
  File "/home/craig/Documents/STScI/asdf/asdf/asdf/versioning.py", line 22, in <module>
    from .version import version as asdf_version
ImportError: Error importing plugin "asdf.tests.schema_tester": No module named 'asdf.version'
```

Discussed with @drdavella and he said it might be a packaging issue.  What I had to install was `pytest-astropy` and then things seemed to work.

drdavella:
This overhauls the packaging infrastructure. It does away with `astropy_helpers` once and for all. It also adds a `pyproject.toml` file and moves nearly all `setuptools` configuration to `setup.cfg`.

This also fixes #667.


setup.cfg
```diff
@@ -25,17 +86,6 @@ addopts = --doctest-rst
 [ah_bootstrap]
 auto_use = True
```
jdavies-st: Can this `[ah_bootstrap]` section be removed?

drdavella: Yes, good catch!

PR #675, opened 2019-04-08
add .eggs to gitignore

eteq:
Does just what it says.  It *may* have been that actually `eggs` was a typo and it was supposed to be `.eggs`?  But just in case I figured I'd make it a separate entry...

(Encountered when doing `python setup.py build`)


drdavella:
Thanks @eteq!

Issue #642, opened 2019-02-01
tuple keys not handled


bblais:
Smallest program which reproduces the error (asdf from conda 2.3.1):

```
import asdf

mydata={'hello':'there',
        (6,5):'bad key',
        'that':6,
       }
# Make the tree structure, and create a AsdfFile from it.
tree = {'hello':'there',
        'data':mydata,
       }
ff = asdf.AsdfFile(tree)  # saves, but error on load below
ff.write_to("test.asdf")

ff = asdf.AsdfFile(mydata)  # gives TypeError: unhashable type: 'list'
ff.write_to("test2.asdf")

with asdf.open('test.asdf') as af: # gives ConstructorError : while constructing a mapping
    print(af.tree)

```
Is this a known issue?  thanks!

drdavella:
Hi @bblais, thanks for the bug report. I am able to reproduce this and will investigate further.

This looks like it may actually be a limitation of the YAML implementation we use, but I need to dig deeper.

drdavella:
This does in fact appear to be a limitation of `pyyaml`. Consider the following example:

```python
import yaml

mydata = { 'hello':'there', (6,5):'bad key', 'that':6 }
s = yaml.safe_dump(mydata)
yaml.safe_load(s)
```

The result is the same error as in your example:
```python
~/miniconda3/envs/asdfdev/lib/python3.6/site-packages/yaml/constructor.py in construct_mapping(self, node, deep)
    126             if not isinstance(key, collections.Hashable):
    127                 raise ConstructorError("while constructing a mapping", node.start_mark,
--> 128                         "found unhashable key", key_node.start_mark)
    129             value = self.construct_object(value_node, deep=deep)
    130             mapping[key] = value

ConstructorError: while constructing a mapping
  in "<unicode string>", line 1, column 1:
    hello: there
    ^
found unhashable key
  in "<unicode string>", line 2, column 3:
    ? [6, 5]
      ^
```

There doesn't seem to be anything in the YAML spec that would prevent it from being used this way, so this may actually be a bug in the `pyyaml` implementation.

drdavella:
It's worth noting that [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) appears to handle this case correctly. We have an issue for exploring the use of that package vs `pyyaml` (#391), so this would be a motivating use case.

drdavella:
[...]

PR #677, opened 2019-04-11
Replace pyyaml with ruamel.yaml

### LINKED ISSUE #391: Research/experiment with ruamel.yaml as an alternative to pyyaml
http://yaml.readthedocs.io/en/latest/pyyaml.html
https://pypi.python.org/pypi/ruamel.yaml

`ruamel.yaml` appears to be under active development, whereas `pyyaml` has not been updated since 2016. It might be worth considering whether we should migrate to a newer library, as long as it meets our needs. It's also possible that `ruamel.yaml` might be more performant since it appears to be built with C extensions and/or Cython. This may make it less portable, though. It's also worth noting that `ruamel.yaml` supports YAML 1.1 and 1.2, whereas `pyyaml` only supports 1.1.

drdavella:
This PR closes #391. By updating `ruamel.yaml` it appears that we can remove some older logic in `asdf` that provided handling for `omap` types. It looks like `pyyaml` handled those in an idiosyncratic way, whereas `ruamel.yaml` handles them better (see https://github.com/yaml/pyyaml/issues/78).

This PR does not address #642. The solution does not appear to be straightforward and so it will need to be fixed in a separate PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@
 - Enable handling of subclasses of known custom types by using decorators for
   convenience. [#563]
 
+- Replace use of ``pyyaml`` with ``ruamel.yaml``. [#677]
+
 2.3.4 (unreleased)
 ------------------
 
--- a/asdf/__init__.py
+++ b/asdf/__init__.py
@@ -19,9 +19,9 @@ __all__ = [
 ]
 
 try:
-    import yaml as _
+    import ruamel.yaml as _
 except ImportError:
-    raise ImportError("asdf requires pyyaml")
+    raise ImportError("asdf requires ruamel.yaml")
 
 try:
     import jsonschema as _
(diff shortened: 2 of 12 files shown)


Cadair:
This looks awesome :smile: Do you have any indication if this has any performance impact one way or another?

jdavies-st:
All the `jwst` package unit tests pass with this change.  So we are 👍

drdavella:
@Cadair I have not had time to look into performance yet. My intuition is that it will be no worse than `pyyaml`, but it should be measured.

drdavella:
This isn't a great benchmark, but I did
```python
%timeit asdf.test()
```

On this branch:
`18.7 s ± 27.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

On `master`:
`18 s ± 32.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

WilliamJamieson:
Closing. This has become too out of date and stale.

PR #680, opened 2019-05-09
Add test for schema url_mapping via entry points

jdavies-st:
This is to try to debug a problem on Windows.

https://github.com/astropy/astropy/issues/8683

### DIFF
--- a/asdf/tests/test_schema.py
+++ b/asdf/tests/test_schema.py
@@ -452,12 +452,12 @@ def test_self_reference_resolution():
 
 def test_schema_resolved_via_entry_points():
     """Test that entry points mappings to core schema works"""
-    resolver = asdf.AsdfFile().resolver
-    s = schema.load_schema(
-        'http://stsci.edu/schemas/asdf/core/asdf-1.1.0',
-        resolver=resolver,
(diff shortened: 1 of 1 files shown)


pllim:
From my error:
```
 Invalid argument: '/C:/.../site-packages/asdf/schemas/stsci.edu/asdf/core/table-1.0.0.yaml'
```

I think it was trying to read in a schema that shipped with `asdf`, not from some HTTP site.

pllim:
Now, I see the failure on Astropy too: https://ci.appveyor.com/project/Astropy/astropy/builds/24435056

xref astropy/astropy#8688

jdavies-st:
Well, at least we now have consistent failures on Windows.

jdavies-st:
Let's see if we can reproduce that failure over in this package.

pllim:
Still passed here! That's bad, right?

jdavies-st:
It means that the problem is in the `astropy` version of the test.  Fixed over there.  Closing this PR.

jdavies-st:
Maybe it would be good to actually add this test in `asdf`, as I believe it utilizes a different path through the code from the others.  It can't hurt.

PR #682, opened 2019-05-31
Fix load_schema memory issue and add convenience function for fetching the default resolver.

### LINKED ISSUE #672: Add a function to API for resolver that uses installed extensions
Currently in order to resolve ids or tags to schema file locations for all installed extensions, it is necessary to use the `resolver` method of an `AsdfFile` instance. It would be useful to have a standalone function as part of the API that provides the same functionality. The function `asdf.resolver.default_resolver` only knows about schemas that are actually installed as part of the `asdf` package; it is not aware of any installed extensions.

Along similar lines, it might be useful to have a convenience function that loads a schema by id. It would be similar to `asdf.schemas.load_schema` but instead of taking an absolute file path, it would take a tag or id and use the resolver function mentioned above to find the real schema location.

cc @jdavies-st

### LINKED ISSUE #681: Problematic load_schema LRU Cache Memory Usage
When saving fairly large arrays into multiple different files, i encounter a nasty issue, eventually leading to asdf claiming my complete RAM.
It seems to be related to the load_schema function who's lru_cache for some reason (i am not familiar with the internals of asdf) also includes the complete data arrays.

Please see the following example (and watch your memory allocation grow):
```python
import numpy as np
import asdf
import gc
for i in range(1000):
    print(i)
    a = {'a': np.random.rand(1000*1000*100)}
    with asdf.AsdfFile(a) as fh:
        fh.write_to(
            '{:08d}.asdf'.format(i),
        )
    # uncomment to cure the issue by manually clearing said cache and force a gc afterwards
    # asdf.schema.load_schema.cache_clear()
    # gc.collect()
```

Here my conda environment (excerpt):
```
python                    3.7.3                h5b0a415_0    conda-forge
asdf                      2.3.1                      py_0    conda-forge
numpy                     1.16.3           py37he5ce36f_0    conda-forge
jsonschema                2.6.0                 py37_1002    conda-forge
pyyaml                    5.1              py37h14c3975_0    conda-forge
``` 
on Linux 3.16.0-8-amd64

eslavich:
This PR does the following:

- `AsdfFile.resolver` is now a property instead of a method, so caching it does not prevent the `AsdfFile` from being garbage collected.  The property returns an instance of `ResolverChain`, which is callable, so this change will hopefully be transparent to users.
- Fix Resolver's hash behavior
- Add `extension.get_default_resolver` for retrieving the default resolver without first instantiating an `AsdfFile`
- Deprecate `resolvers.default_url_mapping`, `resolvers.default_tag_to_url_mapping`, and `resolvers.default_resolver`, since none of these include mappings from loaded extensions
- Deprecate `extension.default_ext_resolver` because `load_schema` won't recognize it as identical to the default `ResolverChain`
- Deprecate `resolver.Resolver.add_mapping` because `Resolver` needs to be immutable to be used as a key in lru_cache's dictionary
- Add some `Resolver` tests

Fixes #681 
Fixes #672

### DIFF
--- a/asdf/resolver.py
+++ b/asdf/resolver.py
@@ -3,9 +3,11 @@
 
 import sys
 import os.path
+import warnings
 
 from . import constants
 from . import util
+from .exceptions import AsdfDeprecationWarning
 
 
 def find_schema_path():
@@ -24,11 +26,11 @@ class Resolver:
     A class that can be used to map strings with a particular prefix
     to another.
     """
-    def __init__(self, mapping=[], prefix=''):
+    def __init__(self, mappings, prefix):
         """
         Parameters
         ----------
-        mapping : list of tuple or callable, optional
+        mappings : list of tuple or callable
             A list of mappings to try, in order.
             For each entry:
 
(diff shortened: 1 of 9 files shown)


eslavich:
The latest commit changes `load_schema(...)` to return a deep copy of the cached schema object. 
 The schema object is mutable, and includes mutable subschema objects that are shared among many different schemas, and it was a bit of a nasty surprise to discover that changes made to one schema may be reflected in many other schemas.

@jdavies-st discovered this problem when testing this branch of asdf with jwst, where the MultiExposureModel makes adjustments to its own schema when instantiated.  With the caching working properly, it was removing FITS keywords from the core schema for all datamodels.

asdf/schema.py
```diff
@@ -385,7 +391,20 @@ def load_schema(url, resolver=None, resolve_references=False,
         control local reference resolution separately.
     """
     if resolver is None:
-        resolver = default_ext_resolver
+        # We can't just set this as the default in load_schema's definition
+        # because invoking get_default_resolver at import time leads to a circular import.
+        resolver = extension.get_default_resolver()
+
+    # We want to cache the work that went into constructing the schema, but returning
```
perrygreenfield: Do we have examples of where a schema is changed on the fly to worry about this? If not, is there a good use case you have in mind?

eslavich: One of the tests in this repo does, which leads to thrilling heisenbugs in other tests.  I'm not aware of a real use case but this kind of problem is so difficult to track down that it seemed best to be defensive here.

perrygreenfield: If it were possible to make the schema immutable, that would be my solution (I haven't looked into that). But this will do for now.

jdavies-st:
Conflicts were resolved last week.  Thanks @eslavich !

jdavies-st:
Btw, the failing test

https://travis-ci.org/spacetelescope/asdf/jobs/617425043#L881

Is it possible it is counting the number times the `url_mapping` is getting called based on the number of URLs or tags in the ASDF file? It's very possible that has changed in this PR?

perrygreenfield:
Oddly enough, one of the changes in this PR changes the expected number of _nreads from 5 to 6. If left at the original number, apparently this test would pass. Comments @eslavich ?

jdavies-st:
Hmm, that suggests it was a moving target?  Was it failing before and that's why you changed it Ed?

eslavich:
I chatted with @jdavies-st about the _nreads failure.  I don't have a clear memory of this change, but I suspect it was just a mistake.  I didn't notice when running the tests locally because I didn't think to add the `--remote-data` flag.

asdf/resolver.py
```diff
@@ -48,63 +50,104 @@ def __init__(self, mapping=[], prefix=''):
...
+        for resolver in self._resolvers:
+            input = resolver(input)
+        return input
+
+    def __hash__(self):
+        return hash(self._resolvers)
+
+    def __eq__(self, other):
+        if not isinstance(other, ResolverChain):
+            return NotImplemented
```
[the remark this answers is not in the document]
eslavich: (we discussed this, the default implementation of `__ne__` is just to invert the result from `__eq__`, so we should be covered here)

nden:
`Dev` Travis job passes now, merging.

pllim:
The deprecation here broke `astropy` tests (astropy/astropy#9816). What is the non-deprecated way to call `default_tag_to_url_mapping`? Thanks.

PR #686, opened 2019-07-10
Update license year

pllim:
But really, I want to check new bot behavior. Still, this is a real PR.


pllim:
@rendinam , bot works here! 👍

PR #692, opened 2019-08-16
Tighten PEP8 checks, clean up how version is handled.

jdavies-st:
- Check all F and W codes via `flake8`.
- Remove `_internal_init.py` and do things in the standard way in `__init__.py`.


jdavies-st:
This PR should probably be replaced with 2 separate PRs:

1. Tighten `flake8` checks for `F` and `W` codes
2. Remove `_internal_init.py` and do things in the standard way in `__init__.py`.

And don't make any changes (yet) with how __version__ is handled is built.

PR #697, opened 2019-08-28
fix classifiers so that pypi accepts the package

nden:
Uploading to PyPi of 2.4.0 failed with an error
```
Invalid value for classifiers. Error: ''Programming Language :: Python'' is not a valid choice for this field
```
This fixes the classifiers (works on testpypi now).


Cadair:
The amount of times this catches me out...

PR #700, opened 2019-08-29
limit semantic_version to <=2.6

nden:


olebole:
Is there a specific reason for that?
In Debian,we have now now semantic_version 2.8.2,which makes it impossible to create the Debian package here.

nden:
Well, in general, it should be possible to relax. The reason was that at the time of the previous asdf release `semantic_versioning` changed rapidly in a few days - there were four pypi releases in two days. `SpecItem` is deprecated in these and deprecation warnings raise an error in astropy. So it was a stop-gap measure. And we haven't had time to figure out what replaces `SpecItem`. I guess another stop-gap measure would be to disallow 3.x releases and ignore the DeprecationWarning.

olebole:
Since asdf is used by other packages (like sunpy) , it should not restrict the versions of its dependencies by itself too much, otherwise there is the danger of conflicts. And `SpecItem` is [mentioned being private](https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.SpecItem).
Isn't `SimpleSpec` (or `BaseSpec`) the right replacement here?

nden:
Perhaps. I'm not familiar with the reasons `SpecItem` was used. I'll take a look. Thanks for the pointer.

PR #704, opened 2019-09-09
Updated to the latest asdf-standard

nden:
- Updated the supported versions 
- Updated to the latest asdf-standard (to pick up ortho-polynomial)


nden:
@perrygreenfield If you look at the Travis jobx, the test with astropy dev passes and the tests with astropy stable fail expectedly.
I think this is ready for review and merging.

PR #705, opened 2019-09-19
Remove obsolete ez_setup.py

embray:
This used to be needed to bootstrap setuptools on systems that didn't have it.  But there is no modern Python distribution that doesn't include setuptools (virtualenv, conda, etc. all have it by default, as well as anything that has pip installed).  It's also supplanted by pyproject.toml.

So no reason to keep this in the repository.  I don't think it even works anymore.


nden:
Thanks @embray !

PR #706, opened 2019-09-30
Skip schema test of ortho_polynomial-1.0.0

jdavies-st:
Since `ortho_polynomial-1.0.0` tag is defined in dev version of `astropy`, it is currently failing schema checks when run against the current release version of `astropy`.  This PR disables testing of the schema.

Once `astropy 4.0.0` is out, then the test can be re-enabled.

If any changes are made to this schema in the meantime, be sure to back out this change and make sure it gets tested against the appropriate dev version of `astropy`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ open_files_ignore = test.fits asdf.fits
 text_file_format = rst
 # Account for both the astropy test runner case and the native pytest case
 asdf_schema_root = asdf-standard/schemas asdf/schemas
-asdf_schema_skip_names = asdf-schema-1.0.0 draft-01
+asdf_schema_skip_names = asdf-schema-1.0.0 draft-01 ortho_polynomial-1.0.0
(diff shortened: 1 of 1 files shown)


nden:
There's one other new schema - fix_inputs.

However, since this is not conditional on version this means it never get tested, even with astropy dev, right?

nden:
Actually, the asdf-standard pointer is probably not updated yet to include the "fix_inputs" schema.

jdavies-st:
Yeah, we will have to update the `asdf-standard` submodule.  We can do that first and then update this PR if you like.

When merging this PR, we should file an issue to turn the testing back on once `astropy 4.0` is out.

nden:
We discussed this with @jdavies-st and decided it's better to make tests with public releases allowed to fail and require tests with astropy-dev to pass. We know the asdf release works with the astropy release and asdf master will work with astropy dev only.

jdavies-st:
Yes, closing this PR.

PR #712, opened 2019-10-21
update pointer to asdf-standard

nden:


nden:
This PR should pass with astropydev and fail with astropy release. Which it does.

PR #715, opened 2019-11-15
replace SpecItem with SimpleSpec

### LINKED ISSUE #702: semantic_version.SpecItem is deprecated
`semantic_version.Specitem` is deprecated in v 2.8 and will be removed in v 3.0. Astropy raises the `DeprecationWarning` as an error. To work around this `semantic_version` was limited in setup.py to <=2.6 but the code in `asdf.versioning.py` needs a replacement of the deprecated object.

nden:
Fixes #702

This PR replaces `SpecItem` with `SimpleSpec` (both from `semantic_version`).
It also fixes the semantic_version to >= 2.8 . This is a backwards incompatible change in `semantic_version` so if we need to support earlier versions more changes will be needed.
@olebole Do you think support for earlier versions is needed?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ package_dir =
     asdf.reference_files = asdf-standard/reference_files
 include_package_data = True
 install_requires =
-    semantic_version>=2.3.1,<=2.6.0
+    semantic_version>=2.8
     pyyaml>=3.10
     jsonschema>=2.3,<4
     six>=1.9.0
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@
 
 - Added asdf-standard 1.4.0 to the list of supported versions. [#704]
 
+- ``SpecItem`` and ``Spec`` were depreicated  in ``semantic_version``
+  and were replaced with ``SimpleSpec``. [#715]
+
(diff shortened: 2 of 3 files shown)


olebole:
I don't need support for <2.8. If one would backport asdf (to Debian Buster f.e.), they would just need to backport semantic-version as well, which is acceptable. So, I am in favour for this PR.
And, I *very* like the good cooperation with you! That is great and much fun!!!

nden:
Note: The only valid test here is the one with astrodev (passing). The rest fail because they use the public release of astropy and references to new models are missing.

setup.cfg
```diff
@@ -31,7 +31,7 @@ package_dir =
     asdf.reference_files = asdf-standard/reference_files
 include_package_data = True
 install_requires =
-    semantic_version>=2.3.1,<=2.6.0
+    semantic_version>=2.8
```
jdavies-st: Could this be relaxed to 2.7, when `SimpleSpec` was introduced?  Or were functional changes made in 2.8 that we need?

nden: 2.7 was a short lived release, looked like it wasn't stable.

jdavies-st: Ah ok.  👍

PR #716, opened 2019-12-04
Refresh NDArrayType when underlying mmap is closed

eslavich:
This is the start of a fix for https://github.com/spacetelescope/asdf/issues/711.

When an asdf file is updated, we call [calculate_updated_layout](https://github.com/spacetelescope/asdf/blob/cfa6a039d2e8c5dc93f2fd52e7d0faa9c7b16d51/asdf/block.py#L1253), which shuffles blocks around the file as needed to make room for the expansion of the tree and changes to the other blocks.  When a block gets moved, its underlying memory map is closed and then reopened in the new location.

This is a problem for NDArrayType, which keeps a [memoized copy](https://github.com/spacetelescope/asdf/blob/d8d8b3b62b7a3a6e0ce06dc83708b4c4fb807935/asdf/tags/core/ndarray.py#L251) of an ndarray that is built on top of that mmap.  The segfault occurs when the tree calls `__repr__` on the ndarray.

This PR gives NDArrayType the ability to detect that the block has been moved, and rebuild the ndarray in that case.  It does fix the issue as reported but it might be worth revisiting whether the optimizations that cause this problem are even necessary?  Maybe we don't need to memoize the ndarray at all, or maybe it's not really necessary to be able to update a small part of the asdf file without rewriting the entire thing.

### DIFF
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -222,7 +222,7 @@ class NDArrayType(AsdfType):
         self._asdffile = asdffile
         self._source = source
         self._block = None
-        self._block_data_ref = None
+        self._block_data_weakref = None
         self._array = None
         self._mask = mask
 
@@ -246,11 +246,16 @@ class NDArrayType(AsdfType):
             self._make_array()
 
     def _make_array(self):
-        if self._block_data_ref is not None:
-            block_data = self._block_data_ref()
+        # If the ASDF file has been updated in-place, then there's
+        # a chance that the block's original data object has been
+        # closed and replaced.  We need to check here and re-generate
+        # the array if necessary, otherwise we risk segfaults when
+        # memory mapping.
+        if self._block_data_weakref is not None:
+            block_data = self._block_data_weakref()
             if block_data is not self.block.data:
                 self._array = None
-                self._block_data_ref = None
+                self._block_data_weakref = None
 
         if self._array is None:
             block = self.block
@@ -259,7 +264,7 @@ class NDArrayType(AsdfType):
             self._array = np.ndarray(
                 shape, self._dtype, block.data,
                 self._offset, self._strides, self._order)
-            self._block_data_ref = weakref.ref(block.data)
+            self._block_data_weakref = weakref.ref(block.data)
             self._array = self._apply_mask(self._array, self._mask)
             if block.readonly:
                 self._array.setflags(write=False)
@@ -362,7 +367,7 @@ class NDArrayType(AsdfType):
             self._make_array().__setitem__(*args)
         except Exception as e:
             self._array = None
-            self._block_data_ref = None
+            self._block_data_weakref = None
             raise e from None
 
     @classmethod
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import sys
+import weakref
 
 import numpy as np
 from numpy import ma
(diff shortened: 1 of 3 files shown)


asdf/tags/core/ndarray.py
```diff
@@ -221,6 +222,7 @@ def __init__(self, source, shape, dtype, offset, strides,
         self._asdffile = asdffile
         self._source = source
         self._block = None
+        self._block_data_ref = None
```
jdavies-st: It might be nice to name this, indicating it is (or will be) a `weakref`.

asdf/tags/core/ndarray.py
```diff
@@ -244,13 +246,20 @@ def __init__(self, source, shape, dtype, offset, strides,
             self._make_array()
 
     def _make_array(self):
+        if self._block_data_ref is not None:
```
jdavies-st: A comment string something like
```
# Check if self.block.data has moved
```
or similar will help our future selves remember why this block of code is here.

asdf/tags/core/ndarray.py
perrygreenfield: I'm fairly sure I would not vote to remove update. It can be an essential function if someone is trying to update a small part of an existing file. Something I might consider is if an update requires rewriting most or all blocks, just write out a new copy of the file and delete the old. That might simplify the code. An option is just to make an existing block unused and add a new one at the end if it is being replaced. That grows the file with dead space, so it would be an option someone might explicitly choose in using update. Anyway, these are topics for a different PR...

sosey: Make a consideration also to how we might use the update in the AWS S3 environment, where update of the object store that represents the file is not allowed.. I don't believe S3 supports full CRUD.

I'm thinking here of the case for an exploded file where the blocks are stored externally. If we want to replace a block with updated data, and also need to update the reference to the block. It would be nice to update just the block, and the reference in the header instead of replacing and rewriting the entire file and it's associated blocks

perrygreenfield: @sosey I think you are basically saying that that situation doesn't support update. For example, if you use exploded files and need to update a block, you need to rewrite the whole block (right?). In that sense, update is only on the ensemble, but not on any file. Would some sort of exploded update tool serve to do that just as well?

sosey: If we are updating just one block, which exists externally, with a reference to it's location in the tree, I want to make sure that only the one block we are changing gets updated, and the reference to it is updated in the tree. 

I think to do this in S3 we need to delete the external block, write the new block, then delete the header object and write the new header object with everything the same except the new reference to the updated block.  

I think it's possible to have a mix of block storage types, what if data changes from internal to external. I want to not only have the case where a reference to an external array changes (because the data changed), but also the case where data that was stored internally is moved externally, for that case, it might make sense to optimize the internal array storage while updating the reference to the external data.

eslavich: I think the current state of affairs is that asdf rewrites every external block on update.  That happens here: https://github.com/spacetelescope/asdf/blob/master/asdf/block.py#L354

S3 doesn't permit updates to a range within an object, but you can overwrite an entire existing object in one operation, so it wouldn't be necessary to delete the old blocks/headers first.

asdf/tags/core/ndarray.py
eslavich: Added a comment

jdavies-st:
The following PEP8 issues should be fixed.  Then ready for merging I think.
```
asdf/tests/test_schema.py:194:1: W293 blank line contains whitespace
asdf/tags/core/tests/test_ndarray.py:879:8: E111 indentation is not a multiple of four
```

PR #722, opened 2019-12-19
Ignore failing schema tests on astropy < 4.0

eslavich:
It turns out there's already an example in pytest_asdf of ignoring a schema test for older versions of astropy.


nden:
Awesome! Thanks!

PR #728, opened 2019-12-31
Add 'package' property to ExtensionMetadata and deprecate 'software'

eslavich:
This renames the extension metadata `software` property to `package` to match the ASDF standard.  Files written by older versions of the library will read successfully, but emit a deprecation warning.

Addresses https://github.com/spacetelescope/asdf/issues/709


asdf/asdf.py
```diff
@@ -28,7 +28,7 @@
 from .exceptions import AsdfDeprecationWarning
 from .extension import AsdfExtensionList, default_extensions
 
-from .tags.core import AsdfObject, Software, HistoryEntry, ExtensionMetadata
+from .tags.core import AsdfObject, Software, HistoryEntry, ExtensionMetadata, Software
```
perrygreenfield: OK, I give up. How does importing Software twice change things? :-)

asdf/tags/core/__init__.py
```diff
@@ -25,9 +26,17 @@ class ExtensionMetadata(AsdfType):
     name = 'core/extension_metadata'
     version = '1.0.0'
 
-    def __init__(self, extension_class=None, software={}):
+    def __init__(self, extension_class=None, package=None, software=None):
+        if software is not None:
+            warnings.warn("Previous versions of the ASDF library stored a 'software' property " +
+                          "in extension metadata.  That property has been renamed to 'package' " +
```
perrygreenfield: The use of the term property here appears ambiguous. It wasn't strictly a Python property previously, though it is now.

asdf/asdf.py
eslavich: Ha!  It's, uh, a test for code review quality.  Like Van Halen's brown M&Ms.

asdf/tags/core/__init__.py
eslavich: Good point, would "attribute" or "field" be better?

nden:
I am going to merge this so we can see the jwst regression tests tonight.

asdf/tags/core/__init__.py
perrygreenfield: Yes, probably. Maybe something like "...saved this information in the .software attribute. That value is now saved in the .package attribute"

PR #735, opened 2020-01-08
Add a collection-like object of ExternalArrayReference objects

Cadair:
This is mainly motivated by the fact that I have a large number of `ExternalArrayReference` objects in my trees and because each one is a separate schema they all have to be validated. Not to mention that it takes up a lot of lines of yaml. I am hoping this will massively improve my load performance.

### DIFF
--- a/asdf/__init__.py
+++ b/asdf/__init__.py
@@ -39,7 +39,7 @@ from .extension import AsdfExtension
 from .stream import Stream
 from . import commands
 from .tags.core import IntegerType
-from .tags.core.external_reference import ExternalArrayReference
+from .tags.core.external_reference import ExternalArrayReference, ExternalArrayReferenceCollection
 
 from jsonschema import ValidationError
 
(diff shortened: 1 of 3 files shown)


eslavich:
Thanks for the PR!  I'm hesitant to incorporate this into the core library/standard, since it seems special-purpose and I'm not aware of any wider demand for this kind of feature.  What do you think about creating an extension instead?  The ASDF extension mechanism is how we handle serialization of custom objects here at STScI.  Documentation on writing extensions can be found here:

https://asdf.readthedocs.io/en/latest/asdf/extensions.html

We'd be happy to support creating your extension with code reviews/troubleshooting/etc.

Alternatively, we have an [open issue](https://github.com/spacetelescope/asdf/issues/565) to allow users to skip validation.  If you're creating the files and validating on write anyway, it may be reasonable to disable the read validation outright.  We're planning to add that feature within the next couple of months.

Cadair:
Hi @eslavich 

I have already implemented this as an extension in my DKIST tools, along with a bunch of DKIST specific tags.

I actually added the `ExternalArrayReference` type for this use case to core at the start, and thought that this collection might also be generically useful.

PR #737, opened 2020-01-23
Try adding s390x architecture to Travis build

eslavich:


eslavich:
In case anyone's curious, the issue turned out to be a bug in Travis's s390x container -- packages are failing to install because part of the pip cache is owned by root:

```
$ ls -lsAht ~/.cache/pip/
total 13K
1.5K drwxr-xr-x 10 root   root   10 Jan 20 10:45 wheels
 512 drwxrwxr-x  7 travis travis  7 Jan 20 10:42 http
 11K drwxrwxr-x  2 travis travis  3 Jan 20 10:42 selfcheck
```

For comparison, here is what an AMD64 container looks like:
```
$ ls -lsAht ~/.cache/pip/
total 12K
4.0K -rw-r--r--  1 travis travis  550 Mar 25  2019 selfcheck.json
4.0K drwxr-xr-x  3 travis travis 4.0K Mar 25  2019 wheels
4.0K drwx------ 15 travis travis 4.0K Mar 25  2019 http
```

I'll submit a report to Travis, but in the meantime we can install with `--no-cache-dir` on s390x.

eslavich:
Closing this in favor of #740

PR #738, opened 2020-01-24
Resolve external references in custom schemas

eslavich:
Previously, external references in custom schemas were not being resolved (see https://github.com/spacetelescope/asdf/issues/683).  This was clearly deliberate, though I can't think of a reason why that would be the desired behavior.  This PR changes the custom schema behavior to resolve external references, and fixes an issue in the tests that prevented the example custom schema from resolving correctly.

Resolves https://github.com/spacetelescope/asdf/issues/683

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -109,17 +109,15 @@ class AsdfFile(versioning.VersionedMixin):
             standard.
 
         """
+        self._extensions = []
+        self._extension_metadata = {}
+        self._process_extensions(extensions)
 
         if custom_schema is not None:
-            self._custom_schema = schema.load_custom_schema(custom_schema)
-            schema.check_schema(self._custom_schema)
+            self._custom_schema = schema.load_schema(custom_schema, self.resolver, True)
         else:
             self._custom_schema = None
 
-        self._extensions = []
-        self._extension_metadata = {}
-
-        self._process_extensions(extensions)
         self._ignore_version_mismatch = ignore_version_mismatch
         self._ignore_unrecognized_tag = ignore_unrecognized_tag
         self._ignore_implicit_conversion = ignore_implicit_conversion
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -346,23 +346,12 @@ HARDCODED_SCHEMA = {
 
 @lru_cache()
 def load_custom_schema(url):
-    # Avoid circular import
-    from .tags.core import AsdfObject
-    custom = load_schema(url, resolve_local_refs=True)
-    core = load_schema(AsdfObject.yaml_tag)
-
-    def update(d, u):
-        for k, v in u.items():
-            # Respect the property ordering of the core schema
-            if k == 'propertyOrder' and k in d:
-                d[k] = u[k] + d[k]
-            elif isinstance(v, Mapping):
-                d[k] = update(d.get(k, {}), v)
-            else:
-                d[k] = v
-        return d
-
-    return update(custom, core)
+    warnings.warn(
+        "The 'load_custom_schema(...)' function is deprecated. Use"
+        "'load_schema' instead.",
+        AsdfDeprecationWarning
+    )
+    return load_schema(url, resolve_references=True)
 
 
 def load_schema(url, resolver=None, resolve_references=False,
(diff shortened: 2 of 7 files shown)


asdf/asdf.py
```diff
@@ -109,17 +109,15 @@ def __init__(self, tree=None, uri=None, extensions=None, version=None,
...
 
         """
+        self._extensions = []
+        self._extension_metadata = {}
+        self._process_extensions(extensions)
 
         if custom_schema is not None:
-            self._custom_schema = schema.load_custom_schema(custom_schema)
-            schema.check_schema(self._custom_schema)
+            self._custom_schema = schema.load_schema(custom_schema, self.resolver, True)
```
eslavich: This was moved after the call to self._process_extensions so that we could use our AsdfFile's resolver when loading the custom schema.

asdf/tests/data/custom_schema_definitions.yaml
```diff
@@ -1,9 +1,7 @@
 %YAML 1.1
 ---
 $schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
-id: "http://stsci.edu/schemas/asdf/core/asdf-1.1.0"
```
eslavich: Re-using the core schema id causes problems when resolving external references.  I don't think it's a good idea anyway, since the ids are meant to be globally unique.

asdf/schema.py
```diff
@@ -346,23 +346,12 @@ def _load_draft4_metaschema():
 
 @lru_cache()
 def load_custom_schema(url):
-    # Avoid circular import
-    from .tags.core import AsdfObject
-    custom = load_schema(url, resolve_local_refs=True)
-    core = load_schema(AsdfObject.yaml_tag)
-
-    def update(d, u):
```
eslavich: I don't think it's actually necessary to merge the core schema with the custom schema, since we already check against the core schema in a first validation pass.

asdf/tests/test_schema.py
```diff
@@ -775,6 +775,64 @@ def test_custom_validation_with_definitions_bad(tmpdir):
             pass
 
 
+def test_custom_validation_with_external_ref_good(tmpdir):
+    custom_schema_path = helpers.get_test_data_path('custom_schema_external_ref.yaml')
+    asdf_file = os.path.join(str(tmpdir), 'out.asdf')
```
jdavies-st:
```
asdf_file = str(tmpdir.join("out.asdf"))
```
is the more compact way to do this, using the `join()` method on the path object supplied by the `tmpdir` fixture.

asdf/tests/test_schema.py
```diff
@@ -775,6 +775,64 @@ def test_custom_validation_with_definitions_bad(tmpdir):
...
+
+    # This tree conforms to the custom schema
+    tree = {
+        'foo': asdf.tags.core.Software(name="Microsoft Windows", version="95")
+    }
+
+    with asdf.AsdfFile(tree, custom_schema=custom_schema_path) as ff:
+        ff.write_to(asdf_file)
+
+    with asdf.open(asdf_file, custom_schema=custom_schema_path) as ff:
```
jdavies-st: `as ff` is not needed here as it you don't use that in the `with` block below.  This might actually trigger a PEP8 warning (if we're checking)

asdf/tests/test_schema.py
```diff
@@ -775,6 +775,64 @@ def test_custom_validation_with_definitions_bad(tmpdir):
...
+        'foo': False
+    }
+
+    # Creating file without custom schema should pass
+    with asdf.AsdfFile(tree) as ff:
+        ff.write_to(asdf_file)
+
+    # Creating file with custom schema should fail
+    with pytest.raises(ValidationError):
+        with asdf.AsdfFile(tree, custom_schema=custom_schema_path) as ff:
```
jdavies-st: Again, `as ff` is not needed as the `with` block just contains as `pass`.

asdf/tests/test_schema.py
eslavich: Nice, I'll update this

asdf/tests/test_schema.py
eslavich: Oops, I think it would, I'll change these too

PR #739, opened 2020-01-24
Remove six dependency

eslavich:
This removes our dependency on six, which shouldn't be necessary since we dropped Python 2 support.

Resolves https://github.com/spacetelescope/asdf/issues/724


pllim:
Thanks!

PR #750, opened 2020-02-13
Add versioning documentation and tests, fix versioning issues

eslavich:
This PR updates the asdf-standard pointer to latest master, and adds some tests to ensure that each of the ASDF Standard versions are fully supported.  Happily, only one of the core schemas was updated since 1.0.0.  Here's a summary of the changes:

- Fix handling of history entries for ASDF Standard 1.0.0.  Previously we had claimed to support 1.0.0, but bugs in the `AsdfType` subclass definitions resulted in corrupted 1.0.0-ish files being written.  Probably we should drop this version of the standard in library 3.0.
- Separate `AsdfObjectType` from `AsdfObject`.  This is the one core schema that received an update, and it turns out the library doesn't gracefully handle `AsdfType` classes that support multiple versions and (de)serialize themselves.  I think we should separate the `AsdfType` from the serialized class in all cases, but I'll open a separate issue to discuss that.
- Add explicit versions to all `AsdfType` subclasses.  There is a default 1.0.0, but I think it's safer to be explicit -- relying on the default is what got us into trouble with CompoundModel serialization.  I'll open an additional issue to consider deprecating the defaults so that users are forced to set a version.
- Fix an issue in a test where a new AsdfType was registered, which caused heisenbugs in downstream tests.
- Fix the yaml_to_asdf test helper, which conflated the ASDF file version with the standard version, and did not correctly set the tree's tag for all standard versions.
- Change schema validator to use the latest standard version that supports the tested schema (instead of the earliest).
- Add a new `test_extension_correctness` test helper that checks an extension's types for the correct features (version, name, etc) and confirms that the extension is able to resolve each type's schema.  We should be able to use this helper in gwcs and astropy.
- Test `BuiltinExtension` with the new helper
- Add tests for each of the supported ASDF Standard versions that confirm availability of the standard's component schemas, and confirms that an `AsdfType` exists to support each schema.  These tests pass for the core schemas, but are xfail'ed for the "standard" schemas that are supported by astropy, which hasn't been updated yet.
- Exclude `AsdfType` itself from `BuiltinExtension`'s types, as it is an abstract base class with no schema and can't actually be written to an ASDF file.
- Generalize the code that dumps trees to YAML, which previously had only set a tag namespace when the tree was a self-serializing ASDF core object.
- Add a new documentation page on the joys of versioning

### DIFF
--- a/docs/asdf/developer_versioning.rst
+++ b/docs/asdf/developer_versioning.rst
@@ -133,6 +133,11 @@ it proceeds as follows:
   and AsdfType are available for 1.1.0 and 1.3.0, it will
   use the 1.1.0 subclass.
 - If the above fails, use the earliest available AsdfType
+- If no AsdfType exists that supports any version of that tag,
+  then ASDF will deserialize the data into vanilla diff.
+
+The library does not currently emit a warning in either of the
+first two cases, but in the third case, a warning is emitted.
 
 The rules for selecting an AsdfType for a given tag are implemented
 by ``asdf.type_index.AsdfTypeIndex.fix_yaml_tag``.
@@ -142,6 +147,10 @@ to the ASDF Standard version in use, which dictates the subset of
 schema versions that are available.  From the subset of AsdfType
 subclasses that handle those schema versions, it selects the subclass
 that is able to handle the type of the core object being serialized.
+
+If an object is not supported by an AsdfType, its serialization will be
+handled by pyyaml.  If pyyaml doesn't know how to serialize, it will
+raise ``yaml.representer.RepresenterError``.
 
 The rules for selecting an AsdfType for a given serializable object
 are implemented by ``asdf.type_index.AsdfTypeIndex.from_custom_type``.
--- a/docs/asdf/developer_versioning.rst
+++ b/docs/asdf/developer_versioning.rst
@@ -1,0 +1,253 @@
+Schema Versioning and You
+=========================
+
+Here we'll explore ASDF schema versioning, and walk through the process
+of supporting new and updated schemas with AsdfType subclasses.
+
+ASDF versioning conventions
+---------------------------
+
+The ASDF Standard document provides a helpful overview_ of the various ASDF
+versioning conventions.  We will be concerned with the *standard version*
+and individual *schema versions*.
+
+.. _overview: https://asdf-standard.readthedocs.io/en/latest/versioning.html
+
+Overview
+--------
+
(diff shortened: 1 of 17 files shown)


docs/asdf/developer_versioning.rst
```diff
@@ -0,0 +1,253 @@
...
+class attribute or ``supported_versions``.  If that is not possible,
+it proceeds as follows:
+
+- Use the AsdfType that supports the latest version that is
+  less than the tag version.  For example, if the tag is example-1.2.0,
+  and AsdfType are available for 1.1.0 and 1.3.0, it will
+  use the 1.1.0 subclass.
+- If the above fails, use the earliest available AsdfType
+
+The rules for selecting an AsdfType for a given tag are implemented
```
perrygreenfield: Might be good to summarize when and if warnings are issued when exact matches are not found and it assumes the earliest.

docs/asdf/developer_versioning.rst
```diff
@@ -0,0 +1,253 @@
...
+version, so that this class can continue to handle it:
+
+.. code-block:: python
+
+    class UpdatedObjectType(AsdfType):
+        name = "core/updated_object"
+        version = "1.1.0"
+        supported_versions = {"1.0.0", "1.1.0"}
+
+Support the breaking schema
```
perrygreenfield: I wonder if we shouldn't add a statement suggesting that backward incompatible changes should be highly discouraged. I can't say that they should never be permitted (I suspect we can think of examples), but sometimes a new type object (and core tag name) may be much more preferable than doing this. I guess the issue comes down to when we are creating a Python object that we do not want to change the class name, but do want to change how it is serialized in a backward incompatible way. Since the class type is used to find the AsdfType, does he machinery support selecting between two different AsdfTypes depending on some indication of which now is most recent when writing? Does what I'm saying make any sense? :-)

eslavich: Happily, it does support selecting between different AsdfTypes!  The first thing the library does on write is consult the version map for relevant schema versions, and then select the subset of AsdfTypes that support those versions.  So we can have two AsdfTypes that (de)serialize the same type of object, and as long as they specify different schema versions, they won't conflict with one another.

docs/asdf/developer_versioning.rst
eslavich: That's a good idea, I don't know the answer but I'll look into it

eslavich: (added documentation on the warnings)

PR #758, opened 2020-02-19
Test for ASDF file compatibility between library versions

eslavich:
This adds a new test to check that files produced by the current version of the code are compatible with older versions of the library (and vice versa).  I confirmed that it would have caught the issue with https://github.com/spacetelescope/asdf/pull/728.


nden:
This looks great. Would it make sense in the future to generate files from the examples in a schema? The specific problem we encountered was general for all tags. I wonder if it makes sense to check specific tags in the future.

compatibility_tests/common.py
```diff
@@ -0,0 +1,17 @@
...
+
+def generate_file(path, version):
+    if version not in supported_versions:
+        raise ValueError("ASDF Standard version {} is not supported by version {} of the asdf library".format(version, asdf.__version__))
+
+    af = asdf.AsdfFile({"array": np.ones((8, 16))}, version=version)
+    af.write_to(path)
+
+
+def assert_file_correct(path):
```
jdavies-st: You may want to add `__tracebackhide__ = True` to this assertion helper so that the code in the helper doesn't become part of the traceback when the test fails.  You just want to see the assertion failure rewrite in pytest.

compatibility_tests/test_file_compatibility.py
```diff
@@ -0,0 +1,184 @@
...
+
+    assert env_run(path, "pip", "install", "asdf=={}".format(asdf_version), capture_output=True), (
+        "Failed to install asdf version {}".format(asdf_version)
+    )
+
+    return path
+
+
+@pytest.fixture(autouse=True)
+def change_tmpdir(tmpdir):
```
jdavies-st: Perhaps this could function could be called `pushd`.?

eslavich: Haha, I can hardly [disagree](https://github.com/spacetelescope/jwst/pull/4567#discussion_r378637258)

compatibility_tests/common.py
eslavich: Yessir, thanks!

Issue #762, opened 2020-02-19
simpler example for creating arrays in README.rst


perrygreenfield:
Peter Teuben points out that the example on readthedocs (which comes from the README.rst) of the following:

```
sequence = np.array([x for x in range(100)])
squares = np.array([x**2 for x in range(100)])
```

is better expressed as 

```
sequence = np.arange(100)
squares  = sequence**2
```

It is hard to argue with that.

nden:
Fixed in #763

PR #763, opened 2020-02-19
simplify example

perrygreenfield:
Address issue [#762]


perrygreenfield:
Hmmm, an update to asdf-standard got into this; how should that be handled?

eslavich:
It looks like the update to the `asdf-standard` pointer pushes it back to an earlier commit than `asdf` currently has.  I think if you set it to `7b4a28fccac950cdfee4cac3d4fa690222a9f23e` then there won't be a diff.

PR #764, opened 2020-02-24
Update asdf-standard pointer to add missing 1.4.0 schemas

eslavich:
This advances the asdf-standard commit pointer to the current master, which I'll use in the 2.5.2 release.


eslavich:
Oops, I just realized I pushed my branch to spacetelescope/asdf instead of eslavich/asdf.  I'm going to delete this branch and re-open the PR correctly.

PR #771, opened 2020-03-05
Add pytest plugin option to ignore unrecognized tags

eslavich:
This is needed so that schemas examples can be tested before the schema's ExtensionType exists (or, tested when the package containing the ExtensionType isn't installed).


eslavich:
I'm going to merge this early so we can move forward with @shannnnnyyy's asdf-standard work, I'll address any comments in an additional PR.

eslavich:
> So it looks like this PR adds the plumbing for the pytest configuration and propagates it to the functions that need it. But where's the actual tag ignoring functionality implemented?

Happily, that was already implemented in asdf, as a flag that is passed to the `AsdfFile` initializer.  Once I propagate the flag to that point, the existing code takes over.  The bit that ignores the warning if the flag is set is here:

https://github.com/spacetelescope/asdf/blob/master/asdf/yamlutil.py#L254

jdavies-st:
Ah cool!  I bet you were happy to find it already implemented!  😂

PR #773, opened 2020-03-18
Fix generic_io read_blocks() reading past the requested size

lgarrison:
Hi there,

I was looking at some anomalies in the lz4 decompression performance, which led me in a roundabout way to find this bug.  Basically, `GenericFile.read_blocks()` was reading more than the requested amount, causing the decompressor to start reading an extra compression block, the first int of which is the size of the compression block.  The lz4 decompressor would then allocate a `bytearray` of this size, which is zero-initialized.  So it was sometimes doing a `memset()` of up to 4 GB, which is a noticeable performance hit!  I think this could also have resulted in corrupted data reads, but that would require the "random" int value to be less than 4K (the file system block size).  I think all the other compressors were at similar risk, depending on how big their internal buffers were.

Thanks for developing this package, by the way—I've found it very useful for my work!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 
 - Add pytest plugin option to skip warning when a tag is
   unrecognized. [#771]
+  
+- Fix generic_io `read_blocks()` reading past the requested size [#773]
 
 2.5.2 (2020-02-28)
 ------------------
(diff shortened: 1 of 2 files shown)


eslavich:
Thanks for the fix @lgarrison!  And for completing what sounds to be an epic debugging quest.  

Don't worry about the failing tests, I'll figure that out before I merge your PR.  I imagine something else in this library is coded to compensate for the incorrect read_blocks implementation.

lgarrison:
Ah, hang on, looks like the debugging quest isn't over!  I think my fix fails when the requested size is less than the block size.  Let me try again...

PR #777, opened 2020-04-06
Support (de)serialization of custom objects containing implicit references

### LINKED ISSUE #744: Serialization of models with circular definitions of user inverses fails
This is essentially xref to https://github.com/spacetelescope/tweakwcs/issues/97

Example of a failure:

```
from asdf import AsdfFile
from astropy.modeling.models import Mapping, Identity, Const1D
c2tan = ((Mapping((0, 1, 2)) / Mapping((0, 0, 0), n_inputs=3)) | Mapping((1, 2)))
tan2c = (Mapping((0, 0, 1), n_inputs=2) | (Const1D(1) & Identity(2)))
c2tan.inverse = tan2c
tan2c.inverse = c2tan # <- comment this out to make the model serializable
AsdfFile({'c2tan': c2tan})  # recursion limit error
```

In my opinion `tan2c.inverse = c2tan` should work.

### LINKED ISSUE #745: Recursive references serialize and deserialize incorrectly
A simple example from a discussion with @jdavies-st:

```python
In [16]: target = {"foo": "bar"}
    ...: target["nested"] = target
    ...: af = asdf.AsdfFile(tree)
    ...: af.info()
root.tree (AsdfObject)
└─target (dict)
  ├─foo (str): bar
  └─nested (dict)
    ├─foo (str): bar
    └─nested (dict) (recursive reference)
```
Surprise!  Somehow we ended up with more distinct objects than we started with.  The display from info() is correctly depicting the situation:
```python
In [17]: af["target"] is af["target"]["nested"]
Out[17]: False

In [18]: af["target"]["nested"] is af["target"]["nested"]["nested"]
Out[18]: True
```
That's fine, we can create the recursive tree another way:
```python
In [20]: target = {"foo": "bar"}
    ...: target["nested"] = target
    ...: af = asdf.AsdfFile()
    ...: af.tree["target"] = target
    ...: af.info()
root.tree (AsdfObject)
└─target (dict)
  ├─foo (str): bar
  └─nested (dict) (recursive reference)
```
That's more like it.  Now serialize:
```python
In [21]: af.write_to("test.asdf")
    ...: print(open("test.asdf").read())
```
```
#ASDF 1.0.0
#ASDF_STANDARD 1.4.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
  name: asdf, version: 2.5.2a1.dev15+g90e5a4e.d20200130}
history:
  extensions:
    package: !core/software-1.0.0 {name: asdf, version: 2.5.2a1.dev15+g90e5a4e.d20200130}
target:
  foo: bar
  nested: &id001
    foo: bar
    nested: *id001
...
```
The object was serialized with YAML anchors and aliases, which is legal but unexpected, since the asdf standard recommends using `$ref` instead.  When we ask asdf to read the file again:
```python
asdf.open("test.asdf").info()
```
```
root.tree (AsdfObject)
├─asdf_library (Software)
│ ├─author (str): Space Telescope Science Institute
│ ├─homepage (str): http://github.com/spacetelescope/asdf
│ ├─name (str): asdf
│ └─version (str): 2.5.2a1.dev15+g90e5a4e.d20200130
├─history (dict)
│ └─extensions (list)
│   └─[0] (ExtensionMetadata)
└─target (dict)
  ├─foo (str): bar
  └─nested (dict)
    ├─foo (str): bar
    └─nested (dict)
      ├─foo (str): bar
      └─nested (dict)
        ├─foo (str): bar
        └─nested (dict) (recursive reference)
```

It did eventually figure out the recursive link, but only after running in circles an extra 3 times.

So a few things to fix/resolve here:
- Building a file from a tree with recursive references shouldn't add an extra layer of unique objects
- Should trees be serialized with `$ref` instead of anchors/aliases?  What happens when you deserialize a tree with a `$ref`?
- Trees with anchors/aliases should deserialize correctly

This is related to #744 but is a distinct issue.

eslavich:
This PR fixes a variety of issues related to serialization/deserialization of "implicit references", that is to say, references to the same object id that aren't explicitly wrapped in `asdf.reference.Reference`.  Changes include:

- Fix infinite loops and preserve implicit references through calls to `walk_and_modify` (this is what fixes `custom_tree_to_tagged_tree` and `tagged_tree_to_custom_tree` as well as loading schemas containing reference cycles).
- Fix infinite loops when deserializing reference cycles in our custom YAML loader
- Fix infinite loops when validating trees containing reference cycles in our custom jsonschema validator
- Eliminate need to call `custom_tree_to_tagged_tree` within extension code; instead, any nested custom types returned by a `ExtensionType` subclass will be automatically handled by asdf.  Re-entering `custom_tree_to_tagged_tree` is still supported at this time, but I intend to add a deprecation warning after fixing up other packages and a hard error in 3.0.
- Eliminate need to call `tagged_tree_to_custom_tree` within extension code; instead, any nested tagged objects will already be converted before the `ExtensionType` subclass receives a call to `from_tree_tagged`.  If an object isn't yet available due to a reference cycle, it will have the special value `PendingValue` in its place.  Re-entering `tagged_tree_to_custom_tree` will also be an error in 3.0.
- Allow `ExtensionType` subclasses to return a generator from `from_tree_tagged` and `to_tree_tagged` instead of a fully (de)serialized value.  This will allow the subclass to start building the object but yield it before setting a field that might lead back to the current node.  Taking advantage of this feature in astropy's `CompoundType` will allow us to handle @mcara's user inverses.

This should all be fully backwards compatible, though I still still need to run our dependent packages through their paces.

Resolves #745 
Resolves #744 

(but to fully resolve these issues I'll be opening another PR in astropy)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,11 @@
 - Enable validation and serialization of previously unhandled numpy
   scalar types. [#778]
 
+- Fix handling of trees containing implicit internal references and
+  reference cycles.  Eliminate need to call ``yamlutil.custom_tree_to_tagged_tree``
+  and ``yamlutil.tagged_tree_to_custom_tree`` from extension code,
+  and allow ``ExtensionType`` subclasses to return generators. [#777]
+
 2.5.2 (2020-02-28)
 ------------------
 
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -127,6 +127,11 @@ class AsdfFile(versioning.VersionedMixin):
 
         self._file_format_version = None
 
+        # Context of a call to treeutil.walk_and_modify, needed in the AsdfFile
+        # in case walk_and_modify is re-entered by extension code (via
+        # custom_tree_to_tagged_tree or tagged_tree_to_custom_tree).
+        self._tree_modification_context = treeutil._TreeModificationContext()
+
         self._fd = None
         self._closed = False
         self._external_asdf_by_uri = {}
@@ -1037,7 +1042,6 @@ class AsdfFile(versioning.VersionedMixin):
         if version is not None:
             self.version = version
 
-
         with generic_io.get_file(fd, mode='w') as fd:
             # TODO: This is not ideal: we really should pass the URI through
             # explicitly to wherever it is required instead of making it an
@@ -1058,12 +1062,8 @@ class AsdfFile(versioning.VersionedMixin):
         Finds all external "JSON References" in the tree and converts
         them to `reference.Reference` objects.
         """
-        # Since this is the first place that the tree is processed when
-        # creating a new ASDF object, this is where we pass the option to
-        # ignore warnings about implicit type conversions.
         # Set directly to self._tree, since it doesn't need to be re-validated.
-        self._tree = reference.find_references(self._tree, self,
-                ignore_implicit_conversion=self._ignore_implicit_conversion)
+        self._tree = reference.find_references(self._tree, self)
 
     def resolve_references(self, do_not_fill_defaults=False):
         """
@@ -1125,7 +1125,8 @@ class AsdfFile(versioning.VersionedMixin):
             if hook is not None:
                 return hook(node, self)
             return node
-        tree = treeutil.walk_and_modify(self.tree, walker)
+        tree = treeutil.walk_and_modify(self.tree, walker,
+            ignore_implicit_conversion=self._ignore_implicit_conversion)
 
         if validate:
             self._validate(tree)
--- a/asdf/reference.py
+++ b/asdf/reference.py
@@ -128,7 +128,7 @@ class Reference(AsdfType):
         pass
 
 
-def find_references(tree, ctx, ignore_implicit_conversion=False):
+def find_references(tree, ctx):
     """
     Find all of the JSON references in the tree, and convert them into
     `Reference` objects.
@@ -139,7 +139,7 @@ def find_references(tree, ctx, ignore_implicit_conversion=False):
         return tree
 
     return treeutil.walk_and_modify(
-        tree, do_find, ignore_implicit_conversion=ignore_implicit_conversion)
+        tree, do_find, ignore_implicit_conversion=ctx._ignore_implicit_conversion)
 
 
 def resolve_references(tree, ctx, do_not_fill_defaults=False):
@@ -154,7 +154,8 @@ def resolve_references(tree, ctx, do_not_fill_defaults=False):
 
     tree = find_references(tree, ctx)
 
-    return treeutil.walk_and_modify(tree, do_resolve)
+    return treeutil.walk_and_modify(
+        tree, do_resolve, ignore_implicit_conversion=ctx._ignore_implicit_conversion)
 
 
 def make_reference(asdffile, path):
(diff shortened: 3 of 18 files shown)


eslavich:
I added some missing code comments/docstrings, updated documentation on creating custom serializers, and added a test that includes a reference cycle between custom types.  Next I'm planning to run the gwcs, astropy, and jwst (unit and regression) tests with this branch.

eslavich:
For any asdf friends who weren't at our standup: some of the jwst regression tests failed due to jwst's use of a custom validator to populate fits headers.  This branch improves the efficiency of validation by skipping nodes that have already been validated, but this doesn't suit jwst because it needs to execute its callback on a node in each of its positions in the tree.  I added a flag that enables re-visiting nodes and am running the regression tests again.

eslavich:
With the latest commits and a small change to jwst (to use the new flag), the regression tests pass.

perrygreenfield:
I need to ask some stupid questions. Is this relying on pyyaml to generate the anchors (&)  and aliases (*)  automatically when the tree has shared objects? If not, where is that handled, and if so, what previously kept it from happening? I haven't found where this happens. Next stupid question: when deserializing, is pyyaml creating common nodes that have the same python id but are referenced at different places in the tree? (using anchors and aliases I presume).

perrygreenfield:
I wonder about the terminology of "draining generators" used in the comments. Is this a standard term? I ask since I wonder if it gives a misimpression that the generators aren't being asked to finalize. Or perhaps they need an oil change instead.

perrygreenfield:
I'm probably spending more time understanding how this all works than it took you to write it! It is an impressive amount of work.

eslavich:
> I need to ask some stupid questions. Is this relying on pyyaml to generate the anchors (&) and aliases (*) automatically when the tree has shared objects? If not, where is that handled, and if so, what previously kept it from happening? I haven't found where this happens. Next stupid question: when deserializing, is pyyaml creating common nodes that have the same python id but are referenced at different places in the tree? (using anchors and aliases I presume).

Yup, pyyaml is doing a lot of the heavy lifting here -- it automatically handles serializing the anchors/aliases and deserializing them back into a tree with shared objects.  This didn't work before because our `treeutil.walk_and_modify` was modifying the tree without preserving the "sharedness" of the objects, so by the time it got to pyyaml all the objects had distinct ids.

> I wonder about the terminology of "draining generators" used in the comments. Is this a standard term? I ask since I wonder if it gives a misimpression that the generators aren't being asked to finalize. Or perhaps they need an oil change instead.

I... seem have invented that term.  It does give the impression of prepping a small engine for winter.  Would "exhausting" generators work better?  Hmm, exhaust, more engine stuff.

> I'm probably spending more time understanding how this all works than it took you to write it! It is an impressive amount of work.

Heh, it only seems that way because I swept my sordid commit history under the rug...

eslavich:
It occurred to me to compare the runtime of the regression tests with this branch vs the nightly run on asdf 2.5.2.  These are the last three nightly runtimes:
```
4h 10min
4h 22min
4h 2min
```
And the times for my two runs with this branch:
```
3h 18min
3h 20min
```
How nice if that holds true!

PR #778, opened 2020-04-09
Enable validation/serialization of numpy scalar types

eslavich:
This PR enables validation and/or serialization of `numpy.str_`, `numpy.bytes_`, `numpy.floating`, and `numpy.integer` values.  These values will now all validate against the relevant jsonschema types, and serialize to YAML primitives.


eslavich:
(oops, wrong button)

PR #784, opened 2020-04-27
Improve validation performance by avoiding deep copy

eslavich:
When we fixed caching of schema objects in #682, we had to continue returning unique copies of those objects from `load_schema`, since a certain package (ahem, jwst) was mutating some schemas and thereby corrupting the cache.  That copy isn't necessary within asdf, since it doesn't mess with schemas once they're loaded, but at the time I didn't think to skip the copy step internally.

It turns out that copy adds a significant performance penalty in the validator, where schemas are loaded by tag for every tagged object in an asdf file.  For files with complex trees, making deep copies about doubles the read time!  So, this PR changes internal calls to `load_schema` to `_load_schema_cached`, which dodges the copy.

I tested this with the asdf file described [here](https://github.com/spacetelescope/asdf/issues/159#issuecomment-125716549) and on my laptop, the read time decreases from an average of 11.6 seconds to 5.7 seconds.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@
 - Fix bug causing duplicate elements to appear when calling
   ``copy.deepcopy`` on a ``TaggedList``. [#788]
 
+- Improve validator performance by skipping unnecessary step of
+  copying schema objects. [#784]
+
 2.6.0 (2020-04-22)
 ------------------
 
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -117,7 +117,7 @@ class AsdfFile(versioning.VersionedMixin):
         self._process_extensions(extensions)
 
         if custom_schema is not None:
-            self._custom_schema = schema.load_schema(custom_schema, self.resolver, True)
+            self._custom_schema = schema._load_schema_cached(custom_schema, self.resolver, True, False)
         else:
             self._custom_schema = None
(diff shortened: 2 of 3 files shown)


jdavies-st:
:tada:

Ed, can you file an issue over at `jwst` recommending not mutating the schema?  I believe it was in `exp_to_source`, though it is possible it is doing it at a lower level in `datamodels`.  I don't recall the details you found when investigating #682 .

The fundamental issue between the 2 packages is that `asdf` uses schemas to serialize/deserialize tagged objects.  That's the only purpose.  But `jwst` uses schemas to validate user data, so there's a bit of a different purpose in some sense. That said, `jwst` should be doing things correctly by `asdf`, especially if it is a performance penalty not to be doing so.

perrygreenfield:
Can you clarify if this requires JWST to do an explicit copy of the schema, or are they calling load_schema directly thus still forcing the copy?

eslavich:
JWST is calling load_schema directly, that happens here:

https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/model_base.py#L113

So I think this is a have-our-cake-and-eat-it-too situation.  JWST shouldn't see any performance impact from the copy since it only calls load_schema once per file, whereas asdf calls the function 10,004 times when reading my complex test file.

PR #785, opened 2020-04-27
Fix validation of default values in schemas

eslavich:
This PR addresses a bug in `check_schema` that has been preventing us from thoroughly validating default values in schemas.  Happily, the only schemas which seem to be invalid are the old wcs schemas, which are deprecated anyway.  The specific issue is in frame-1.1.0:

```
      obsgeoloc:
        description: |
          3-vector giving the position of the observer relative to the
          center-of-mass of the Earth, oriented the same as
          BCRS/ICRS. Defaults to `[0, 0, 0]`, meaning "true" GCRS.  Used
          when `reference_frame` is `GCRS` or `precessed_geocentric`.
        type: array
        items:
          $ref: ../unit/quantity-1.1.0
        minItems: 3
        maxItems: 3
        default:
          - { value: [0], unit: m }
          - { value: [0], unit: m }
          - { value: [0], unit: m }
```

The elements of the default array aren't valid because they aren't tagged, instead they're generic objects that merely resemble quantity-1.1.0 in structure.  The default should look something like this:

```
        default:
          - !<tag:stsci.edu:asdf/unit/quantity-1.1.0>
            value: !<tag:stsci.edu:asdf/core/ndarray-1.0.0> [0]
            unit: !<tag:stsci.edu:asdf/unit/unit-1.0.0> m
          - !<tag:stsci.edu:asdf/unit/quantity-1.1.0>
            value: !<tag:stsci.edu:asdf/core/ndarray-1.0.0> [0]
            unit: !<tag:stsci.edu:asdf/unit/unit-1.0.0> m
          - !<tag:stsci.edu:asdf/unit/quantity-1.1.0>
            value: !<tag:stsci.edu:asdf/core/ndarray-1.0.0> [0]
            unit: !<tag:stsci.edu:asdf/unit/unit-1.0.0> m
```

I've excluded anything that uses frame-1.1.0 from the schema tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,8 +39,11 @@
 - Fix bug that corrupted ndarrays when the underlying block
   array was converted to C order on write. [#827]
 
-# Fix bug that produced unreadable ASDF files when an
+- Fix bug that produced unreadable ASDF files when an
   ndarray in the tree was both offset and broadcasted. [#827]
+
+- Fix bug preventing validation of default values in
+  ``schema.check_schema``. [#785]
 
 2.6.1 (unreleased)
 ------------------
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -20,6 +20,7 @@ from . import reference
 from . import treeutil
 from . import util
 from . import extension
+from . import yamlutil
 from .exceptions import AsdfDeprecationWarning, AsdfWarning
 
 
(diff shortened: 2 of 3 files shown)


perrygreenfield:
What is the behavior after the fix when it encounters the deprecated schema? Does it now fail to load the file?

eslavich:
> What is the behavior after the fix when it encounters the deprecated schema? Does it now fail to load the file?

This is only a fix of the check_schema function, which isn't used during normal operation, only in the pytest plugin that validates the schemas themselves.  Behavior on opening files that use the wcs schemas shouldn't change.

The reason why these files open successfully is due to an another bug in the validator: it loads schemas without resolving `$ref` fields to referenced schema content.  So, the referenced schema's validations aren't actually checked, and defaults from the referenced schema aren't added.  I would like to fix this too but doing so will cause trouble for files created with the old wcs schemas.

eslavich:
We might as well include this in the 2.7 release, I rebased it and confirmed that it still works as intended (catches errors in the deprecated schemas that have invalid defaults).

PR #786, opened 2020-04-28
Fix diffing of simplified ndarray

eslavich:
This PR fixes a bug in the diff tool, which wasn't able to handle a serialized ndarray in "simplified" form, meaning an array that is serialized like this:

```
array: !core/ndarray-1.0.0 [0, 1, 2]
```
Instead of the usual:
```
array: !core/ndarray-1.0.0
  data: [0, 1, 2]
  datatype: int64
  shape: [3]
```

Both are permitted by the ndarray-1.0.0 schema.  I noticed this when working on fixing issues with schema defaults -- the frame-1.1.0 schema's defaults are represented in the simplified form, and once they start appearing in files, tests in tests_diff.py start failing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+2.6.1 (unreleased)
+------------------
+
+- Fix bug preventing diff of files containing ndarray-1.0.0
+  objects in simplified form. [#786]
+
 2.6.0 (2020-04-22)
 ------------------
 
--- a/asdf/commands/diff.py
+++ b/asdf/commands/diff.py
@@ -190,12 +190,17 @@ def print_dict_diff(diff_ctx, tree, node_list, keys, other):
 
 def compare_ndarrays(diff_ctx, array0, array1, keys):
     """Compares two ndarray objects"""
+    if isinstance(array0, list):
+        array0 = {"data": array0}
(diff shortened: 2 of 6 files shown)


perrygreenfield:
I'm a bit confused about the code and the expectations of it (I've never looked this part over). It looks like that this is not checking whether two kinds of alternate definitions of arrays will test as equal, but whether it will compare inline arrays as equal. So are you saying that there were no tests for inline arrays previously (there obviously were not), and when you added the test, they failed, and the code was changed to deal with these tests? Should there be tests to test equality of inline and block arrays too?

eslavich:
> So are you saying that there were no tests for inline arrays previously (there obviously were not), and when you added the test, they failed, and the code was changed to deal with these tests?

Yup, that's exactly it.

> Should there be tests to test equality of inline and block arrays too?

The diff tool is unusual in that it operates on a tagged tree and not a fully deserialized tree with real ndarray values.  Since that's the case, it's sensitive to how objects were actually represented in the YAML.  This weird simplified form of the inline array is unique in that the YAML representation of the ndarray is itself an array instead of an object with an array nested in a data data.  I don't think we need to test inline and block arrays explicitly because they all have the more common YAML representation.

Issue #570, opened 2018-10-22
Try to combine filling of default values with the main validation pass


drdavella:
This could potentially be a major performance improvement since currently there are two separate validation passes.

stscijgbot:
This ticket is now being tracked at [AL-121](https://jira.stsci.edu/browse/AL-121)

PR #787, opened 2020-04-28
Improve read performance by skipping rebuild of tree before validation

eslavich:
While I was profiling validation to assess the value of #570, I noticed this unrelated opportunity for a significant speedup (~ 13% less time for complex files).  When the tree is first read, it's already in tagged form.  Then we modify it twice: once to wrap `$ref` fields in `reference.Reference` objects, and again to fill in defaults from the schema.  From the validator's point of view, neither of these steps wrecks the tagged form of the tree, since references are ignored during validation, and the schema defaults are already tagged objects.

Given that, we can proceed to validation without first calling `yamlutil.custom_tree_to_tagged_tree` on the tree, which is an expensive operation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+2.7.0 (unreleased)
+------------------
+
+- Improve read performance by skipping unnecessary rebuild
+  of tagged tree. [#787]
+
 2.6.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


perrygreenfield:
Is this one of those flaky build errors?

eslavich:
> Is this one of those flaky build errors?

Seems that way... at some point I lost my ability to restart jobs on Travis, I'll try to figure out why I can't do that anymore.

PR #788, opened 2020-04-29
Fix deepcopy of TaggedList

eslavich:
The current behavior:

```
In [1]: from asdf.tagged import TaggedList; from copy import deepcopy

In [2]: foo = TaggedList([1, 2, 3])

In [3]: deepcopy(foo)
Out[3]: [1, 2, 3, 1, 2, 3]

In [4]: deepcopy(deepcopy(foo))
Out[4]: [1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3]
```

This has something to do with the `TaggedList` class inheriting from both `list` and `UserList`, but we can't drop `list`, because code elsewhere is probably relying upon this:

```
In [5]: isinstance(foo, list)
Out[5]: True
```

Which isn't the case when we inherit from `UserList` only -- users are expected to check `collections.abc.Sequence` instead.  I'm in favor of dropping list eventually, but that seems like a 3.0 change.

This PR overrides `__deepcopy__` to correct the issue with `UserList`.  It does the same on `UserDict`, even though there wasn't a problem there, since the custom `__deepcopy__` ended up being ~ 60% faster than whatever was happening without it.  Overriding `__copy__` offers no improvement, so I just left that alone.

Thanks to @jdavies-st for suggesting both `__deepcopy__` and profiling `UserDict`!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 - Fix bug preventing diff of files containing ndarray-1.0.0
   objects in simplified form. [#786]
 
+- Fix bug causing duplicate elements to appear when calling
+  ``copy.deepcopy`` on a ``TaggedList``. [#788]
+
 2.6.0 (2020-04-22)
 ------------------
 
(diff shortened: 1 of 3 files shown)


eslavich:
Turns out we also need a `__copy__` override -- in Python < 3.8, copying the objects will result in `MemoryError`.

PR #792, opened 2020-05-06
Add flag to skip validation on read

### LINKED ISSUE #565: Add option to skip validation
Reasons for doing this will probably be related to performance. I'm inclined to say that skipping validation should result in a loud warning so that users do not get into the habit of doing this all the time.

eslavich:
This PR adds a flag to `asdf.open` and `fits_embed.AsdfInFits.open` that disables validation on read, for situations when the file is already known to be valid and the user wishes to avoid the performance penalty of an unnecessary validation pass.

Resolves #565

### DIFF
--- a/asdf/tests/test_api.py
+++ b/asdf/tests/test_api.py
@@ -85,8 +85,6 @@ def test_open_readonly(tmpdir):
 
 
 def test_open_validate_on_read(tmpdir):
-    tmpfile = str(tmpdir.join('invalid.asdf'))
-
     content = """
 invalid_software: !core/software-1.0.0
   name: Minesweeper
--- a/asdf/tests/test_api.py
+++ b/asdf/tests/test_api.py
@@ -88,7 +88,7 @@ def test_open_validate_on_read(tmpdir):
     tmpfile = str(tmpdir.join('invalid.asdf'))
 
     content = """
-invalid_software: !<http://stsci.edu/schemas/asdf/core/software-1.0.0>
+invalid_software: !core/software-1.0.0
   name: Minesweeper
   version: 3
 """
--- a/asdf/tests/test_api.py
+++ b/asdf/tests/test_api.py
@@ -10,6 +10,8 @@ from astropy.modeling import models
 
 import pytest
 
+from jsonschema.exceptions import ValidationError
+
 import asdf
 from asdf import treeutil
 from asdf import extension
@@ -17,7 +19,7 @@ from asdf import resolver
 from asdf import schema
 from asdf import versioning
 from asdf.exceptions import AsdfDeprecationWarning
-from .helpers import assert_tree_match, assert_roundtrip_tree, display_warnings
+from .helpers import assert_tree_match, assert_roundtrip_tree, display_warnings, yaml_to_asdf
 
 
 def test_get_data_from_closed_file(tmpdir):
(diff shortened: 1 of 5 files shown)


asdf/tests/test_api.py
```diff
@@ -82,6 +84,27 @@ def test_open_readonly(tmpdir):
             pass
 
 
+def test_open_validate_on_read(tmpdir):
+    tmpfile = str(tmpdir.join('invalid.asdf'))
```
perrygreenfield, line 88: Is there ever anything written to this file path? I don't see it in the code below, but maybe there is some test magic going on that isn't apparent.

jdavies-st:
It looks like this `validate_on_read=True` flag doesn't change the anything by default, but can be switched to `False` and it will then not go through any schema validation?  Is the use of the schema to deserialize the ASDF content into a python object separate from validation of the ASDF against the various schemas?

asdf/tests/test_api.py
eslavich: Uhhhhh... _waves hand vaguely_ this is not the tempfile you're looking for

eslavich:
> It looks like this validate_on_read=True flag doesn't change the anything by default, but can be switched to False and it will then not go through any schema validation?

Yup!  Well, it'll skip the main validation pass, anyway.  The validator will still be used to fill in default values, but that pass doesn't check data types or anything.

> Is the use of the schema to deserialize the ASDF content into a python object separate from validation of the ASDF against the various schemas?

AFAIK, if validation and filling defaults are both disabled, the schema isn't used at all.

jdavies-st:
Ah, I see.  So then only the `CustomType` subclass is used to turn the ASDF YAML into the custom python object?

eslavich:
Right, and that matches to the YAML by tag, which is already known in both the file and the subclass, so the schema isn't consulted.

perrygreenfield:
But tag code is still executed?

jdavies-st:
Wow, I didn't realize the schemas were so superfluous.  =)

eslavich:
> But tag code is still executed?

I made an interesting test: I deleted a bunch of core schemas and opened a file with `validate_on_read=False` and `do_not_fill_defaults=True`, and it opens and deserializes correctly without so much as a warning.

perrygreenfield:
So the only subtlety is regarding defaults. In one case (using schemas), one has the attribute with a default value, but if `validate_on_read=False`, then the attribute is missing unless the tag code also handles defaults implicitly.

perrygreenfield:
The more I see the implications, the less I like schemas handling defaults. But I may be having a brain short circuit...

jdavies-st:
I agree that handling defaults is bad, as it makes data values depend on the vagaries of schemas.  Just to save a few bytes.

And there's the whole issue of not being able to serialize `None` because that's the value that `asdf` uses to know whether or not to fill defaults.

eslavich:
> So the only subtlety is regarding defaults. In one case (using schemas), one has the attribute with a default value, but if validate_on_read=False, then the attribute is missing unless the tag code also handles defaults implicitly.

Actually, the defaults are still filled in.  That happens on a separate pass through the tree that this PR doesn't impact.  My next PR does combine those two passes when both validation and defaults are enabled, but even there the defaults are still filled in with validation disabled.

That said, I would love it if we could drop the defaults feature in 3.0.  The more time I spend trying to figure out a sane procedure for choosing a default from a complex schema the more I suspect there isn't any such procedure.  These schemas just aren't designed to provide a definitive answer to the question "what is the value of an annotation for this node?".  IMO the default field should exist only as advice to people writing deserializers, and not be acted upon by asdf code.  The type class seems like a more natural place to set defaults.

PR #793, opened 2020-05-06
Validate and fill/remove defaults in a single pass

### LINKED ISSUE #570: Try to combine filling of default values with the main validation pass

eslavich:
This PR combines filling defaults and validation on read into a single pass, when both are enabled, and the same treatment for removing defaults and validation on write.  This improves performance since until now we've been walking the tree and schema twice.

Branched off of #792, so probably worth waiting for that one to merge before reviewing this.

Resolves #570

### DIFF
--- a/asdf/commands/tests/data/frames0.asdf
+++ b/asdf/commands/tests/data/frames0.asdf
@@ -6,35 +6,29 @@
 asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
   name: asdf, version: 1.2.2.dev858}
 frames:
-- !wcs/celestial_frame-1.1.0
-  axes_names: [lon, lat]
+- axes_names: [lon, lat]
   name: CelestialFrame
   reference_frame: {type: ICRS}
   unit: [!unit/unit-1.0.0 deg, !unit/unit-1.0.0 deg]
-- !wcs/celestial_frame-1.1.0
-  axes_names: [lon, lat]
+- axes_names: [lon, lat]
   name: CelestialFrame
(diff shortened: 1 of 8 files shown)


asdf/commands/tests/data/frames0.asdf
```diff
@@ -6,35 +6,29 @@
 asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
   name: asdf, version: 1.2.2.dev858}
 frames:
-- !wcs/celestial_frame-1.1.0
```
eslavich, line 9: I removed the frame tags because defaults in the frame-1.1.0 schema fail validation, and this change allows us to fill in defaults in nested objects that we were missing before.

asdf/schema.py
```diff
@@ -168,45 +168,52 @@ def validate_type(validator, types, instance, schema):
...
+            if "default" in subschema:
+                instance.setdefault(property, subschema["default"])
+
+    yield from _PROPERTIES_VALIDATOR(validator, properties, instance, schema)
 
 
+# Used when filling defaults in a separate validation pass:
 FILL_DEFAULTS = util.HashableDict()
-for key in ('allOf', 'anyOf', 'oneOf', 'items'):
+for key in ('allOf', 'anyOf', 'oneOf', 'items', '$ref'):
```
eslavich, line 185: Adding `$ref` here is a bug fix on the existing defaults behavior, which until now has ignored defaults from referenced schemas.  I noticed the issue when comparing to the result of the combined defaults/validation pass, which gave a different answer that turned out to be correct.

PR #794, opened 2020-05-06
Prevent validation of empty tree when AsdfFile is initialized

eslavich:
I noticed this while playing with #792 -- some validation was still occurring despite the new flag.  Turns out when `AsdfFile` is initialized, we're setting the `tree` property to an empty dict, which results in a validation pass over that object.  This PR bypasses that validation by setting `_tree` directly.

This PR also fixes a minor bug when opening a file containing only blocks, and fixes some tests that relied upon a side-effect of the unnecessary validation pass.


asdf/asdf.py
```diff
@@ -140,7 +140,9 @@ def __init__(self, tree=None, uri=None, extensions=None, version=None,
             readonly=_readonly)
         self._uri = None
         if tree is None:
-            self.tree = {}
+            # Bypassing the tree property here, to avoid validating
+            # an empty tree.
+            self._tree = AsdfObject()
```
jdavies-st, line 145: Might it be better to just pass `tree` to the setter for tree, and have the setter handle None?  This whole block seems like it should be handled by the setter, not just on `__init__`.

eslavich: When you say whole block, is that the whole if/elif/else?  I don't think we want the setter to handle copy-constructing from another AsdfFile, since that requires messing around with blocks and copying other AsdfFile internal variables besides the tree.

I have a slight preference for keeping the commented line as-is, since users can already pass `{}` to set an empty tree, and it seems tidy for there to be only one way to do it.

jdavies-st: That makes sense.  👍

PR #797, opened 2020-05-14
Add F and W flags to automated flake8 checks in setup.cfg

jdavies-st:
This PR adds all `F` and most `W` codes to the `flake8` automated checks.  This finds several bugs (and an unknown direct dependency for the doc builds).

All the rest of the changes are to fix the bugs and improve code style.

Replaces the PEP8 updates in #692


asdf/asdftypes.py
```diff
@@ -9,6 +9,9 @@
 from .types import (AsdfType, CustomType, format_tag, ExtensionTypeMeta,
                     _all_asdftypes)
 
+__all__ = ["join_tag_version", "split_tag_version", "AsdfType", "CustomType",
+    "format_tag", "ExtensionTypeMeta", "_all_asdftypes"]
```
eslavich, line 13: Should we leave `_all_asdftypes` out of this list?  With the leading underscore it doesn't seem like it's already part of the public API.

asdf/conftest.py
```diff
@@ -1,18 +1,13 @@
...
+
+# This contains pytest fixtures used in asdf tests.
+# by importing them here in conftest.py they are discoverable by pytest
 # no matter how it is invoked within the source tree.
 
-import os
 
 import pytest
 
-from .extern.RangeHTTPServer import RangeHTTPRequestHandler
```
eslavich, line 11: Is this just a test dependency?  Doesn't seem worth vendorizing just for that.

setup.cfg
```diff
@@ -92,4 +93,5 @@ addopts = --doctest-rst
 
 [flake8]
 exclude = extern
-select = E101 W191 W291 W292 W293 W391 E11 E502 E722 E901 E902
+select = F, W, E101 W191 W291 W292 W293 W391 E11 E502 E722 E901 E902
```
eslavich, line 96: Can we can remove the W### from this list?  Seems like they should already be selected by `F`

asdf/conftest.py
jdavies-st: Yes, a dependency circuitously for `test_generic_io.py`.  I suspect something like `requests-mock` would be more appropriate.

setup.cfg
jdavies-st: Done

jdavies-st:
Yeah, I'm not sure what is going on with the docs.  I can reproduce locally, but I'm not entirely clear one what is going on in the docs as it is.  It seems that it uses `automodapi`, but the sphinx conf doesn't actually use that extension.  A head scratcher.  The error message is not very useful either.  Hmm.

jdavies-st:
So the doc build failure was a result of 2 odd things.

First, `docs/conf.py` imports everything from `sphinx_astropy.conf.v1`, not just the functions.  Very odd.

And 2nd, in the top level `asdf/__init__.py`, we make available `ValidationError` from `jsonschema` as if it is coming from `asdf`. But it can't be part of `__all__` or else it mucks up the `automodapi` stuff, as it is looking for the asdf code for `ValidationError`, but of course can't find it.

Is there a reason we don't just use `jsonschema.ValidationError` instead of `asdf.ValidationError`?

asdf/conftest.py
eslavich: I'll make an issue for this

jdavies-st:
I think I've addressed the issues we were seeing.

PR #803, opened 2020-05-20
Enable tests on Windows

eslavich:
Turns out with a few small adjustments we don't need to skip these tests on Windows.  I suspect the Travis environment may have disabled the firewall since last time someone tested this, and the path manipulation stuff is straightforward enough if we use a cross-platform utility instead of constructing paths ourselves.

### DIFF
--- a/asdf/tests/helpers.py
+++ b/asdf/tests/helpers.py
@@ -233,7 +233,7 @@ def _assert_roundtrip_tree(tree, tmpdir, *, asdf_check_func=None,
             asdf_check_func(ff)
 
     # Now try everything on an HTTP range server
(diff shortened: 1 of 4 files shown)


jdavies-st:
Btw, does the windows test job report coverage?  If there's no Windows-only forking in the code, it probably doesn't need to.  But if there is, it would be good to have it reporting coverage.

eslavich:
> Btw, does the windows test job report coverage? If there's no Windows-only forking in the code, it probably doesn't need to. But if there is, it would be good to have it reporting coverage.

I don't think it does, coverage is based on a linux job only.  We have tests for both Windows and OS X in the code, so this is probably a good idea.

PR #804, opened 2020-05-20
Use AsdfWarning instead of UserWarning

eslavich:
Many of our warnings are `UserWarning` instead of `AsdfWarning` of some kind, which makes it difficult for users to squelch warnings from the whole package all at once.  This converts all previously untyped warnings to `AsdfWarning` and cleans up warning checks in the tests, which weren't making effective use of the `pytest.warns` match parameter.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@
 - Prevent validation of empty tree when ``AsdfFile`` is
   initialized. [#794]
 
+- All warnings now subclass ``asdf.exceptions.AsdfWarning``. [#804]
+
 2.6.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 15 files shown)


eslavich:
> So this looks good. One concern I have is that previously all these warnings were throwing UserWarning, so if anyone has code that depends on that warning type being thrown, say they want to filter them, it now is broken.

Hmm, didn't think about that.  Given that it's only a warning, is the worst case scenario is a user suddenly sees a bunch of noise in the output that they thought they'd already silenced?  If that's the case, then I think it's okay, though we might want to drop a note to the mailing list as a "warning warning".

jdavies-st:
Yeah, I think that's reasonable.  It would be more serious if it was a change in exception type.

PR #805, opened 2020-05-20
Fix collection issues with doctestplus 0.7.0

eslavich:
The new version stops automatically ignoring import errors on collection.  This causes a failure at collection time due to missing dependencies for docs/sphinxext.  We don't have any docstrings or tests in docs/sphinxext anyway, so I'm just going to ask pytest to ignore that directory.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -76,7 +76,7 @@ show-response = 1
 [tool:pytest]
 testpaths = asdf docs asdf-standard/schemas
 minversion = 3.1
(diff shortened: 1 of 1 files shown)


eslavich:
@jdavies-st I'm going to merge this right away so that I can continue with other PRs, but let me know if you disagree with my choice of fix.

jdavies-st:
Looks good.  Maybe we should clean out that directory if there's stuff in there that doesn't import?

PR #808, opened 2020-06-11
Minimum support for astropy-asdf

eslavich:
This PR introduces a new API for extending asdf.  The `AsdfConverter` class is a replacement for the existing `ExtensionType`.  `AsdfConverterProvider` is a plugin class in the vein of the existing `AsdfExtension`, except that it does not provide schemas.

I've also included a global configuration mechanism.  Not sure yet if that's a good idea.

Here's an example of how to set up a converter for a new type:

```python
import asdf

class Foo:
    def __init__(self, value):
        self.value = value

class FooConverter(asdf.AsdfConverter):
    tags = {"http://asdf-format.org/schemas/test/foo-1.0.0"}
    types = {Foo}
    
    def to_yaml_tree(self, obj):
        return {"value": obj.value}
   
    def from_yaml_tree(self, node):
        return Foo(node["value"])

class FooConverterProvider(asdf.AsdfConverterProvider):
    converter_classes = [FooConverter]

asdf.configure(converter_providers=[FooConverterProvider()])
```
With that in place, we can now write and read `Foo` objects from ASDF files:
```python
In [2]: af = asdf.AsdfFile()

In [3]: af["foo"] = Foo("bar")

In [4]: af.write_to("test.asdf")

In [5]: asdf.open("test.asdf")["foo"].value
Out[5]: 'bar'
```
The new `asdf.configure` method sets options globally:
```python
In [13]: print(len(asdf.config.get_config().converter_providers))
    ...: asdf.configure(converter_providers=[FooConverterProvider()])
    ...: print(len(asdf.config.get_config().converter_providers))
0
1
```
But if used as a context manager, it reverts to the previous options on exit:
```python
In [16]: print(len(asdf.config.get_config().converter_providers))
    ...: with asdf.configure(converter_providers=[FooConverterProvider()]):
    ...:     print(len(asdf.config.get_config().converter_providers))
    ...: print(len(asdf.config.get_config().converter_providers))
0
1
0
```
Anyway this PR is open on the epic branch, so there will be plenty of opportunity to iterate on the details later.  No tests yet because the code is so likely to change.


asdf/config.py
```diff
@@ -0,0 +1,81 @@
...
+        converters = None
+        converter_index = None
+
+    _config = AsdfConfig(
+        converter_providers=converter_providers,
+        converters=converters,
+        converter_index=converter_index,
+    )
+
+    return _ConfigContext(original_config)
```
perrygreenfield, line 81: If we wanted to set additional ASDF preferences, is this where they would go? (e.g., array inline size thresholds, etc.)

asdf/config.py
```diff
@@ -0,0 +1,81 @@
...
+
+_config = AsdfConfig()
+
+
+def get_config():
+    return _config
+
+
+class _ConfigContext:
+    def __init__(self, original_config):
```
perrygreenfield, line 50: I have a preference for something like current_config or previous_config. Original makes me think it is the first.

asdf/config.py
```diff
@@ -0,0 +1,81 @@
+from .util import NotSet
+from .providers import get_registered_converter_providers, create_converters
+from .converter import ConverterIndex
+
+
+__all__ = ["get_config", "configure"]
+
+
+class AsdfConfig:
+    def __init__(
```
perrygreenfield, line 10: Docstrings with examples needed (example for the last function in file).

asdf/converter.py
```diff
@@ -0,0 +1,192 @@
...
+            attribute, or a generator that yields such an instance.
+        """
+
+        raise NotImplementedError("AsdfConverter subclasses must implement from_yaml_tree")
+
+
+# TODO(eslavich): Implement base class for converters that manipulate blocks
+
+
+class ConverterIndex:
```
perrygreenfield, line 154: Is Index the best term for this, particularly in the Python context. First thing I think of is that it is going to give me a position in a list or array, whereas it is more like a book index. ConverterDirectory, ConverterWhitePages (sorry) or something more like that? It really provides two different modes of lookups

asdf/yamlutil.py
```diff
@@ -221,15 +223,42 @@ def custom_tree_to_tagged_tree(tree, ctx):
...
+        if isinstance(node, tagged.Tagged):
+            return node
+        else:
+            return tagged.tag_object(converter.tag, node, ctx=ctx)
+
+    def _walker(node):
+        converter = get_config().converter_index.from_type(type(node))
+        if converter is not None:
+            return _to_yaml_tree_tagged(converter, node)
+
```
perrygreenfield, line 252: Does this coexist with the older TransformType? For example, won't it have to deal with tags not yet using the new Converter approach? This doesn't look like it handles that but I probably am missing where that is happening (e.g. those type classes presented as converters?)

asdf/config.py
eslavich: Yup!  Assuming everyone agrees that this is a reasonable idea.

asdf/config.py
eslavich: I like "previous", I'll change it.

asdf/converter.py
eslavich: ConverterCollection?  ConverterManager?  I took the Index name from the previous asdf class (TypeIndex) but I agree it could be better.

asdf/config.py
perrygreenfield: Should also document that it isn't thread safe (yet, anyway)

asdf/yamlutil.py
eslavich: It does fall back to the Type classes when Converter support doesn't exist.  If object returned from `converter_index.from_type` is None, it proceeds to the original (unaltered) code.

perrygreenfield: OK, I see how that is done now. Thanks

asdf/config.py
eslavich: (the config module changed enough that this is no longer relevant)

asdf/config.py
eslavich: The latest commit makes this thread safe!

asdf/config.py
eslavich: Added docstrings

asdf/config.py
```diff
@@ -1,81 +1,237 @@
...
+def config_context(
+    converter_providers=NotSet,
+    validate_on_read=NotSet,
+):
+    """
+    Create a context manager that temporarily overrides asdf
+    configuration.  Overrides are only applied in the current
+    thread.  Nested calls to this method are applied in order,
+    with the innermost context overriding the others. Changes
+    affect currently open `AsdfFile` instances.
```
perrygreenfield, line 215: Hmmm. Is this accurate? I'm wondering if that the configuration may be bound to an open file and persist until it is closed, though that does complicate things. For example, inside a function, I open a file that has a configuration parameter for the inline array threshold, and I set a configuration for a certain threshold before I open it, exit the configuration after opening the file, and return the file. I may have wanted to have that threshold apply so long as the file is open even though the context that the configuration was created has since exited. So I wonder in the open a reference to the configuration can be associated with the AsdfFile instance. Is this crazy thinking? I raise this since I think I was naive in making the analogy to numpy (or numarray where it was conceived) for handling floating point errors. The difference there is the errors arise out of operations performed on arrays, and those operations are bound to the code that is performing them. In this case, I tend to think the behaviors are more bound to the object than the code that is manipulating them. Would we want to change the configuration that applies to a given file at different places in the code?

I still think the current context for configuration still is useful, but I wonder if it should only be associated with where it is executed and not only with the objects created. Sorry for the monkey wrench. They are greasy and slip out my hand often.

perrygreenfield: The behavior could be that the configuration currently set means that is what will be bound to a file if opened or created while in effect, but it can persist with the object.

eslavich: We can do it either way without much trouble I think -- the AsdfFile should be able to grab whatever the current config object is and keep a reference to it.  The individual configs are immutable so that is all it needs to retain the options that were set when the AsdfFile was created.

I've been thinking about this in terms of familiar desktop applications like text editors.  Some settings are editor-level, like disabling spell check, automatically trimming whitespace, etc.  Others are file-level like the file mode (TextEdit has "plain text" and "rich text"), margin width, etc.  Some are both -- you can configure the font size on the file level, but you can also configure the default font size on the editor level.

It seems like we also have some of each category.  We might break ours down like this:

"editor level"
- flag to warn on unrecognized tag
- flag to validate on read
- list of enabled plugins (extensions and the new providers)
- default for memory mapping enabled/disabled
- default ASDF Standard version for new files
- default inline array threshold for new files

"file level"
- ASDF Standard version (overrides default set at editor level)
- memory mapping enabled/disabled (overrides default set at editor level)
- inline array threshold (overrides default set at editor level)

What do you think about breaking down the settings in this way, and keeping the "editor level" settings in asdf.config?  If users changed them in the middle of a session, they'd affect existing AsdfFile instances.  The "file level" settings, by contrast, would be sticky for existing AsdfFile instances.

eslavich: BTW, inline array threshold seems borderline to me -- it's a sort of stylistic choice we make when writing the file, but not really a property of the file itself.  We don't write the `auto_inline` value into the file so other users won't know what threshold was used.

Come to think of it, that's a decent rule of thumb for deciding whether a setting should be "file level" or "editor level": if it gets saved as a property of the actual ASDF file, then it's probably file level, otherwise not.  Though memory mapping enabled/disabled is an obvious exception.

perrygreenfield: The two contexts (editor/file) seem reasonable to me. 

As for inline threshold, that's a bit more murky. I tend to think it is more associated with the file (e.g., someone may want everything inline, or everything smaller than x inline, and they don't want to deal with every level in the code that may add arrays). On the other hand, if they control when the file is written to disk, they could force it there, but will they always control it at that point? I can see variations of where these details are or aren't available to the user (the user may be a developer using someone else's library that adds to the tree, and possibly writes it out. Perhaps we need to think a bit more.

perrygreenfield: As example use cases,

1) someone want to run Roman pipeline (aqueduct?) how would we best allow the user to indicate a inline threshold? In this case they specify an input ASDF file, which is not the output. In this case it simply seems like it should be a option to the pipeline, and thus a global parameter rather than file specific, if running from the command line. If running from a python session they could make it their configuration, and avoid having to specify it every time they execute a pipeline task.

2) They are building their own ASDF product in their code, so they create the file object and manage the tree construction. Either approach is fine.

3) The create or read in a file object then pass it to code that will modify or add to it and writes it out. This seems a bit unusual case but possible. In this event can they be sure that the underlying code won't change the configuration. No, but they don't have any control over that, but they can change the global configuration and hope the code won't override the threshold aspect. If the configuration was attached to the file, I guess the presumption would be that library code should not override it unless there is a strong reason to; so that would argue for allowing it to be attached to the file.

3) another solution may be to provide utilities to copy files to the flavor the user wants and avoid the configuration mess. Inefficient in requiring copies sometime, but straightforward.

I'm still giving this a little more thought.

perrygreenfield: I guess my inclination is to leave it out of file level until we see a strong use case to change that. I don't see one yet.

asdf/converter.py
perrygreenfield: ConverterCollection seems appropriate

perrygreenfield: Should we poll to see what name everyone likes? Otherwise, this whole PR looks like a good basis to start from (after learning about secret RLock context manager behavior tricks and the like).

jdavies-st: `ConverterCollection` is good.

eslavich: `ConverterCollection` suits me fine too, I'll make that update

asdf/config.py
```diff
@@ -0,0 +1,237 @@
...
+    if len(_local.config_stack) == 0:
+        return _global_config
+    else:
+        return _local.config_stack[-1]
+
+
+def configure(
+    converter_providers=NotSet,
+    validate_on_read=NotSet,
+):
```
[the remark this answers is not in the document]
eslavich: `None` does in fact have a distinct meaning here -- it means "load the providers registered with entry points".  The purpose of `NotSet` is to allow users to configure options piecemeal.  If the defaults were `None`, then this call:

```python
asdf.configure(validate_on_read=True)
```

would cause the current value of converter_providers to be wiped out and reloaded from entry points, which I doubt is what the user wants.

An alternative way to implement this same interface is to use `**kwargs`:
 
```python
def configure(**kwargs):
  if "converter_providers" in kwargs:
    converter_providers = kwargs["converter_providers"]
  else:
    converter_providers = <the previous value>
```

but I find this significantly less usable because IDEs won't be able to detect and list possible keywords, the user will have to dig into the docstring.

asdf/asdf.py
```diff
@@ -616,14 +617,23 @@ def _open_asdf(cls, self, fd, uri=None, mode='r',
                    _force_raw_types=False,
                    strict_extension_check=False,
                    ignore_missing_extensions=False,
-                   validate_on_read=True):
+                   validate_on_read=NotSet):
```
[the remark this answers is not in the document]
eslavich: Since `None` is falsy, I thought it best to make the new default something unambiguous.

asdf/config.py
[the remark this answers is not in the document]
eslavich: This may not be clear from the documentation, but the intended behavior of this method is to "make a selective update to the current configuration", not "set all configuration options and use defaults for anything not explicitly set".

asdf/asdf.py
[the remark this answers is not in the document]
jdavies-st: I see.  You're using it to detect if the optional argument has been passed to the function.

Since default values get filled in python when the function is defined, maybe the [following design pattern](https://stackoverflow.com/questions/14749328/how-to-check-whether-optional-function-parameter-is-set) works?

```
def configure(converter_providers=None, validate_on_read=None): 
    if converter_providers is configure.__defaults__[0]: 
        print("cp default") 
    else: 
        print("cp passed in call") 
    if validate_on_read is configure.__defaults__[1]: 
        print("vor default") 
    else: 
        print("vor passed in call") 
```

This gives a slightly more readable API for the user (at the expense of weird code in the function itself to handle it).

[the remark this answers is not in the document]
jdavies-st: In python, generally optional arguments that don't do anything unless something is passed to them are given `None`.  I think this is much more common than some custom type of which I don't (as a user) know what it means.  In the context of function args, `None` is not falsy in python.  It merely means, "does not apply".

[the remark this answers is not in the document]
eslavich: I just mean, we use the value of the flag like this:

```python
if validate_on_read:
  ...
```
so if someone had been passing `None` for some reason, it will have the opposite effect that it did in the previous version of asdf if we start treating `None` as "use the default value of `True`".  Too fine a point to worry about?

asdf/config.py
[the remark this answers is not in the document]
eslavich: I don't think that behaves the way we'd need, for example this:
```python
In [24]: configure(converter_providers=None)
cp default
vor default
```
gives the same result as this:
```python
In [25]: configure()
cp default
vor default
```
What I'm looking for is the following behavior:
```python
In [26]: configure()
cp use previous value
vor use previous value

In [26]: configure(converter_providers=[CustomConverterProvider()])
cp set to [CustomConverterProvider instance]
vor use previous value

In [26]: configure(converter_providers=None)
cp reload providers from entry points
vor use previous value

In [26]: configure(validate_on_read=False)
cp use previous value
vor set to False
```

asdf/config.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+_global_config = AsdfConfig()
+_local = _ConfigLocal()
+
+
+def get_config():
+    """
+    Get the current config, which may have been altered by
+    one or more containing `AsdfConfigContext`.
```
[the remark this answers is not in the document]
eslavich: I'm not sure, the scenario I'm trying to describe is this:

```
with asdf.config_context(...):
  with asdf.config_context(...):
    asdf.config.get_config()
```

so, the result of get_config() is altered by the two context managers that surround it.

asdf/asdf.py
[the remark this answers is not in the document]
jdavies-st: Nobody should be passing `validate_on_read=None`, as it was previously a boolean in the API.  If they were passing `None` for some reason, it was not explicitly supported.

asdf/config.py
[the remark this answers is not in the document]
eslavich: Oh that's good, thanks, I'll update it.  Theoretically user code could obtain an `AsdfConfigContext` and reuse it multiple times to apply and remove a particular set of config overrides, but I don't see that being a common use case.

asdf/asdf.py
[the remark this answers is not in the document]
eslavich: Thinking about this a little more, what we actually want to detect is the fact that someone is calling this method with the `validate_on_read` keyword.  When they do that, we emit a deprecation warning to save them from an error later when the keyword disappears.

So maybe the best option here is to use `**kwargs`, check for presence of "validate_on_read" in that dict, and use it but warn appropriately.  In this case we don't care that the method signature will be missing `validate_on_read`, that's actually a virtue because we're removing the keyword and it will discourage people from using it.

asdf/config.py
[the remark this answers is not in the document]
eslavich: Yes, that's right, at least in the case of `converter_providers`.  Does it makes sense to make that a consistent response to `None` across all options?  `validate_on_read` isn't currently coded up that way, but it could be.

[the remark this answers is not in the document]
perrygreenfield: Yes, it probably would be good to be consistent with that.

asdf/config.py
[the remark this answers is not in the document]
jdavies-st: The ambiguity is in your use of "containing" as an adjective.  Not sure of a way to make it clearer.  Your example makes it very clear.  =)

asdf/config.py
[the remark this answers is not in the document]
jdavies-st: Ok, this is clearer for me now.

asdf/config.py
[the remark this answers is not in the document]
eslavich: The latest commit changes the language to this:
```
    Get the current config, which may have been altered by
    one or more surrounding calls to `config_context`.
```
How does that strike you?

eslavich:
> Also, do you think using abc.ABC might be an alternate (perhaps simpler) way to implement AsdfConverter instead of defining a metaclass? The methods that need to be defined in the subclasses can be decorated @abstractmethod in AsdfConverter if it subclassed abc.ABC.

@jdavies-st I would still like to validate the content of the `tags` and `types` class attributes, don't I need a metaclass for that?

jdavies-st:
Ah, that's the reason.  I learned something new today!  🎉 

Awesome.

PR #810, opened 2020-06-12
Fix misinterpretation of byte order of blocks stored in FITS files

eslavich:
This PR fixes an issue with reading binary blocks stored in FITS files.

When astropy reads data that requires scaling, it performs said scaling and returns the result in native byte order.  Until now asdf has been assuming that all data from a FITS file is big-endian, since that's how the bytes are stored on disk, but due to the scaling operation that's not always the case.

The fix is to accept that the dtype reported by an array from a FITS file is correct.

CC @stscieisenhamer

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,9 @@
 - Fix bug with ``auto_inline`` option where inline blocks
   are not converted to internal when they exceed the threshold. [#802]
 
+- Fix misinterpretation of byte order of blocks stored
+  in FITS files. [#810]
+
 2.6.0 (2020-04-22)
 ------------------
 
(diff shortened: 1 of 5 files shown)


stscieisenhamer:
FYI: I tested on the case that prompted the issue report and it does fix the issue.

jdavies-st:
I suspect this was hidden because `jwst` is the only package that writes ASDF to FITS files, and there was a bug in `jwst.datamodels` layered on top of it that also munged the unsigned ints that @eslavich fixed a year ago.

https://github.com/spacetelescope/jwst/pull/3680

perrygreenfield:
Yeah but it has been doing that for a while. Did the jwst.datamodels bug actually counteract the ASDF bug? (so two wrongs do make a right?)

PR #819, opened 2020-07-10
Add new resource Mapping API

eslavich:
This PR introduces a new API for extending asdf with additional schemas.  The entry point accepts any `collections.abc.Mapping`, where the key is the schema URI `str` and the value is the content as `bytes`.

Included is a `Mapping` implementation that makes it easy to serve up schemas from a directory or package.  For asdf_transform_schemas, it'll look something like this:

```python
from pathlib import Path

from asdf.resource import DirectoryResourceMapping

def entry_point():
  schemas_root = Path(__file__).parent/"schemas". 
  return [
    DirectoryResourceMapping(schemas_root, "http://asdf-format.org/schemas/transform")
  ]
```

But any `Mapping` will work, so users can also add a schema from the console using a `dict`:
```python
import asdf

schema = """
type: object
properties:
  bar:
    type: integer
"""

asdf.get_config().add_resource_mapping({"http://somewhere.org/schemas/foo-1.0.0": schema})
```

### DIFF
--- a/asdf/_config.py
+++ b/asdf/_config.py
@@ -1,0 +1,185 @@
+"""
+Methods for getting and setting asdf global configuration
+options.
+"""
+import threading
+from contextlib import contextmanager
+import copy
+
+from . import entry_points
+from .resource import ResourceManager
+
+
+DEFAULT_VALIDATE_ON_READ = True
+
+
+class AsdfConfig:
+    """
+    Container for ASDF configuration options.  Users are not intended to
+    construct this object directly; instead, use the `asdf.get_config` and
+    `asdf.config_context` module methods.
+    """
+
+    def __init__(
+        self,
+        resource_mappings=None,
+        resource_manager=None,
+        validate_on_read=None,
+    ):
+        self._resource_mappings = resource_mappings
+        self._resource_manager = resource_manager
+
+        if validate_on_read is None:
+            self._validate_on_read = DEFAULT_VALIDATE_ON_READ
+        else:
+            self._validate_on_read = validate_on_read
+
+        self._lock = threading.RLock()
+
+    @property
+    def resource_mappings(self):
+        """
+        Get the list of enabled resource Mapping instances.  Unless
+        overridden by user configuration, this includes every Mapping
+        registered with an entry point.
+
+        Returns
+        -------
+        list of collections.abc.Mapping
+        """
+        if self._resource_mappings is None:
+            with self._lock:
+                if self._resource_mappings is None:
+                    self._resource_mappings = entry_points.get_resource_mappings()
+        return self._resource_mappings
+
+    def add_resource_mapping(self, mapping):
+        """
+        Register a new resource Mapping.
+
+        Parameters
+        ----------
+        mapping : collections.abc.Mapping
+            map of `str` resource URI to `bytes` content
+        """
+        with self._lock:
+            resource_mappings = self.resource_mappings.copy()
+            resource_mappings.append(mapping)
+            self._resource_mappings = resource_mappings
+            self._resource_manager = None
+
+    def remove_resource_mapping(self, mapping):
+        """
+        Remove a registered resource mapping.
+
+        Parameters
+        ----------
+        mapping : collections.abc.Mapping
+        """
+        with self._lock:
+            resource_mappings = [m for m in self.resource_mappings if m is not mapping]
+            self._resource_mappings = resource_mappings
+            self._resource_manager = None
+
+    def reset_resources(self):
+        """
+        Reset registered resource mappings to the default list
+        provided as entry points.
+        """
+        with self._lock:
+            self._resource_mappings = None
+            self._resource_manager = None
+
+    @property
+    def resource_manager(self):
+        """
+        Get the `ResourceManager` instance.  Includes resources from
+        registered resource Mappings and any Mappings added at runtime.
+
+        Returns
+        -------
+        asdf.resource.ResourceManager
+        """
+        if self._resource_manager is None:
+            with self._lock:
+                if self._resource_manager is None:
+                    self._resource_manager = ResourceManager(self.resource_mappings)
+        return self._resource_manager
+
+    @property
+    def validate_on_read(self):
+        """
+        Get configuration that controls schema validation of
+        ASDF files on read.
+
+        Returns
+        -------
+        bool
+        """
+        return self._validate_on_read
+
+    @validate_on_read.setter
+    def validate_on_read(self, value):
+        """
+        Set the configuration that controls schema validation of
+        ASDF files on read.  If `True`, newly opened files will
+        be validated.
+
+        Parameters
+        ----------
+        value : bool
+        """
+        self._validate_on_read = value
+
+    def __repr__(self):
+        return (
+            "AsdfConfig(\n"
+            "  resource_mappings=[...],\n"
+            "  validate_on_read={!r},\n"
+            ")"
+        ).format(self.validate_on_read)
+
+
+class _ConfigLocal(threading.local):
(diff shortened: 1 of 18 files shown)


asdf/_config.py
```diff
@@ -0,0 +1,185 @@
+"""
```
eslavich, line 1: This module changed a bit from the version that was merged into the `asdf-standard-2.0.0` branch.  I ended up wanting to make the config mutable after all, so that developers could add resources at runtime from the console.  This ended up simplifying the method signatures a good bit.

eslavich:
Note to self: the changelog bot seems to have gotten lost in the transition to the asdf-format org

asdf/tests/conftest.py
```diff
@@ -11,3 +16,10 @@ def small_tree():
 @pytest.fixture
 def large_tree():
     return create_large_tree()
+
+
+@pytest.fixture(autouse=True)
+def restore_default_config():
+    yield
+    _config._global_config = _config.AsdfConfig()
+    _config._local = _config._ConfigLocal()
```
jdavies-st, line 25: Is this to avoid having to use the `with` context manager in tests?

setup.cfg
```diff
@@ -27,6 +27,7 @@ install_requires =
     pyyaml>=3.10
     jsonschema>=3.0.2,<4
     numpy>=1.10
+    importlib_resources>=3;python_version<"3.9"
```
jdavies-st, line 30: What is `;python_version<"3.9"` doing here?  I believe `importlib_resources` is only needed for `python<=3.6`

asdf/entry_points.py
```diff
@@ -0,0 +1,27 @@
+from pkg_resources import iter_entry_points
+import warnings
+from collections.abc import Mapping
+
+from .exceptions import AsdfWarning
+
+
+RESOURCE_MAPPINGS_GROUP = "asdf.resource_mappings"
```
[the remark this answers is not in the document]
eslavich: Just a matter of style, I think.  I like to see constants like this defined LOUDLY at the top so they're easy to find and to quarantine the magic strings in one place.

asdf/resource.py
```diff
@@ -0,0 +1,180 @@
...
+import asdf
+
+
+__all__ = ["TraversableResourceMapping", "ResourceManager"]
+
+
+class TraversableResourceMapping(Mapping):
+    """
+    Resource mapping that reads resource content
+    from a root directory.
```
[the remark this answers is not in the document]
eslavich: I think that just snuck in because I called the parameter "root", maybe this should be "from a directory tree"?

asdf/tests/conftest.py
eslavich: Yea, and global state is such a pain when it breaks tests far away.  This is a hedge that will prevent that particular frustration.

setup.cfg
eslavich: I'm using `importlib.resources.files`, which is implemented in the shim library but only becomes available in 3.9.

asdf/resource.py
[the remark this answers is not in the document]
jdavies-st: I think that might be a bit clearer.  Or a directory tree-like object.

asdf/resource.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Parameters
+    ----------
+    root : str or importlib.resources.abc.Traversable
```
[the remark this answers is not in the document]
jdavies-st: Or maybe not.  Building the docs seems to link it without the `~`

setup.cfg
jdavies-st: Ah, ok!  Interesting.  Didn't know about that feature.

jdavies-st:
Oh, and that

```
asdf.get_config().add_resource_mapping({"http://somewhere.org/schemas/foo-1.0.0": schema})
```

feature is really cool and will make testing much, much easier.  🚀

asdf/entry_points.py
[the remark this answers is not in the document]
perrygreenfield: That is conventional practice in Python.

asdf/resource.py
```diff
@@ -0,0 +1,180 @@
...
+            self._root = Path(root)
+        else:
+            self._root = root
+
+        if uri_prefix.endswith("/"):
+            self._uri_prefix = uri_prefix[:-1]
+        else:
+            self._uri_prefix = uri_prefix
+
+        for traversable, path in self._iterate_files(self._root, []):
```
perrygreenfield, line 62: I find the use of the name traversable when a file (or equivalent is returned) misleading. Shouldn't it be something based on file? (name, that is)?

asdf/resource.py
```diff
@@ -0,0 +1,180 @@
...
+    Parameters
+    ----------
+    resource_mappings : list of collections.abc.Mapping
+        Underlying resource mappings.  In the case of
+        a duplicate URI, the latest mapping in the list
+        will override.
+    """
+    def __init__(self, resource_mappings):
+        self._resource_mappings = resource_mappings
+
```
perrygreenfield, line 114: It might be good to explain why the two level mechanism has to be retained. In other words, why not create a dictionary keyed by URI to the end object immediately? Is the idea to allow changing the individual mappings in some way after the fact?

asdf/tests/conftest.py
jdavies-st: I'm just worried that other packages will copy our test style and not realize we have an autouse fixture (and that it is needed if not using a context manager).  I'm wondering if we should eat our own dog food in our tests instead.

asdf/resource.py
eslavich: I think it would be fair to change this from `traversable` to `file` (and maybe the name of the class from `TraversableResourceMapping` to `DirectoryResourceMapping`.  I was trying to show that it works with any implementation of [importlib.abc.Traversable](https://github.com/python/cpython/blob/3.9/Lib/importlib/abc.py#L393), but that's a bit esoteric given that this interface is introduced in the yet-unreleased Python 3.9.  I can document the possible use of a non-directory object in one of the docstrings.

asdf/resource.py
eslavich: My thought was to avoid loading the bytes until needed, in case the resources are being requested over HTTP or are otherwise slow to access.  Most resources won't be needed in a given session so loading them all up-front is unnecessary and possibly time consuming.  Implemented this way, only the URIs are read on load.

asdf/tests/conftest.py
eslavich: I hadn't thought of that angle.  I'd prefer to keep this fixture in place, in case someone forgets to use a context manager, but I'll go ahead and update the tests with on-brand dog food :)

asdf/resource.py
eslavich: Changed the name to `DirectoryResourceMapping`, and added a test that confirms that the class still works with a non-pathlib Traversable.

asdf/tests/conftest.py
eslavich: (updated, and added a comment about the autouse fixture at the end of the test that sets a global config item)

asdf/resource.py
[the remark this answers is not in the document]
eslavich: Changed this to "from a directory or directory tree."

eslavich:
I'm going to file an issue to address this:

> Overall this look good. Would be nice to add some of the content from the PR message, including the example usage either into a docstring somewhere or into the documentation (or both). Reading the code as it is written makes the usage not particularly clear.

I think it'll be more efficient to document the overall extending-the-library process once all the new APIs are finished.

PR #822, opened 2020-07-10
Clean up tox env dependencies

jdavies-st:


jdavies-st:
I see the MacOS build is failing for unrelated issues, but it looks like you've (at least temporarily) fixed that in another PR.  So this should be good to merge.

Not that the `prelease` build is failing because `pytest-doctestplus` is not compatible with `pytest 6.0rc1`.  Again, unrelated to this PR.

PR #825, opened 2020-07-13
Drop support for automatic serialization of subclass attributes

eslavich:
This PR bids farewell to the experimental feature that performed automatic serialization of subclass attributes.


eslavich:
> If a file has been written out using a subclass attr, will it now just get read in as a raw dict?

Since it still bears the tag associated with the parent class, then the parent class will be instantiated and the subclass attributes will be lost.  That'd be pretty bad, but since the feature was marked experimental and I can't find any evidence of anyone using it, I don't think we need to worry too much about that case.

PR #827, opened 2020-07-14
Allow arrays to be written to blocks in non-contiguous order

### LINKED ISSUE #826: [Bug] Wrong seriaization of higher dimensional ndarrays that have been transposed multiple times
Have a look at the following script:

~~~ python
from io import BytesIO

import asdf
import numpy as np


def print_array_information(name, array):
    print(name)
    print("---")
    print("data[0]")
    print(array[0])
    print(f"data c_contiguous: {array.flags.c_contiguous}")
    if array.base is not None:
        print("base[0]")
        print(array.base[0])
        print(f"base c_contiguous: {array.base.flags.c_contiguous}")
    else:
        print("base is 'None'")
    print("\n")


arr1 = np.array([[[0, 1], [-1, 0]], [[0, 1], [-1, 0]]], dtype=float)
print_array_information("array", arr1)


arr2 = arr1.transpose(0, 2, 1)
print_array_information("transposed array", arr2)


arr3 = arr2 * 1
print_array_information("multiplied transposed array", arr3)


arr4 = arr3.transpose(0, 2, 1)
print_array_information("transposed multiplied transposed array", arr4)


assert np.array_equal(arr1, arr4)


tree = {"root": arr4}

# Write the data to buffer
with asdf.AsdfFile(tree, ignore_version_mismatch=False,) as ff:
    buff = BytesIO()
    ff.write_to(buff)
    buff.seek(0)

# read back data from ASDF file contents in buffer
with asdf.open(buff, copy_arrays=True, lazy_load=False) as af:
    data = af.tree
    print(data["root"])
    assert np.array_equal(arr4, data["root"])
~~~

All it does is that it creates a 3d array, swaps the 2 innermost dimensions, multiplies it by 1, swaps the innermost dimensions again, and writes the resulting array `arr4`, which is equal to the original array `arr1`, to an asdf file/buffer. If you restore the data from the file/buffer, the assertion `assert np.array_equal(arr4, data["root"])` fails. Observing the data shows, that the innermost dimensions are transposed. The outputs of the `print_array_information` method show where the issue comes from. 

relevant `arr1` output:
~~~
data c_contiguous: True
base is 'None'
~~~

relevant `arr2` output:
~~~
data c_contiguous: False
base[0]
[[ 0.  1.]
 [-1.  0.]]
base c_contiguous: True
~~~

relevant `arr3` output:
~~~
data c_contiguous: False
base is 'None'
~~~

relevant `arr4` output:
~~~
data c_contiguous: True
base[0]
[[ 0. -1.]
 [ 1.  0.]]
base c_contiguous: False
~~~

The problem here is the following: Transposing a matrix "destroys" the "c continuity". Transposing it twice restores it. Multiplication, division, and probably several other operations keep the current "c continuity"/data interpretation but write to a new block of memory, which resets the arrays `base`.
Because of this,  `arr3.base` is `None` but `arr3.flags.c_contiguous` is `False` as it is for `arr2`, even though it uses a fresh memory location. Now if we transpose again, `arr4` has  `c_contiguous` data but a non `c_contiguous` base. Why is this a problem and causes an error?

I have looked into the ndarray `to_tree` method.

https://github.com/asdf-format/asdf/blob/f5b5900dc45f0668d04281203ae1d9a1c0f8ea1c/asdf/tags/core/ndarray.py#L418-L432

Here you only check if `data` is  `c_contiguous` (line 425) and correct it if not. However, in the `find_or_create_block_for_array` method, which is used afterwards to get the data that should be written into the file (if I am not mistaken), you use the data of `base` (line738 - below) without checking if it's underlying order/"c continuity" is still correct:

https://github.com/asdf-format/asdf/blob/f5b5900dc45f0668d04281203ae1d9a1c0f8ea1c/asdf/block.py#L716-L745

I guess this is what causes the error. Maybe a simple fix would be to add this:

~~~
if not base.flags.c_contiguous:
    base = np.array(base)
~~~

or something similar after line 738, but I don't know if that would break some other stuff. However, creating a new numpy array with the "constructor"/`__init__` method will fix the `c_contiguous` issue of the base, since it resets it. When you replace `arr4 = arr3.transpose(0, 2, 1)` with `arr4 = np.array(arr3.transpose(0, 2, 1))` in the script, the assertion passes.

**Additional Notes**

If one removes the second matrix or reduces the dimension to 2d, the error disappears. So it seems to be a problem of higher dimensional arrays.

For reference, here are the version numbers of my setup:

~~~
Python Version: 3.7.7
Numpy Version: 1.18.1
ASDF Version: 2.6.0
~~~

eslavich:
This PR allows binary blocks to be written to disk in non-contiguous order.  Previously we had allowed Fortran order but converted to C order in all other cases, and this causes problems when a view over a block is depending on the original order.

While I was investigating this one, I found another bug that produced unreadable ASDF files when an array in the tree was both offset and broadcasted.

Resolves #826

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,12 @@
 - Fix bug in version map caching that caused incompatible
   tags to be written under ASDF Standard 1.0.0. [#821]
 
+- Fix bug that corrupted ndarrays when the underlying block
+  array was converted to C order on write. [#827]
+
+# Fix bug that produced unreadable ASDF files when an
+  ndarray in the tree was both offset and broadcasted. [#827]
+
 2.6.1 (unreleased)
 ------------------
 
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -790,13 +790,7 @@ class Block:
 
     def __init__(self, data=None, uri=None, array_storage='internal',
                  memmap=True, lazy_load=True):
-        if isinstance(data, np.ndarray) and not data.flags.c_contiguous:
-            if data.flags.f_contiguous:
-                self._data = np.asfortranarray(data)
-            else:
-                self._data = np.ascontiguousarray(data)
-        else:
-            self._data = data
+        self._data = data
         self._uri = uri
         self._array_storage = array_storage
 
@@ -916,7 +910,7 @@ class Block:
 
     def _calculate_checksum(self, data):
         m = hashlib.new('md5')
-        m.update(self.data.flatten())
+        m.update(self.data.ravel('K'))
         return m.digest()
 
     def validate_checksum(self):
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -93,7 +93,7 @@ size : integer
 
 
 def _array_tofile_chunked(write, array, chunksize):  # pragma: no cover
-    array = array.view(np.uint8).flatten()
+    array = array.view(np.uint8)
     for i in range(0, array.nbytes, chunksize):
         write(array[i:i + chunksize].data)
 
@@ -370,7 +370,7 @@ class GenericFile(metaclass=util.InheritDocstrings):
     """
 
     def write_array(self, array):
-        _array_tofile(None, self.write, array.ravel(order='A'))
+        _array_tofile(None, self.write, array.ravel(order='K'))
 
     def seek(self, offset, whence=0):
         """
@@ -749,7 +749,7 @@ class RealFile(RandomAccessFile):
             arr.flush()
             self.fast_forward(len(arr.data))
         else:
-            _array_tofile(self._fd, self._fd.write, arr.ravel(order='A'))
+            _array_tofile(self._fd, self._fd.write, arr.ravel(order='K'))
 
     def can_memmap(self):
         return True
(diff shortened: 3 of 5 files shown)


eslavich:
> How is this going to handle old files that do these operations? Or are old files already corrupt since they were written out with buggy code.

Unfortunately, I think it's the second case -- any such files were corrupted on write and now we have no way to distinguish them from files that were written that way intentionally.

nden:
This may have caused errors in the jwst pipeline 

https://travis-ci.org/github/spacetelescope/jwst/jobs/709569204

eslavich:
> This may have caused errors in the jwst pipeline

I _think_ those errors are all due to #785, which was merged at around the same time as this PR.  The jwst schemas contain default values the aren't valid against the schema, and we're only just now catching those problems.  For example, `jwst/datamodels/schemas/amilg.schema.yaml` contains this snippet:

```
    fit_image:
      title: Fitted image
      fits_hdu: FIT
      default: 0.0
      ndim: 2
      datatype: float32
```
but `0.0` isn't a valid 2-dimensional ndarray.  Perhaps we ought to add an option to the pytest plugin to control default-validation?  Since we're not going to be filling defaults going forward, it may make sense in many cases to treat the field more like a descriptive annotation than a literal default value.

PR #828, opened 2020-07-15
Check resource manager before passing schema URI through resolver

eslavich:
This PR is a minor adjustment to #819.  I had intended to use schemas from the new API before resorting to the old API, but my original implementation passed the schema URI through the old resolver first, which may be converting it to a filesystem URL that the new code can't recognize.  This change is simply to check the schemas from the new API first, and only pass the URI through the resolver if necessary.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@
 ------------------
 
 - Add new Mapping API for extending asdf with additional
-  schemas. [#819]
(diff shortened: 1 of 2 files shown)


jdavies-st:
This looks good.  Is `codecov` no longer set up on this repo?

eslavich:
Closing and re-opening this to see if codecov is fixed

PR #831, opened 2020-07-20
Make validation of schema default configurable

eslavich:
A [recent bug fix](https://github.com/asdf-format/asdf/pull/785) made this check more strict, which is causing tests in jwst to fail.  This PR adds an option to disable validation of the schema `default` property for projects that want to use a descriptive `default` value instead of providing a literal default.

### DIFF
--- a/pytest_asdf/plugin.py
+++ b/pytest_asdf/plugin.py
@@ -26,6 +26,12 @@ def pytest_addoption(parser):
         default=False,
     )
     parser.addini(
+        "asdf_schema_validate_default",
+        "Set to true to enable validation of the schema 'default' property",
+        type="bool",
+        default=True,
+    )
+    parser.addini(
         "asdf_schema_ignore_unrecognized_tag",
         "Set to true to disable warnings when tag serializers are missing",
         type="bool",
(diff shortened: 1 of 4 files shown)


jdavies-st:
We should add this functionality over in 

https://github.com/asdf-format/asdf-schema-helpers/blob/master/src/pytest_asdf_schema/plugin.py

as well.

pytest_asdf/plugin.py
```diff
@@ -25,6 +25,12 @@ def pytest_addoption(parser):
         type="bool",
         default=False,
     )
+    parser.addini(
+        "asdf_schema_validate_default",
+        "Set to true to enable validation of the schema 'default' property",
+        type="bool",
+        default=True,
```
jdavies-st, line 32: Shouldn't the default here be `False`, as that was the previous behavior?

jdavies-st: Or do we not care since this does not affect runtime code and is only an issue for testing?

eslavich: The previous behavior wasn't quite `False` because it was successfully checking non-object defaults against the schema.  I'm not sure there's a good answer.

eslavich:
> We should add this functionality over in

Yes!  Thanks for the reminder.  I'll create an issue right now.

PR #832, opened 2020-07-22
Prevent errors when extra properties are present in extension metadata

eslavich:
The next version of ASDF is going to add some additional properties to extension metadata, which the schema permits but would cause the library to raise an error due to the way the `ExtensionMetadata` class was initialized.  This PR changes the `ExtensionMetadata` class to a dict (like the other core objects) so that it can hold additional properties without complaint.

I'd like to get this into the 2.7 release so that files produced by 2.8 do not cause these errors.

### DIFF
--- a/asdf/tags/core/__init__.py
+++ b/asdf/tags/core/__init__.py
@@ -41,25 +41,17 @@ class HistoryEntry(dict, AsdfType):
     version = '1.0.0'
 
 
-class ExtensionMetadata(AsdfType):
+class ExtensionMetadata(dict, AsdfType):
     name = 'core/extension_metadata'
     version = '1.0.0'
 
-    def __init__(self, extension_class=None, software={}):
-        self.extension_class = extension_class
-        self.software = software
+    @property
+    def extension_class(self):
+        return self['extension_class']
 
-    @classmethod
-    def from_tree(cls, node, ctx):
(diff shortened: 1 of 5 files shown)


asdf/tags/core/__init__.py
```diff
@@ -41,25 +41,17 @@ class HistoryEntry(dict, AsdfType):
...
+class ExtensionMetadata(dict, AsdfType):
     name = 'core/extension_metadata'
     version = '1.0.0'
 
-    def __init__(self, extension_class=None, software={}):
-        self.extension_class = extension_class
-        self.software = software
+    @property
+    def extension_class(self):
+        return self['extension_class']
```
jdavies-st, line 50: If `extension_class` is not passed in the constructor, shouldn't it return `None`?  Something like

```
        return self.get("extension_class", None)
```

eslavich: `extension_class` is required by the schema:

https://asdf-standard.readthedocs.io/en/1.5.0/generated/stsci.edu/asdf/core/extension_metadata-1.0.0.html#Original%20Schema

so we would have hit a validation error before this point if the property was missing.

jdavies-st: Ah, ok.  No problem then.

PR #835, opened 2020-07-28
Initial refactor of extensions

eslavich:
In an attempt to make this easier to review, I pulled out the following changes into a separate PR:

- Files are now written with only the extensions used listed in the metadata (instead of every extension used at some point in the session)
- Create wrapper for resource Mappings that carries around the package name and version of the entry point (this is just a debugging tool)
- Move code that fetches extensions from entry points to the entry_points module
- Create new extension entry point ("asdf.extensions" vs "asdf_extensions").  Extensions from the old entry point are marked "legacy" and behave differently.  At this point, the only difference is that "legacy" extensions are always enabled when reading and writing files, and new-style extensions are enabled based on the file's metadata.
- Create config property that returns the list of extensions.  AsdfFile now retrieves the extensions from the config object, which means asdf.extension.default_extensions is no longer needed.  
- Create wrapper for AsdfExtension that carries package name/version and also provides default implementations for optional methods.
- Use this wrapper as the source of extension metadata information instead of maintaining a dict on AsdfExtensionList and AsdfFile.
- Remove warning when reading a file with an older version of the package that produced it.  In a future PR we'll introduce a URI that will allow extensions themselves to be versioned, so the package version won't be relevant.
- Restrict extension lists provided by users to list of AsdfExtension.  This is how the API is documented, but currently it is possible to provide a single AsdfExtension or an AsdfExtensionList. 
- Change the AsdfFile `_extensions` attribute from AsdfExtensionList to list of AsdfExtension.  The AsdfExtensionList is now stored in `_extension_list`.  In a future PR I'll be adding a new class that collects a group of enabled extensions.
- Add `always_enabled` property to AsdfExtension that instructs the library to use the extension even when not requested by the user or the file metadata.  This is the behavior of all extensions thus far, so the property defaults to `True` for legacy extensions.
- Add `legacy_class_names` property to AsdfExtension.  These are additional class names that will allow older metadata to match to a new-style extension.

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -4,7 +4,6 @@ import time
 import copy
 import datetime
 import warnings
-from pkg_resources import parse_version
 
 import numpy as np
 from jsonschema import ValidationError
@@ -20,9 +19,10 @@ from . import util
 from . import version
 from . import versioning
 from . import yamlutil
+from . import tagged
 from . import _display as display
 from .exceptions import AsdfDeprecationWarning, AsdfWarning, AsdfConversionWarning
-from .extension import AsdfExtensionList, default_extensions
+from .extension import AsdfExtensionList, ExtensionProxy
 from .util import NotSet
 from .search import AsdfSearchResult
 
@@ -107,8 +107,7 @@ class AsdfFile(versioning.VersionedMixin):
             standard.
 
         """
-        self._extensions = []
-        self._extension_metadata = {}
+        # This method sets the _extensions and _extension_list attributes.
         self._process_extensions(extensions)
 
         if custom_schema is not None:
@@ -143,7 +142,7 @@ class AsdfFile(versioning.VersionedMixin):
             # an empty tree.
             self._tree = AsdfObject()
         elif isinstance(tree, AsdfFile):
-            if self._extensions != tree._extensions:
+            if set(e.delegate for e in self._extensions) != set(e.delegate for e in tree._extensions):
                 raise ValueError(
                     "Can not copy AsdfFile and change active extensions")
             self._uri = tree.uri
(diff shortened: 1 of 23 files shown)


asdf/asdf.py
```diff
@@ -802,7 +764,21 @@ def _write_tree(self, tree, fd, pad_blocks):
         fd.write(b'\n')
 
         if len(tree):
-            yamlutil.dump_tree(tree, fd, self)
+            # yamlutil.custom_tree_to_tagged_tree looks for this attribute
+            # on the AsdfFile and if present, passes it to TypeIndex to
+            # record extensions used for conversion.
+            self._extensions_used = set()
```
eslavich, line 770: This is awful, but it's the way we currently handle state that only exists during serialization and I don't want to rock that particular boat right this moment.  I'd like to create a new class that gets passed around during serialization instead of the AsdfFile itself.

asdf/extension.py
```diff
@@ -107,24 +107,88 @@ def url_mapping(self):
...
+        for extension in self._extensions:
             if not isinstance(extension, AsdfExtension):
                 raise TypeError(
                     "Extension must implement asdf.types.AsdfExtension "
                     "interface")
             tag_mapping.extend(extension.tag_mapping)
-            # New-style extensions will not include a url_mapping attribute.
-            if hasattr(extension, "url_mapping"):
-                url_mapping.extend(extension.url_mapping)
+            url_mapping.extend(extension.url_mapping)
```
eslavich, line 191: This is the benefit of wrapping the extensions in that `ExtensionProxy` object, we don't have to make these `hasattr` checks all over the place.  We're going to have more and more optional and deprecated properties on these things and I think this will save some sanity.

asdf/resource.py
```diff
@@ -25,6 +26,61 @@
 ]
 
 
+class ResourceMappingProxy(Mapping):
```
eslavich, line 29: There's less of a strong case for this wrapper, but since the plugin can be a simple dict, it's not obvious what extension package provided them.  This'll give us some clues when we're trying to figure out where a busted resource plugin is coming from.

asdf/type_index.py
```diff
@@ -47,7 +47,6 @@ def __init__(self, version, index):
         self._class_by_subclass = {}
         self._types_with_dynamic_subclasses = {}
         self._extension_by_cls = {}
-        self._extensions_used = set()
```
eslavich, line 50: Keeping this state here is a bad idea because this object gets reused by all `AsdfFile` in the session, so files end up with extensions in their metadata that weren't actually used when serializing the file.

jdavies-st: Ah, this explains why I was seeing extensions not needed listed for a particular file. I had assumed this was a general bad feature, not that it was a path-dependent bug.

eslavich:
I need to add something to this, I'll comment again when I'm done

eslavich:
Ok, this is ready to go.  It needs some additional tests, but that's hard to do right now b/c there isn't any infrastructure for adding extensions to the config at runtime.  That'll show up two PRs from now and at that point I'll add the missing test coverage.

eslavich:
I just finished implementing support for the converters, and some of the code in this PR had to change.  Since review hasn't started, I'm going to close this and reopen with the updated code in even more bite-sized pieces.

PR #836, opened 2020-07-28
Add asdf_standard_requirement property to AsdfExtension interface

eslavich:
This PR adds an `asdf_standard_requirement` property to `AsdfExtension` that will allow an extension to specify the ASDF Standard version(s) that it supports.  This will be needed for the transform extensions, some of which pair with one ASDF Standard version only.

Opened against #835's branch for now, I'll switch this to master once that one's been merged.


eslavich:
Closing for now, will re-open soon

PR #845, opened 2020-08-03
Use __qualname__ when building fully qualified class names

eslavich:
Previously an inner class name wouldn't include the fact that it was nested in an outer class:

```python
# some_module.py
class SomeClass:
  class SomeInnerClass:
    pass
```

The name of `SomeClass` has been correct, but `SomeInnerClass` would come out as `some_module.SomeInnerClass` instead of `some_module.SomeClass.SomeInnerClass`.  This PR changes it to the latter.

### DIFF
--- a/asdf/tests/test_util.py
+++ b/asdf/tests/test_util.py
@@ -1,4 +1,5 @@
 from asdf import util
+from asdf.extension import BuiltinExtension
 
 
 def test_is_primitive():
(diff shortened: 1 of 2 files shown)


eslavich:
I'm going to merge this now, but let me know if you have any concerns.  Happy to revise it in a subsequent PR.

PR #847, opened 2020-08-03
Allow installed extensions to be configured with AsdfConfig

eslavich:
This PR refactors the existing extension code to make the new API easier to implement.  Changes include:
- Move extension loading from `_DefaultExtensions` to the new `asdf.entry_points` module
- Move list of registered extensions from `_DefaultExtensions` to `AsdfConfig`
- Remove most of the code from `_DefaultExtensions` so that it becomes a thin interface around `AsdfConfig`
- Add methods for adding and removing registered extensions at runtime
- Add wrapper that allows extensions to carry around their own metadata

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -22,7 +22,7 @@ from . import versioning
 from . import yamlutil
 from . import _display as display
 from .exceptions import AsdfDeprecationWarning, AsdfWarning, AsdfConversionWarning
-from .extension import AsdfExtensionList, default_extensions
+from .extension import AsdfExtensionList, AsdfExtension, ExtensionProxy
 from .util import NotSet
 from .search import AsdfSearchResult
 from ._helpers import validate_version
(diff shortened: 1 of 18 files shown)


asdf/extension/_legacy.py
```diff
@@ -186,65 +184,27 @@ def url_mapping(self):
...
+        from ..config import get_config
+        return [e for e in get_config().extensions if e.legacy]
 
     @property
     def extension_list(self):
-        if self._extension_list is None:
-            self._extension_list = AsdfExtensionList(self.extensions)
-
-        return self._extension_list
+        return AsdfExtensionList(self.extensions)
```
perrygreenfield, line 194: This is exceedingly too clever for me. This class(?) property uses self.extensions, but the class has no __init__ nor inherits from another class, and yet has that attribute? I need to catch up on my meta meta classes.

eslavich: The extensions property is still there!  It's just obscured on GitHub by the ugly diff.

https://github.com/asdf-format/asdf/blob/c900f3fd2e3fd472df37f2ba21d60b9a29c571b1/asdf/extension/_legacy.py#L188

perrygreenfield: Indeed. Sorry...

asdf/asdf.py
```diff
@@ -214,8 +247,10 @@ def _check_extensions(self, tree, strict=False):
             return
 
         for extension in tree['history']['extensions']:
+            installed = next((e for e in self.extensions if e.class_name == extension.extension_class), None)
```
perrygreenfield, line 250: I guess I would prefer a bit more verbose construction than using next() to provide the default None. But that's just me.

eslavich: This will be changing in a future PR anyway, since the condition gets too complex for one line.  I'll leave it alone for the moment.

PR #848, opened 2020-08-04
Include only extensions used in metadata + introduce SerializationContext

eslavich:
This PR continues from #847 and fixes the issue where extensions not used during serialization were written to the file metadata.  The fix necessitated adding a hook to `yamlutil.dump_tree` to modify the tagged tree after it had been converted, since only then do we know what extensions were used.

My long-term goal for the new `SerializationContext` class is for us to pass that object around instead of the `AsdfFile` itself.  Currently when `AsdfFile.write_to` is called, it modifies the `AsdfFile` instance in order to write with the specified parameters.  Some of these modifications are permanent, which breaks the contract of `write_to` (it's documented as an "export" type of operation that doesn't change the `AsdfFile`), and some are merely awkward, like these:

```python
    def _post_write(self, fd):
        if len(self._tree):
            self.run_hook('post_write')

        # TODO: there has got to be a better way to do this...
        if hasattr(self, '_all_array_storage'):
            del self._all_array_storage
        if hasattr(self, '_all_array_compression'):
            del self._all_array_compression
        if hasattr(self, '_auto_inline'):
            del self._auto_inline
```

The new `Converter` API will receive an instance of `SerializationContext` in its `ctx` argument instead of `AsdfFile`, and in 3.0 all `ctx` arguments will contain `SerializationContext`.

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -306,8 +306,8 @@ class AsdfFile:
 
         return extensions
 
-    def _update_extension_history(self):
-        if self.version < versioning.NEW_HISTORY_FORMAT_MIN_VERSION:
+    def _update_extension_history(self, serialization_context):
+        if serialization_context.version < versioning.NEW_HISTORY_FORMAT_MIN_VERSION:
             return
 
         if 'history' not in self.tree:
@@ -323,7 +323,7 @@ class AsdfFile:
         elif 'extensions' not in self.tree['history']:
             self.tree['history']['extensions'] = []
 
-        for extension in self.type_index.get_extensions_used():
+        for extension in serialization_context.extensions_used:
             ext_name = extension.class_name
             ext_meta = ExtensionMetadata(extension_class=ext_name)
             if extension.package_name is not None:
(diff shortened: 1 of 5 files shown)


asdf/yamlutil.py
```diff
@@ -318,11 +324,11 @@ def dump_tree(tree, fd, ctx):
 
     ctx : Context
         The writing context.
-    """
-    class AsdfDumperTmp(AsdfDumper):
```
eslavich, line 322: This is just a thing I noticed while I was here.  At some point I changed `AsdfDumper` to not use the `ctx` variable, but then forgot to remove this special class definition.

eslavich:
A person might wonder where the tests are for this change.  It's difficult to test, since the problem only occurs when multiple extensions are loaded via entry points.  There is already a nice test for extensions that are passed to `AsdfFile`:

https://github.com/eslavich/asdf/blob/eslavich-add-serialization-context/asdf/tags/core/tests/test_history.py#L255

I'm content with that, but let me know if you think I should figure out a way to test the entry points case.

asdf/asdf.py
```diff
@@ -876,7 +876,23 @@ def _write_tree(self, tree, fd, pad_blocks):
         fd.write(b'\n')
 
         if len(tree):
-            yamlutil.dump_tree(tree, fd, self)
+            serialization_context = self._create_serialization_context()
+
+            def _tree_finalizer(tagged_tree):
```
jdavies-st, line 881: A minimal docstring here for `_tree_finalizer()` would be useful to indicate what is happening.

asdf/yamlutil.py
```diff
@@ -318,11 +324,11 @@ def dump_tree(tree, fd, ctx):
...
-        pass
-    AsdfDumperTmp.ctx = ctx
 
+    tree_finalizer : callable, optional
+        Callback that receives the tagged tree before it is validated
+        and defaults are removed.
```
jdavies-st, line 330: It receives the tagged tree, does something with it, then what?  I.e. what is the purpose of this hook?

asdf/tests/test_asdf.py
```diff
@@ -100,3 +100,23 @@ class FooExtension:
...
             with open_asdf(path, extensions=arg) as af:
                 pass
+
+
+def test_serialization_context():
+    context = SerializationContext("1.4.0")
+    assert context.version == "1.4.0"
+    assert context.extensions_used == set()
+
+    extension = get_config().extensions[0]
```
jdavies-st, line 110: What extension is used here?

eslavich: It depends on the entry points load order and what the user has installed.  For the purposes of this test it just needs to be any valid extension.  There will always be at least one due to the built-in extension.

asdf/yamlutil.py
eslavich: I'll add a note explaining what the library does with it.

PR #849, opened 2020-08-04
Wait to assign extensions until version is known

eslavich:
This PR continues from #848 and changes `asdf.open` to wait to assign extensions to the `AsdfFile` instance until the version is known.  In the future, extensions will have ASDF Standard version requirements, and assigning them early will result in unnecessary warnings.

This change won't be needed until the new extension API comes, but it's a nicely self-contained bit of work so I thought I'd separate it to make review easier.


eslavich:
> Calling `AsdfFile` and passing extensions was the only way to do it initially.

Ah ha so that's where it comes from.  Entry points sure are convenient by comparison.

PR #850, opened 2020-08-05
Implement support for extension URIs

eslavich:
This PR is branched off of #849.  Here I'm adding support for extension URIs, where are defined as an attribute on the extension instance and are written to the file metadata.  Technically this is part of the new extension API, but the change fit nicely into a separate PR.

There are two new extension attributes:

- `extension_uri`: The URI of the "extension to the ASDF Standard" that the extension class implements.  Something like "asdf://somewhere.org/extensions/foo-1.0".
- `legacy_class_names`: The fully qualified names of extension classes that provided this functionality before the advent of URIs.  This allows us to switch to URIs and change the class name without incurring warnings from the library.

Additionally I changed `AsdfFile.__init__` and `asdf.open` to accept extension URIs in addition to extension instances.

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -250,6 +250,19 @@ class AsdfFile:
         self.close()
 
     def _check_extensions(self, tree, strict=False):
+        """
+        Compare the user's installed extensions to metadata in the tree
+        and warn when a) an extension is missing or b) an extension is
+        present but the file was written with a later version of the
+        extension's package.
+
+        Parameters
+        ----------
+        tree : AsdfObject
+            Fully converted tree of custom types.
+        strict : bool, optional
+            Set to `True` to convert warnings to exceptions.
+        """
         if 'history' not in tree or not isinstance(tree['history'], dict) or \
                 'extensions' not in tree['history']:
             return
@@ -305,6 +318,21 @@ class AsdfFile:
                         warnings.warn(msg, AsdfWarning)
 
     def _process_extensions(self, requested_extensions):
+        """
+        Validate a list of extensions requested by the user and
+        add missing extensions registered with the current `AsdfConfig`.
+
+        Parameters
+        ----------
+        requested_extensions : object
+            May be any of the following: `asdf.extension.AsdfExtension`, `str`
+            extension URI, `asdf.extension.AsdfExtensionList` or a `list`
+            of URIs and/or extensions.
+
+        Returns
+        -------
+        list of asdf.extension.AsdfExtension
+        """
         if requested_extensions is None:
             requested_extensions = []
         elif isinstance(requested_extensions, (AsdfExtension, ExtensionProxy, str)):
@@ -336,6 +364,15 @@ class AsdfFile:
         return extensions
 
     def _update_extension_history(self, serialization_context):
+        """
+        Update the extension metadata on this file's tree to reflect
+        extensions used during serialization.
+
+        Parameters
+        ----------
+        serialization_context : asdf.asdf.SerializationContext
+            The context that was used to serialize the tree.
+        """
         if serialization_context.version < versioning.NEW_HISTORY_FORMAT_MIN_VERSION:
             return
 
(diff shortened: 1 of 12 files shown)


asdf/yamlutil.py
```diff
@@ -318,11 +324,11 @@ def dump_tree(tree, fd, ctx):
...
         The writing context.
-    """
-    class AsdfDumperTmp(AsdfDumper):
-        pass
-    AsdfDumperTmp.ctx = ctx
 
+    tree_finalizer : callable, optional
+        Callback that receives the tagged tree before it is validated
+        and defaults are removed.
+    """
```
perrygreenfield, line 331: Missing the description for the _serialization_context arguement.

asdf/type_index.py
```diff
@@ -131,19 +131,22 @@ def _add_by_tag(self, index, name, version):
...
-    def _mark_used_extension(self, custom_type):
-        self._extensions_used.add(self._extension_by_cls[custom_type])
+    def _mark_used_extension(self, custom_type, serialization_context):
+        extension = self._extension_by_cls[custom_type]
+        self._extensions_used.add(extension)
+        if serialization_context is not None:
+            serialization_context.mark_extension_used(extension)
 
-    def _process_dynamic_subclass(self, custom_type):
+    def _process_dynamic_subclass(self, custom_type, serialization_context):
```
perrygreenfield, line 140: Somewhere the concept of dynamic subclasses should be explained. In my search of the code, I didn't easily find it. Perhaps better explained in the developer documentation; more on that later

asdf/type_index.py
```diff
@@ -163,17 +166,17 @@ def from_custom_type(self, custom_type):
                 # includes classes that are created dynamically post
```
perrygreenfield, line 166: Like what exactly? There is no longer a _CompoundModel class, and no current subclasses of CompoundModel that I'm aware of. Or is this just to handle older versions of astropy?

asdf/yamlutil.py
eslavich: I omitted this intentionally, since it's not part of the public API.  I didn't want to advertise it to external developers who might be using the method, since it's going to disappear in 3.0 anyway.  Bad idea?

perrygreenfield: OK, perhaps a comment after the docstring should clarify the intent

asdf/type_index.py
eslavich: I'm a little hazy on this myself, but I _think_ the intention of this feature was to allow serialization of subclasses that didn't exist at the time that the `TypeIndex` was constructed.  When building the type index, we inspect the `__subclasses__` attribute of each type and insert all the sub-types into a dict for fast lookup.  But that only finds subclasses that have already been imported.  Setting `handle_dynamic_subclasses = True` causes the type index to make an actual `issubclass` check against the listed types so it can catch subclasses that were imported later.  It's slower but more comprehensive.

The new API doesn't support any brand of automatically serializing sub-types.  A type must be explicitly listed in a converter in order to be serialized.  So when `TypeIndex` is dropped in 3.0 we won't have to worry about this import order issue anymore.

asdf/type_index.py
eslavich: See my comment above, but I suspect the problem this was intended to solve was an import order issue.  Without this feature, the following would succeed:

```python
import SomeCustomCompoundModelSubclass
import asdf.type_index.TypeIndex

type_index = TypeIndex(...)

assert type_index.from_custom_type(SomeCustomCompoundModelSubclass) is not None
```

but this would fail:

```python
import asdf.type_index.TypeIndex

type_index = TypeIndex(...)

import SomeCustomCompoundModelSubclass

assert type_index.from_custom_type(SomeCustomCompoundModelSubclass) is not None
```

asdf/asdf.py
```diff
@@ -247,55 +250,76 @@ def _check_extensions(self, tree, strict=False):
             return
```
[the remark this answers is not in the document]
eslavich: I'll add one to `_check_extensions` and `_process_extensions` now, since I just modified them.

eslavich:
I created https://github.com/asdf-format/asdf/issues/861 as a reminder to update the developer documentation before release.

PR #854, opened 2020-08-06
Support asdf:// URI scheme

eslavich:
This PR allow asdf to properly resolve relative references against `asdf://` URIs.  The `urllib.parse` module maintains a list of URI schemes that should behave like `http` with regard to resolving relative paths, and needed `asdf` added to its list.

There's a second change in here that fixes a long-standing issue with resolving references in metaschemas.  It's common for such schemas to define an `id` property, but the asdf library didn't know not to treat that property object as a URI.  The fix is to check that `id` is a string before assuming it's a URI.

### DIFF
--- a/asdf/tests/test_generic_io.py
+++ b/asdf/tests/test_generic_io.py
@@ -453,6 +453,22 @@ def test_relative_uri():
         'http://www.google.com', 'file://local') == 'file://local'
 
 
+@pytest.mark.parametrize("protocol", ["http", "asdf"])
+def test_resolve_uri(protocol):
+    """
+    Confirm that the patched urllib.parse is handling
+    asdf:// URIs correctly.
+    """
+    assert generic_io.resolve_uri(
+        '{}://somewhere.org/some-schema'.format(protocol), '#/definitions/foo'
+    ) == '{}://somewhere.org/some-schema#/definitions/foo'.format(protocol)
+
+    assert generic_io.resolve_uri(
+        '{}://somewhere.org/path/to/some-schema'.format(protocol),
+        '../../some/other/path/to/some-other-schema'
+    ) == '{}://somewhere.org/some/other/path/to/some-other-schema'.format(protocol)
+
+
 def test_arbitrary_file_object():
     class Wrapper:
         def __init__(self, init):
--- a/asdf/util.py
+++ b/asdf/util.py
@@ -2,12 +2,26 @@ import inspect
 import math
 import struct
 import types
+import importlib.util
 
-from urllib.parse import urljoin
 from urllib.request import pathname2url
-from urllib import parse as urlparse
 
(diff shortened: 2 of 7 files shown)


asdf/util.py
```diff
@@ -2,13 +2,27 @@
...
 import numpy as np
 
+# We're importing our own copy of urllib.parse because
+# we need to patch it to support asdf:// URIs, but it'd
+# be irresponsible to do this for all users of a
+# standard library.
+urllib_parse_spec = importlib.util.find_spec('urllib.parse')
+patched_urllib_parse = importlib.util.module_from_spec(urllib_parse_spec)
+urllib_parse_spec.loader.exec_module(patched_urllib_parse)
+del urllib_parse_spec
```
jdavies-st, line 18: What the advantage of doing it this way rather than

```
import urllib.parse as patched_urllib_parse
```

It's still local to this module's namespace, right?

asdf/tests/test_generic_io.py
```diff
@@ -453,6 +453,22 @@ def test_relative_uri():
...
 
+@pytest.mark.parametrize("protocol", ["http", "asdf"])
+def test_resolve_uri(protocol):
+    """
+    Confirm that the patched urllib.parse is handling
+    asdf:// URIs correctly.
+    """
+    assert generic_io.resolve_uri(
+        '{}://somewhere.org/some-schema'.format(protocol), '#/definitions/foo'
+    ) == '{}://somewhere.org/some-schema#/definitions/foo'.format(protocol)
```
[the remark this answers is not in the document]
eslavich: You're asking the good questions.  I'll start an email discussion.

asdf/util.py
eslavich: Unfortunately while that aliases the module reference to a new name, it still points to the same module object.  Import caches the module after loading it the first time so that other imports don't have to read and parse the module's code again.

```python
# module1.py
import urllib.parse as module1_alias

# module2.py
import urllib.parse as module2_alias

# Fire up a console:
In [1]: import module1

In [2]: import module2

In [3]: module1.module1_alias is module2.module2_alias
Out[3]: True
```

jdavies-st: Ah, I see.  I didn't realize there was caching going on.  This bypasses that. #themoreyouknow

asdf/util.py
```diff
@@ -2,13 +2,27 @@
...
+
+# urllib.parse needs to know that it should treat asdf://
+# URIs like http:// URIs for the purposes of joining
+# a relative path to a base URI.
+patched_urllib_parse.uses_relative.append('asdf')
```
perrygreenfield, line 23: Digging deep into the standard library. What a naughty boy. ;-) I had to go look at the library source code to see this explained.

PR #855, opened 2020-08-06
Fix oversights in support for asdf:// URI scheme

eslavich:
It turns out my test didn't catch all the issues that come up during actual validation.  I updated the test, made some fixes, and also made sure that the patched `urilib.parse` is used everywhere.


eslavich:
> I trust you to have found all of them by searching :-)

I hope so!  In this case I had found all the `import foo.bar` imports but forgot to search for `from foo import bar`.

PR #838, opened 2020-07-29
simple version comparison for tag validation

CagtayFabry:
very simple implementation for version parsing discussed in https://github.com/asdf-format/asdf-standard/issues/271
This should enable the following syntax to work:

```yaml
tag: http://stsci.edu/schemas/asdf/core/software-* # allow every version
tag: http://stsci.edu/schemas/asdf/core/software-1 # fix major version
tag: http://stsci.edu/schemas/asdf/core/software-1.2 # fix minor version
tag: http://stsci.edu/schemas/asdf/core/software-1.2.3 # fix patch version
```

PR #858, opened 2020-08-10
Permit wildcard in tag validator URIs

eslavich:
This implements an idea from @CagtayFabry discussed in https://github.com/asdf-format/asdf-standard/issues/271.  See #838 for an alternative implementation.

The PR is branched off of #850.  The relevant changes are in the latest commit.

The idea is to allow pattern matching in the `tag` validator, so that schema authors can permit many versions of a tag.  For example, the following pattern:

```yaml
tag: asdf://somewhere.org/tags/foo-*
```

will match any of the following:

```
asdf://somewhere.org/tags/foo-1.0.0
asdf://somewhere.org/tags/foo-2.0
```

and this will match any transform tag:

```
asdf://asdf-format.org/extensions/transform/tags/*
```

which I expect will save us a lot of tedious updating of schemas when we add or update transforms.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@
 - Add new extension API to support versioned extensions.
   [#850, #851]
 
+- Permit wildcard in tag validator URIs. [#858]
+
 2.7.0 (2020-07-23)
 ------------------
 
--- a/asdf/config.py
+++ b/asdf/config.py
@@ -11,6 +11,8 @@ from .resource import ResourceMappingProxy, ResourceManager
 from . import versioning
 from ._helpers import validate_version
 from .extension import ExtensionProxy
+from . import util
+
 
 __all__ = ["AsdfConfig", "get_config", "config_context"]
 
@@ -165,7 +167,7 @@ class AsdfConfig:
         Parameters
         ----------
         extension : asdf.extension.AsdfExtension or str, optional
-            An extension instance or URI to remove.
+            An extension instance or URI or URI pattern to remove.
         package : str, optional
             Remove only extensions provided by this package.  If the `extension`
             argument is omitted, then all extensions from this package will
@@ -181,7 +183,7 @@ class AsdfConfig:
             result = True
 
             if isinstance(extension, str):
-                result = result and e.extension_uri == extension
+                result = result and util.uri_match(extension, e.extension_uri)
             elif isinstance(extension, ExtensionProxy):
                 result = result and e == extension
 
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -76,18 +76,28 @@ def _type_to_tag(type_):
     return None
 
 
-def validate_tag(validator, tagname, instance, schema):
-
+def validate_tag(validator, tag_pattern, instance, schema):
+    """
+    Implements the tag validation directive, which checks the
+    tag against a pattern which may include '*' wildcards.
+    """
     if hasattr(instance, '_tag'):
         instance_tag = instance._tag
     else:
         # Try tags for known Python builtins
         instance_tag = _type_to_tag(type(instance))
 
-    if instance_tag is not None and instance_tag != tagname:
+    if instance_tag is None:
+        yield ValidationError(
+            "mismatched tags, wanted '{}', got unhandled object type '{}'".format(
+                tag_pattern, util.get_class_name(instance)
+            )
+        )
+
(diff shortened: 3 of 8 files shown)


CagtayFabry:
Thanks for tagging me @eslavich 

I thought a little bit about https://github.com/asdf-format/asdf-standard/issues/271 over last few days but did not find the time to write down everything.
There is one point I would like to discuss regarding the `*` wildcard implementation (which I really like in this PR):
Ending a tag URI on `*` will allow very lenient pattern matching even for completely different tags/versions.

consider the following (passing) addition to `test_uri_match`:
```python
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-bar/tag-3.1", True),
```
Of course this might be intended, but what if only all versions of the actual tag `asdf://*/tags/foo` where supposed to be allowed? Currently I do not see the way to fix the tag but allow all versions. Like this behavior:
```python
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-4.9", True),
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-5.2", True),
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-bar/tag-3.1", False),
```
I am not sure about the side effects this may or may not have.

To avoid this I would propose that all `tag` strings *must* end on a valid version pattern like `-*`, `-1`, `-1.*`, `-1.2`, `-1.2.*`, `-1.2.3` including the hyphen to be more explicit about the versioning. Basically this would enforce a trailing hyphen to always separate tag URI from versioning. (similar to what I was going for in #838)

So for the above example we could get the following behavior to work (which I personally find more explicit):
```python
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-4.9", True),
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-5.2", True),
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-bar/tag-3.1", False),
("asdf://*/tags/foo*-*", "asdf://anywhere.org/tags/foo-bar/tag-3.1", True),
```

eslavich:
> Of course this might be intended, but what if only all versions of the actual tag `asdf://*/tags/foo` where supposed to be allowed? Currently I do not see the way to fix the tag but allow all versions.

Hmm... that's a good point.  I didn't think about that.

> To avoid this I would propose that all tag strings must end on a valid version pattern

I'm only resistant to this because I suspect some people will want to use versions in some other way, maybe in the path ahead of the tag.  Currently the ASDF Standard and the YAML specification both permit the tag to be any valid URI, and I'm not sure it's worth restricting them to support pattern matching on our particular notion of what a versioned URI should look like.

Here's another idea: how about we treat `*` more like a filesystem glob pattern?  So `*` would absorb any character except the `/` and stop matching if it encounters the `/`.  That interpretation would make this example work:

```python
("asdf://*/tags/foo-*", "asdf://anywhere.org/tags/foo-bar/tag-3.1", False),
```

We could even use the `**` in the globstar way to imply matching any number of nested "directories":

```python
("asdf://*/tags/**/tag-*", "asdf://anywhere.org/tags/foo/tag-3.1", True),
("asdf://*/tags/**/tag-*", "asdf://anywhere.org/tags/bar/tag-4.5", True),
```

There is still potential for overly lenient matching within the same "directory":

```python
("asdf://anywhere.org/tags/foo-*", "asdf://anywhere.org/tags/foo-3.1", True),
("asdf://anywhere.org/tags/foo-*", "asdf://anywhere.org/tags/foo-bar-4.5", True),
```

But we can enforce that this does not happen in the ASDF core tags (with an automated test) and in other officially supported ASDF extensions.  If you do the same in your project, would this system work?  Or is the matching on `foo-3.1` and `foo-bar-4.5` still problematic?

CagtayFabry:
Yes, the `*` and `**` glob-like pattern matching the way you described was exactly my idea as well when I thought about this. (I also find it very readable and intuitive in comparison to e.g. regex)

As long as the syntax is well described I don't really see any issues. Sure
```python
("asdf://anywhere.org/tags/foo-*", "asdf://anywhere.org/tags/foo-bar-4.5", True),
```
might be weird but should be avoidable when aware of the implementation.

eslavich:
Great, I'll adapt this PR to implement the glob behavior.  Thanks for helping me think this through!

PR #863, opened 2020-08-13
Support serialization of null values

### LINKED ISSUE #536: None types are not preserved
As @vmarkovtsev pointed out in #534, `None` types are currently stripped from the tree.

I don't think the standard has anything to say about this at the moment. Would it be useful if `None` values are preserved in the tree? This would probably require the definition of a `None` or `Null` tag. I need to think through the implications a little more.

cc @perrygreenfield

eslavich:
This PR resolves a long-standing issue where trees containing `None` values are not faithfully written to an ASDF file.  An example, using 2.7.0:

```python
In [3]: import asdf

In [6]: asdf.AsdfFile({"foo": None, "bar": 42}).write_to("test.asdf")

In [7]: cat test.asdf
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
  name: asdf, version: 2.8.0.dev49+ge6a0d37}
history:
  extensions:
    software: !core/software-1.0.0 {name: asdf, version: 2.8.0.dev49+ge6a0d37}
bar: 42
...

In [8]: af = asdf.open("test.asdf")

In [9]: "foo" in af
Out[9]: False
```

and with this branch:

```python
In [1]: import asdf

In [2]: asdf.AsdfFile({"foo": None, "bar": 42}).write_to("test.asdf")

In [3]: cat test.asdf
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
  name: asdf, version: 2.8.0.dev49+ge6a0d37}
history:
  extensions:
    software: !core/software-1.0.0 {name: asdf, version: 2.8.0.dev49+ge6a0d37}
bar: 42
foo: null
...

In [4]: af = asdf.open("test.asdf")

In [5]: "foo" in af
Out[5]: True
```

In one sense this is only a bug fix, since dropping keys with `None` values from the tree was probably an unintentional side effect.  But the fix requires a breaking change to `asdf.treeutil.walk_and_modify`, which now uses a new special object to indicate "drop this node from the tree" instead of using `None` for that purpose.  This has some consequences for existing `ExtensionType` subclasses, which may be counting on `None`s being removed before validation.

Here's an example of an `ExtensionType` subclass in astropy that will break due to this change:

https://github.com/astropy/astropy/blob/v4.1rc1/astropy/io/misc/asdf/tags/transform/tabular.py#L55

The `name` property may be `None`, but the code doesn't check that.  With this change a `None` name will result in a validation error because the schema doesn't permit null:

https://github.com/asdf-format/asdf-standard/blob/1.5.0/schemas/stsci.edu/asdf/transform/transform-1.2.0.yaml#L18

Resolves #536

### DIFF
--- a/asdf/treeutil.py
+++ b/asdf/treeutil.py
@@ -204,6 +204,19 @@ class _PendingValue:
 PendingValue = _PendingValue()
 
 
+class _RemoveNode:
+    """
+    Class of the RemoveNode singleton instance.  This instance is used
+    as a signal for `asdf.treeutil.walk_and_modify` to remove the
+    node received by the callback.
+    """
+    def __repr__(self):
+        return "RemoveNode"
+
+
+RemoveNode = _RemoveNode()
+
+
 def walk_and_modify(top, callback, ignore_implicit_conversion=False, postorder=True, _context=None):
     """Modify a tree by walking it with a callback function.  It also has
     the effect of doing a deep copy.
@@ -221,7 +234,8 @@ def walk_and_modify(top, callback, ignore_implicit_conversion=False, postorder=T
         - a json id (optional)
 
         It may return a different instance in order to modify the
-        tree.
+        tree.  If the singleton instance `asdf.treeutil.RemoveNode`
+        is returned, the node will be removed from the tree.
 
         The json id is the context under which any relative URLs
         should be resolved.  It may be `None` if no ids are in the file
@@ -287,7 +301,7 @@ def walk_and_modify(top, callback, ignore_implicit_conversion=False, postorder=T
                 result[key] = PendingValue
             else:
                 value = _recurse(value, json_id)
-                if value is not None:
+                if value is not RemoveNode:
                     result[key] = value
 
         yield result
@@ -297,7 +311,7 @@ def walk_and_modify(top, callback, ignore_implicit_conversion=False, postorder=T
             # be available.
             for key, value in pending_items.items():
                 value = _recurse(value, json_id)
-                if value is not None:
+                if value is not RemoveNode:
                     result[key] = value
                 else:
                     # The callback may have decided to delete
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,7 +28,15 @@
 - Stop removing schema defaults for all ASDF Standard versions,
   and automatically fill defaults only for versions <= 1.5.0. [#860]
 
-2.7.1 (2020-08-18)
+- Stop removing keys with ``None`` values from the tree on write.  This
+  fixes a long-standing issue where the tree structure is not preserved
+  on write, but will break ``ExtensionType`` subclasses that depend on
+  this behavior.  Extension developers will need to modify their
+  ``to_tree`` methods to check for ``None`` before adding a key to
+  the tree (or modify the schema to permit nulls, if that is the
+  intention). [#863]
+
+2.7.1 (unreleased)
 ------------------
 
 - Fix bug preventing access to copied array data after
--- a/asdf/tests/test_api.py
+++ b/asdf/tests/test_api.py
@@ -203,13 +203,30 @@ def test_walk_and_modify_remove_keys():
 
     def func(x):
         if x == 42:
-            return None
+            return treeutil.RemoveNode
         return x
 
     tree2 = treeutil.walk_and_modify(tree, func)
 
     assert 'foo' not in tree2
     assert 'bar' in tree2
+
+
+def test_walk_and_modify_retain_none():
+    tree = {
+        'foo': 42,
+        'bar': None
+    }
+
+    def func(x):
+        if x == 42:
+            return None
+        return x
+
+    tree2 = treeutil.walk_and_modify(tree, func)
+
+    assert tree2['foo'] is None
+    assert tree2['bar'] is None
 
 
 def test_copy(tmpdir):
@@ -558,3 +575,12 @@ def test_array_access_after_file_close(tmpdir):
     with asdf.open(path, lazy_load=False, copy_arrays=True) as af:
         tree = af.tree
     assert_array_equal(tree["data"], data)
+
+def test_none_values(tmpdir):
+    path = str(tmpdir.join("test.asdf"))
+
+    af = asdf.AsdfFile({"foo": None})
+    af.write_to(path)
+    with asdf.open(path) as af:
+        assert "foo" in af
+        assert af["foo"] is None


eslavich:
@cadair how will this change impact you?  Other than the issues in astropy, which I'll need to fix.

asdf/treeutil.py
```diff
@@ -204,6 +204,19 @@ def __repr__(self):
 PendingValue = _PendingValue()
 
 
+class _RemoveNode:
```
perrygreenfield, line 207: If someone wanted to remove a node, can't they just delete it from the tree explicitly? But this is presumably when an object is using its tag code to convert to ASDF and it wants ASDF to remove that attribute (or whatever) from its model, correct? But can't their tag code do that explicitly? But (Perry here is carrying on a discussion with himself) this is so the actual class code can signal this rather than having to go back and change the tag code? But code somewhere will have to be changed, right? [he heads to the psychiatrist for two personality disorder treatment...] This all arises from the presumption that a missing attribute will force a use of default, but we are discouraging that now, so this for some temporary mutant backward compatibility?

perrygreenfield:
So basically this now relies on pyyaml to handle None and convert it to null? If so, what a concept!

asdf/treeutil.py
eslavich: That's a good point, I had assumed from the presence of this test:

https://github.com/asdf-format/asdf/blob/master/asdf/tests/test_api.py#L198

that somebody somewhere had a use case for removing nodes from the tree, but I haven't seen it yet.  It wouldn't be a tag class (like you mentioned, there are easier ways to do that) but `asdf.treeutil.walk_and_modify` is itself part of the public API so users can call it directly to process the tree and remove nodes.  That seems handy, but maybe we should leave out `RemoveNode` until someone actually needs it?

perrygreenfield: Yeah, we may want to see who clamors for it first.

eslavich: Funny coincidence, I just needed this feature (though my task was a bit unusual).  I'm trying to compare schemas for validator content, so I needed to walk the schema tree and drop any of the annotation properties like description, title, etc.

perrygreenfield: OK

jdavies-st:
I ran the `jwst 0.17.0` regression tests using this branch of `asdf`.  Everything passes.  👍

eslavich:
> I ran the `jwst 0.17.0` regression tests using this branch of `asdf`. Everything passes. 👍

Good news, thanks for doing that!  There are two jwst unit test failures that I haven't investigated yet.  Hopefully that's just an issue with how the tests are set up.

jdavies-st:
I don't see any unit test failures with `jwst`.  Which ones?

eslavich:
> Which ones did you see?

These two:

```
FAILED jwst/resample/tests/test_resample_spec.py::test_spatial_transform_nirspec - IndexError: list index out of range
FAILED jwst/stpipe/tests/test_step.py::test_saving_pars - assert <jwst.datamod...x7fe2e18e2450> == <jwst.datamod...x7fe2e1d83e10>
```

jdavies-st:
Ah, I see what happened.  Your branch doesn't have the latest `asdf` tags so that it reinstalled `asdf 2.7.1` over this branch version.  That's why I didn't see those failures (and why the regtests didn't see them either).  Could you push the latest 2.7.1 tag to your fork?

eslavich:
I pushed the missing tags to my fork.

Cadair:
I doubt this will have much impact on us :smile:

jdavies-st:
So this does break a bunch of code in `jwst` where we assume that `None` will not get written.  Some of it means it just gets written, and others means `None` does not validate against the schema we have for a particular element of the tree.  So it would be good to investigate those issues in conjunction with PR.

jdavies-st:
Another side effect of this change is that for existing tags (such as in `astropy.modeling`), any attribute of a model that is allowed to be set to `None` or has a default value of `None`, but whose schema doesn't allow `None` will see validation errors on serialization.  An example:

```python
In [1]: from astropy.modeling.models import Tabular1D

In [2]: import numpy as np

In [3]: points = np.arange(5)

In [4]: points
Out[4]: array([0, 1, 2, 3, 4])

In [5]: tabular_lookup = Tabular1D(points=points, lookup_table=points, fill_value=None)

In [6]: import asdf

In [7]: with asdf.AsdfFile(dict(tabular_lookup=tabular_lookup)) as af:
   ...:     af.write_to("test.asdf")
   ...: 
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
<ipython-input-7-6ebfec37604c> in <module>
----> 1 with asdf.AsdfFile(dict(tabular_lookup=tabular_lookup)) as af:
      2     af.write_to("test.asdf")
      3 

~/miniconda3/envs/null/lib/python3.8/site-packages/asdf/asdf.py in __init__(self, tree, uri, extensions, version, ignore_version_mismatch, ignore_unrecognized_tag, ignore_implicit_conversion, copy_arrays, lazy_load, custom_schema, _readonly)
    167             self.find_references()
    168         else:
--> 169             self.tree = tree
    170             self.find_references()
    171         if uri is not None:

~/miniconda3/envs/null/lib/python3.8/site-packages/asdf/asdf.py in tree(self, tree)
    576         asdf_object = AsdfObject(tree)
    577         # Only perform custom validation if the tree is not empty
--> 578         self._validate(asdf_object, custom=bool(tree))
    579         self._tree = asdf_object
    580 

~/miniconda3/envs/null/lib/python3.8/site-packages/asdf/asdf.py in _validate(self, tree, custom, reading)
    606             tagged_tree = yamlutil.custom_tree_to_tagged_tree(
    607                 tree, self)
--> 608         schema.validate(tagged_tree, self, reading=reading)
    609         # Perform secondary validation pass if requested
    610         if custom and self._custom_schema:

~/miniconda3/envs/null/lib/python3.8/site-packages/asdf/schema.py in validate(instance, ctx, schema, validators, reading, *args, **kwargs)
    655     validator = get_validator(schema, ctx, validators, ctx.resolver,
    656                               *args, **kwargs)
--> 657     validator.validate(instance, _schema=(schema or None))
    658 
    659     additional_validators = [_validate_large_literals]

~/miniconda3/envs/null/lib/python3.8/site-packages/jsonschema/validators.py in validate(self, *args, **kwargs)
    351         def validate(self, *args, **kwargs):
    352             for error in self.iter_errors(*args, **kwargs):
--> 353                 raise error
    354 
    355         def is_type(self, instance, type):

ValidationError: None is not of type 'string'

Failed validating 'type' in schema['allOf'][0]['properties']['name']:
    {'description': 'A user-friendly name for the transform, to give it '
                    'extra\n'
                    'meaning.\n',
     'type': 'string'}

On instance['name']:
    None
```

Possible solutions:
- existing models could call a pre-serialization hook that strips `None` values from the tree.
- existing schemas for attributes that can be `None` have NULL added to the possible `type`.

jdavies-st:
Doh!  The above issue is fixed by

https://github.com/astropy/astropy/pull/10674

So nevermind.  That PR just needs to be merged.

eslavich:
The jwst package has been updated to handle this change in https://github.com/spacetelescope/jwst/pull/5371, so here goes nothin...

PR #866, opened 2020-08-13
Add ASDF Standard 1.6.0 and implement restriction of mapping key types

eslavich:
This updates the asdf-standard submodule to include the latest draft of the ASDF Standard 1.6.0 version map, and implements the restriction on mapping key types that will be part of that version.  Starting with 1.6.0, mapping keys will be limited to string, integer, or boolean type.

### DIFF
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -5,6 +5,7 @@ import copy
 from numbers import Integral
 from functools import lru_cache
 from collections import OrderedDict
+from collections.abc import Mapping
 
 from jsonschema import validators as mvalidators
 from jsonschema.exceptions import ValidationError
@@ -20,6 +21,8 @@ from . import treeutil
 from . import util
 from . import extension
 from . import yamlutil
+from . import versioning
+from . import tagged
 from .exceptions import AsdfDeprecationWarning, AsdfWarning
 
 from .util import patched_urllib_parse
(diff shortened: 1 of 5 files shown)


asdf/schema.py
```diff
@@ -547,30 +550,59 @@ def get_validator(schema={}, ctx=None, validators=None, url_mapping=None,
...
 
-        warnings.warn(
-            "Invalid integer literal value {0} detected while reading file. "
-            "The value has been read safely, but the file should be "
-            "fixed.".format(instance),
-            AsdfWarning
-        )
+    for key in instance:
+        if isinstance(key, tagged.Tagged) or not isinstance(key, (str, int, bool)):
+            if reading:
```
perrygreenfield, line 594: If it accepts the bad key, what ends up in the python representation?  If it is a key that is not acceptable in Python, will that just trigger a different exception?

eslavich: If it's a float or a tagged string, then the value will end up in the tree.  If it's something unhashable, then it'll fail when pyyaml attempts to construct the tagged dict.  Maybe we should just fail on read as well?  I copied the pattern we had for large integers, but maybe it doesn't make sense to accept the key with a warning.

perrygreenfield: There is the warning, but on the other hand I worry that many users won't see that in front of the traceback and know that is the root cause. Some options include more explanation that this is likely to cause an exception anyway (meh), or try to hash the key, and if not just throw the exception there. (is the second easy?)

eslavich: Unfortunately the pyyaml failure occurs before this method is even called, so the user won't get a chance to see the warning in the case of an unhashable key.  If the user does get to the warning then I think there isn't likely to be an exception.

perrygreenfield: Well, in that case, there isn't much point in changing it. I wonder if there is a way to intercept it earlier, but it isn't a high priority. Should an issue be created? (no if it is near impossible).

eslavich: We already have our claws sunk firmly into pyyaml via our custom loader, so it wouldn't be much extra effort to catch this case in there.  I'll create an issue.

eslavich: https://github.com/asdf-format/asdf/issues/868

PR #873, opened 2020-09-04
Asdftool edit

kmacdonald-stsci:
I have added an 'edit' subcommand to the asdftool.  It separates the ASDF markings and YAML text into a separate file to be edited, using
```
asdftool -e -f <ASDF file> -o <YAML file name>
```
Once editing the YAML is complete, use 
```
asdftool -s -f <edited YAML file> -o <ASDF to save YAML to>
```
There may be a gotcha when saving the edited YAML back to ASDF file.  If the edited YAML contains for characters than the number of characters before the binary blocks in the ASDF file, an error message is returned.

### DIFF
--- a/asdf/commands/__init__.py
+++ b/asdf/commands/__init__.py
@@ -10,7 +10,7 @@ from .info import info
 from .edit import edit
 
 
-__all__ = [ 'implode', 'explode', 'to_yaml', 'defragment', 'diff', 'list_tags',
+__all__ = ['implode', 'explode', 'to_yaml', 'defragment', 'diff', 'list_tags',
     'find_extensions', 'info', 'edit']
 
 
--- a/asdf/commands/__init__.py
+++ b/asdf/commands/__init__.py
@@ -7,10 +7,11 @@ from .diff import diff
 from .tags import list_tags
 from .extension import find_extensions
 from .info import info
+from .edit import edit
 
 
-__all__ = ['implode', 'explode', 'to_yaml', 'defragment', 'diff', 'list_tags',
-    'find_extensions', 'info']
+__all__ = [ 'implode', 'explode', 'to_yaml', 'defragment', 'diff', 'list_tags',
+    'find_extensions', 'info', 'edit']
 
 
 # Extracting ASDF-in-FITS files requires Astropy
--- a/asdf/commands/edit.py
+++ b/asdf/commands/edit.py
@@ -27,9 +27,9 @@ class Edit(Command):
         """
         desc_string = (
             "Allows for easy editing of the YAML in an ASDF file.  "
-            "For edit mode, the YAML portion of an ASDF file is"
-            "separated from the ASDF into a text file for easy"
-            "editing.  For save mode, the edited text file is written"
+            "For edit mode, the YAML portion of an ASDF file is "
+            "separated from the ASDF into a text file for easy "
+            "editing.  For save mode, the edited text file is written "
             "to its ASDF file."
         )
 
@@ -47,7 +47,7 @@ class Edit(Command):
             type=str,
             required=True,
             dest="fname",
-            help="Input file (ASDF for -e option, YAML for -s option",
+            help="Input file (ASDF for -e option, YAML for -s option)",
         )
 
         # Need an output file
@@ -57,7 +57,7 @@ class Edit(Command):
             type=str,
             required=True,
             dest="oname",
-            help="Output file (YAML for -e option, ASDF for -s option",
+            help="Output file (YAML for -e option, ASDF for -s option)",
         )
 
         """
@@ -130,7 +130,8 @@ def is_valid_path_and_ext(fname, wanted_ext=None):
     ----------
     fname : str
         Input file name.
-    wanted_ext : List of extensions to check.
+    wanted_ext : List of str, optional
+        Extensions to check
 
     Return
     ------
@@ -203,7 +204,8 @@ def check_asdf_header(fd):
 
     Return
     ------
-    The ASDF header line and the ASDF comment as bytes.
+    bytes
+        The ASDF header line and the ASDF comment.
     """
 
     header_line = fd.read_until(b"\r?\n", 2, "newline", include=True)
@@ -234,7 +236,10 @@ def open_and_check_asdf_header(fname):
 
     Return
     ------
-    File descriptor for ASDF file and the ASDF header and ASDF comments as bytes.
+    GenericFile
+        File descriptor for ASDF file.
+    bytes
+        ASDF header and ASDF comments.
     """
     fullpath = os.path.abspath(fname)
     fd = generic_io.get_file(fullpath, mode="r")
@@ -253,11 +258,13 @@ def read_and_validate_yaml(fd, fname, validate_yaml):
     ----------
     fname : str
         Input file name
-    fd : GenericFile for fname.
+    fd : GenericFile 
+        for fname.
 
     Return
     ------
-    The YAML portion of an ASDF file as bytes.
+    bytes
+        The YAML portion of an ASDF file.
     """
     YAML_TOKEN = b"%YAML"
     token = fd.read(len(YAML_TOKEN))
@@ -283,7 +290,7 @@ def read_and_validate_yaml(fd, fname, validate_yaml):
             print("Error: 'yamlutil.load_tree' failed to return a tree.")
             sys.exist(1)
 
-        schema.validate(tree, None)  # Failure raises an exception.
+        schema.validate(tree)  # Failure raises an exception.
 
     return yaml_content
 
@@ -361,7 +368,6 @@ def find_first_block(fname):
     """
     Finds the location of the first binary block in an ASDF file.
 
-
     Parameters
     ----------
     fname : str
@@ -369,7 +375,8 @@ def find_first_block(fname):
 
     Return
     ------
-    Location, in bytes, of the first binary block.
+    int
+        Location, in bytes, of the first binary block.
     """
     with generic_io.get_file(fname, mode="r") as fd:
         reader = fd.reader_until(
@@ -381,22 +388,85 @@ def find_first_block(fname):
         content_to_first_block = reader.read()
     return len(content_to_first_block) 
 
-def write_edited_yaml_larger (oname, edited_yaml, first_block_loc):
-    print("Larger")
+def copy_binary_blocks(ofd, ifd):
+    """
+    Copies the binary blocks from the input ASDF to the output ASDF.
+
+    Parameters
+    ----------
+    ofd: file descriptor
+        Output ASDF file.
+    ifd: file descriptor
+        Input ASDF file.
+    """
+    block_index = [] # A new block index needs to be computed.
+
+    while True:
+        token_length = ifd.read(6)
+        if not token_length.startswith(constants.BLOCK_MAGIC):
+            ifd.seek(-6,os.SEEK_CUR)
+            break
 
 
-def write_edited_yaml(oname, edited_yaml, first_block_loc):
+def write_edited_yaml_larger(fname, oname, edited_yaml, first_block_loc):
+    """
+    The edited YAML is too large to simply overwrite the exiting YAML in an 
+    ASDF file, so the ASDF file needs to be rewritten.
+
+    Parameters
+    ----------
+    oname : str
+        Input ASDF file name.
+    edited_yaml : byte string
+        The edited YAML to be saved to an ASDF file
+    first_block_location : the location in the ASDF file for the first binary
+                           block
+    """
+    tmp_oname = oname + "tmp"
+
+    ifd = open(oname,"rb")
+    ifd.seek(first_block_loc)
+
+    ofd = open(tmp_oname,"wb")
+    ofd.write(edited_yaml)
+
+    pad_length = 10000  
+    padding = b'\0' * pad_length
+    ofd.write(padding)
+
+    copy_binary_blocks(ofd, ifd)
+
+    ifd.close()
+    ofd.close()
+    # os.replace(tmp_oname,oname)
+
+
+def write_edited_yaml(fname, oname, edited_yaml, first_block_loc):
+    """
+    Write the edited YAML is to an existing ASDF file.
+
+    Parameters
+    ----------
+    oname : str
+        Input ASDF file name.
+    edited_yaml : byte string
+        The edited YAML to be saved to an ASDF file
+    first_block_location : the location in the ASDF file for the first binary
+                           block
+    """
     if len(edited_yaml) < first_block_loc:
+        # The YAML in the ASDF can simply be overwritten
         pad_length = first_block_loc - len(edited_yaml)
         padding = b'\0' * pad_length
         with open(oname, "r+b") as fd:
             fd.write(edited_yaml)
             fd.write(padding)
     elif len(edited_yaml) == first_block_loc:
+        # The YAML in the ASDF can simply be overwritten
         with open(oname, "r+b") as fd:
             fd.write(edited_yaml)
     else:
-        write_edited_yaml_larger(oname, edited_yaml, first_block_loc)  
+        write_edited_yaml_larger(fname, oname, edited_yaml, first_block_loc)  
 
 
 def save_func(fname, oname):
@@ -430,23 +500,8 @@ def save_func(fname, oname):
     ifd.close()
     edited_text = iasdf_text + iyaml_text
 
-    # - Find location of first block.
     loc = find_first_block(oname)
-    write_edited_yaml(oname, edited_text, loc)
-    # - Using edited_text length and first block location determine.
-    #   space availability for new YAML.
-    # - Smaller:
-    #   - Overwrite YAML in ASDF, then pad 0x00 to first b'\xd3BLK'.
-    # - Equal:
-    #   - Overwrite YAML in ASDF.  No padding.
-    # - Larger:
-    #   - Create new file.
-    #   - Write new YAML.
-    #   - Add 10,000 0x00 pad bytes between '...\r?\n' and '\xd3BLK'.
-    #   - If no streaming block found, write block index.
-    #   - If streaming block found, handle correctly.
-
-    return
+    write_edited_yaml(fname, oname, edited_text, loc)
 
 
 def edit(args):
--- a/asdf/commands/edit.py
+++ b/asdf/commands/edit.py
@@ -16,7 +16,8 @@ from .. import yamlutil
 
 from .main import Command
 
-__all__ = ['edit']
+__all__ = ["edit"]
+
 
 class Edit(Command):
     @classmethod
@@ -24,48 +25,58 @@ class Edit(Command):
         """
         Set up a command line argument parser for the edit subcommand.
         """
-        desc_string = "Allows for easy editing of the YAML in an ASDF file.  " \
-                      "For edit mode, the YAML portion of an ASDF file is"  \
-                      "separated from the ASDF into a text file for easy" \
-                      "editing.  For save mode, the edited text file is written" \
-                      "to its ASDF file."
+        desc_string = (
+            "Allows for easy editing of the YAML in an ASDF file.  "
+            "For edit mode, the YAML portion of an ASDF file is"
+            "separated from the ASDF into a text file for easy"
+            "editing.  For save mode, the edited text file is written"
+            "to its ASDF file."
+        )
 
         # Set up the parser
         parser = subparsers.add_parser(
             str("edit"),
             help="Edit YAML portion of an ASDF file.",
-            description=desc_string)
+            description=desc_string,
+        )
 
         # Need an input file
         parser.add_argument(
-            '--infile', '-f',
+            "--infile",
+            "-f",
             type=str,
             required=True,
-            dest='fname',
-            help="Input file (ASDF for -e option, YAML for -s option")
+            dest="fname",
+            help="Input file (ASDF for -e option, YAML for -s option",
+        )
 
         # Need an output file
         parser.add_argument(
-            '--outfile', '-o',
+            "--outfile",
+            "-o",
             type=str,
             required=True,
-            dest='oname',
-            help="Output file (YAML for -e option, ASDF for -s option")
+            dest="oname",
+            help="Output file (YAML for -e option, ASDF for -s option",
+        )
 
         # The edit is either being performed or saved
         group = parser.add_mutually_exclusive_group(required=True)
 
         group.add_argument(
-            '-s',
-            action='store_true',
-            dest='save',
-            help="Saves a YAML text file to its ASDF file.  Requires an YAML input file and ASDF output file.")
+            "-s",
+            action="store_true",
+            dest="save",
+            help="Saves a YAML text file to an ASDF file.  Requires a "
+                 "YAML input file and ASDF output file.",
+        )
 
         group.add_argument(
-            '-e',
-            action='store_true',
-            dest='edit',
-            help="Create a YAML text file for a ASDF file.  Requires a ASDF input file.")
+            "-e",
+            action="store_true",
+            dest="edit",
+            help="Create a YAML text file for a ASDF file.  Requires a ASDF input file.",
+        )
 
         parser.set_defaults(func=cls.run)
 
@@ -89,7 +100,7 @@ def is_yaml_file(fname):
     """
 
     base, ext = os.path.splitext(fname)
-    if '.yaml' != ext:
+    if ".yaml" != ext:
         return False
     return True
 
@@ -127,7 +138,7 @@ def is_valid_asdf_path(fname):
     ----------
     fname : The character string of the input file name.
     """
-    ext = ['.asdf']
+    ext = [".asdf"]
     if is_valid_path_and_ext(fname, ext):
         return True
     print(f"Error: '{fname}' should have extension '{ext[0]}'")
@@ -142,7 +153,7 @@ def is_valid_yaml_path(fname):
     ----------
     fname : The character string of the input file name.
     """
-    ext = ['.yaml']
+    ext = [".yaml"]
     if is_valid_path_and_ext(fname, ext):
         return True
     print(f"Error: '{fname}' should have extension '{ext[0]}'")
@@ -163,16 +174,18 @@ def check_asdf_header(fd):
     The ASDF header line and the ASDF comment as bytes.
     """
 
-    header_line = fd.read_until(b'\r?\n', 2, "newline", include=True)
+    header_line = fd.read_until(b"\r?\n", 2, "newline", include=True)
     if not header_line.startswith(constants.ASDF_MAGIC):
         print("Invalid ASDF ID")
         sys.exit(1)
 
-    comment_section = fd.read_until( b'(%YAML)|(' + constants.BLOCK_MAGIC + b')',
-                                     5,
-                                     "start of content",
-                                     include=False,
-                                     exception=False)
+    comment_section = fd.read_until(
+        b"(%YAML)|(" + constants.BLOCK_MAGIC + b")",
+        5,
+        "start of content",
+        include=False,
+        exception=False,
+    )
 
     return header_line + comment_section
 
@@ -196,7 +209,7 @@ def open_and_check_asdf_header(fname):
     # Read the ASDF header and optional comments section
     header_and_comment = check_asdf_header(fd)
 
-    return fd, header_and_comment   # Return GenericFile and ASDF header bytes.
+    return fd, header_and_comment  # Return GenericFile and ASDF header bytes.
 
 
 def read_and_validate_yaml(fd, fname):
@@ -212,18 +225,20 @@ def read_and_validate_yaml(fd, fname):
     ------
     The YAML portion of an ASDF file as bytes.
     """
-    YAML_TOKEN = b'%YAML'
+    YAML_TOKEN = b"%YAML"
     token = fd.read(len(YAML_TOKEN))
     if token != YAML_TOKEN:
         print(f"Error: No YAML in '{fname}'")
         sys.exit(0)
 
     # Get YAML reader and content
-    reader = fd.reader_until(constants.YAML_END_MARKER_REGEX,
-                             7,
-                             'End of YAML marker',
-                             include=True,
-                             initial_content=token)
+    reader = fd.reader_until(
+        constants.YAML_END_MARKER_REGEX,
+        7,
+        "End of YAML marker",
+        include=True,
+        initial_content=token,
+    )
     yaml_content = reader.read()
 
     # Create a YAML tree to validate
@@ -233,9 +248,10 @@ def read_and_validate_yaml(fd, fname):
         print("Error: 'yamlutil.load_tree' failed to return a tree.")
         sys.exist(1)
 
-    schema.validate(tree, None)     # Failure raises an exception.
+    schema.validate(tree, None)  # Failure raises an exception.
 
     return yaml_content
+
 
 def edit_func(fname, oname):
     """
@@ -259,6 +275,7 @@ def edit_func(fname, oname):
 
     # Open a YAML file for the ASDF YAML.
     if not is_yaml_file(oname):
+        print("A YAML file is expected, with '.yaml' extension.")
         sys.exit(1)
 
     # Write the YAML for the original ASDF file.
@@ -267,7 +284,7 @@ def edit_func(fname, oname):
         ofd.write(yaml_text)
 
     # Output message to user.
-    delim = '*' * 70
+    delim = "*" * 70
     print(f"\n{delim}")
     print("ASDF formatting and YAML schema validated.")
     print(f"The text portion of '{fname}' is written to:")
@@ -275,15 +292,12 @@ def edit_func(fname, oname):
     print(f"The file '{oname}' can be edited using your favorite text editor.")
     print("The edited text can then be saved to the ASDF file of your choice")
     print("using 'asdftool edit -s -f <edited text file> -o <ASDF file>.")
-    print('-' * 70)
-    print("Note: This is meant to be a lightweight text editing tool of")
-    print("      ASDF .If the edited text is larger than the YAML portion")
-    print("      of the ASDF file to be written to, the edits may not be")
-    print("      able to saved.")
     print(f"{delim}\n")
 
     return
 
+
+# TODO conosidate the next two functions and change 'buffer' to 'pad'.
 def buffer_edited_text(edited_text, orig_text):
     """
     There is more text in the original ASDF file than in the edited text,
@@ -299,15 +313,12 @@ def buffer_edited_text(edited_text, orig_text):
     The buffered text and the number of spaces added as buffer.
     """
     diff = len(orig_text) - len(edited_text)
-    if diff < 1:
-        print("Error: shouldn't be here.")
-        sys.exit(1)
 
-    wdelim = b'\r\n...\r\n'
-    ldelim = b'\n...\n'
-    if edited_text[-len(wdelim) :]==wdelim:
+    wdelim = b"\r\n...\r\n"
+    ldelim = b"\n...\n"
+    if edited_text[-len(wdelim) :] == wdelim:
         delim = wdelim
-    elif edited_text[-len(ldelim) :]==ldelim:
+    elif edited_text[-len(ldelim) :] == ldelim:
         delim = ldelim
     else:
         print("Unrecognized YAML delimiter ending the YAML text.")
@@ -315,8 +326,7 @@ def buffer_edited_text(edited_text, orig_text):
         print(f"last {len(wdelim)} bytes are {edited_text[-len(wdelim):]}.")
         sys.exit(1)
 
-    # May not be correct.  If on Windows use '\r\n'.
-    buffered_text = edited_text[: -len(delim)] + b'\n' + b' '*(diff - 1) + delim
+    buffered_text = edited_text[: -len(delim)] + b"\n" + b"\0" * (diff - 1) + delim
     return buffered_text, diff - 1
 
 
@@ -333,11 +343,11 @@ def add_buffer_to_new_text(edited_text, buffer_size):
     ------
     Buffered text with the number of spaces requested as buffer.
     """
-    wdelim = b'\r\n...\r\n'
-    ldelim = b'\n...\n'
-    if edited_text[-len(wdelim) :]==wdelim:
+    wdelim = b"\r\n...\r\n"
+    ldelim = b"\n...\n"
+    if edited_text[-len(wdelim) :] == wdelim:
         delim = wdelim
-    elif edited_text[-len(ldelim) :]==ldelim:
+    elif edited_text[-len(ldelim) :] == ldelim:
         delim = ldelim
     else:
         print("Unrecognized YAML delimiter ending the YAML text.")
@@ -345,8 +355,8 @@ def add_buffer_to_new_text(edited_text, buffer_size):
         print(f"last {len(wdelim)} bytes are {edited_text[-len(wdelim):]}.")
         sys.exit(1)
 
-    buf = b' ' * buffer_size
-    buffered_text = edited_text[: -len(delim)] + b'\n' + buf + delim
+    buf = b" " * buffer_size
+    buffered_text = edited_text[: -len(delim)] + b"\n" + buf + delim
 
     return buffered_text
 
@@ -363,14 +373,16 @@ def write_block_index(fd, index):
     if len(index) < 1:
         return
 
+    # TODO - this needs to be changed to use constants.py and pyyaml
     bindex_hdr = b"#ASDF BLOCK INDEX\n%YAML 1.1\n---\n"
     fd.write(bindex_hdr)
     for idx in index:
-        ostr = f'- {idx}\n'
-        fd.write(ostr.encode('utf-8'))
-    end = b'...'
+        ostr = f"- {idx}\n"
+        fd.write(ostr.encode("utf-8"))
+    end = b"..."
     fd.write(end)
     return
+
 
 def get_next_block_header(fd):
     """
@@ -403,6 +415,7 @@ def get_next_block_header(fd):
     header = fd.read(hsz)
     return blk_header + header
 
+
 def rewrite_asdf_file(edited_text, orig_text, oname, fname):
     """
     Rewrite an ASDF file for too large edited YAML.  The edited YAML, a buffer,
@@ -419,19 +432,19 @@ def rewrite_asdf_file(edited_text, orig_text, oname, fname):
     fname : the edit YAML to write to new file.
     """
 
-    tmp_oname = oname + '.tmp'      # Save as a temp file, in case anything goes wrong.
+    tmp_oname = oname + ".tmp"  # Save as a temp file, in case anything goes wrong.
     buffer_size = 10 * 1000
     buffered_text = add_buffer_to_new_text(edited_text, buffer_size)
 
-    ifd = open(oname, "r+b")         # Open old ASDF to get binary blocks
+    ifd = open(oname, "r+b")  # Open old ASDF to get binary blocks
     ifd.seek(len(orig_text))
 
-    ofd = open(tmp_oname, "w+b")     # Open temp file to write
-    ofd.write(buffered_text)        # Write edited YAML
+    ofd = open(tmp_oname, "w+b")  # Open temp file to write
+    ofd.write(buffered_text)  # Write edited YAML
 
     current_location = len(buffered_text)
     block_index = []
-    alloc_loc = 14      # 4 bytes of block ID, 2 blocks of size, 8 blocks into header
+    alloc_loc = 14  # 4 bytes of block ID, 2 blocks of size, 8 blocks into header
     block_chunk = 2048
     while True:
         next_block = get_next_block_header(ifd)
@@ -439,7 +452,7 @@ def rewrite_asdf_file(edited_text, orig_text, oname, fname):
             break
 
         # Get block size on disk
-        alloc = struct.unpack(">Q", next_block[alloc_loc:alloc_loc+8])[0]
+        alloc = struct.unpack(">Q", next_block[alloc_loc : alloc_loc + 8])[0]
 
         # Save block location for block index
         block_index.append(current_location)
@@ -451,10 +464,9 @@ def rewrite_asdf_file(edited_text, orig_text, oname, fname):
             chunk = ifd.read(block_chunk)
             ofd.write(chunk)
             alloc -= block_chunk
-        if alloc>0:
+        if alloc > 0:
             chunk = ifd.read(alloc)
             ofd.write(chunk)
-
 
     write_block_index(ofd, block_index)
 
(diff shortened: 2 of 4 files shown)


asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    # Output message to user.
+    delim = '*' * 70
+    print(f"\n{delim}")
+    print(f"The text in '{fname}' was too large to simply overwrite the")
+    print(f"text in '{oname}'.  The file '{oname}' was rewritten to")
+    print(f"accommodate the larger text size.  Also, {buffer_size:,} bytes")
+    print(f"as a buffer for the text in '{oname}' to allow for future edits.")
+    print(f"{delim}\n")
+
+def save_func ( fname, oname ) :
```
perrygreenfield, line 438: I'm not sure the large file bailout makes any sense. Someone has just spent time editing a header and then is told, sorry I can't use it. I'm going to guess that in almost all circumstances they have spent more time editing then it will take to rewrite the file. If there isn't enough disk space left, that's one issue, but otherwise, just do it (IMHO)

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    fname : The input file name.
+    """
+    ext = ['.yaml']
+    if is_validate_path_and_ext(fname,ext) : 
+        return True
+    print(f"Error: '{fname}' should have extension '{ext[0]}'")
+    return False
+
+
+def validate_asdf_file ( fd ) :
```
perrygreenfield, line 167: The name validate here is going to be somewhat confusing since validation for ASDF files has a specific meaning elsewhere. check_asdf_file? Perhaps someone else will have a better name suggestion.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    ----------
+    fname : The input file name.
+    '''
+
+    base, ext = os.path.splitext(fname)
+    if '.asdf' != ext :
+        return False
+
+    with open(fname,"r+b") as fd :
+        first_string = "#ASDF"
```
perrygreenfield, line 107: I may be a bit too picky on this, but I see a number of cases where there is variable defined from a constant, then that constant is used in a test. Unless that constant is subject to change (in which case it belongs with other such constants), it would just be simpler to put that string in the test. At least in my opinion. Others may not agree with me :-)

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+                                     "start of content", 
+                                     include=False, 
+                                     exception=False)
+    # Maybe do the following for more validate.  But maybe not.
+    #comments = _parse_asdf_comment_section(comment_section)
+    #asdf_standard_version = _get_asdf_version_in_comments(comments)
+
+    return header_line + comment_section 
+    
+def open_and_validate_asdf ( fname ) :
```
perrygreenfield, line 198: likewise

perrygreenfield (commented):
Finally, I have a vague sense that some of this code is similar to what is in AsdfFile and there is going to be unnecessary duplication. On the other hand, what is in AsdfFile doesn't quite meet the needs of this and I think it would need a little reworking to be useful  for this context. I wouldn't expect that you would have to do that reworking. Perhaps some future refactoring?

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    # 2 bytes of header length, after the length field (min 48)
+    # 4 bytes flag
+    # 4 bytes compression
+    # 8 bytes allocated size
+    # 8 bytes used (on disk) size
+    # 8 bytes data size
+    # 16 bytes checksum
+    bmlen = len(constants.BLOCK_MAGIC)
+    min_header = bmlen + 2 + 48  
+    uidx = 22
```
eslavich, line 363: I think we need to use the allocated size here -- the next block will begin after this block's padding.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+
+    # May not be correct.  If on Windows use '\r\n'.
+    buffered_text = edited_text[:-len(delim)] + b'\n' + b' '*(diff-1) + delim
+    return buffered_text, diff-1 
+
+
+def add_buffer_to_new_text ( edited_text, buffer_size ) :
+    """ Adds buffer to edited text.
+    """
+    wdelim = b'\r\n...\r\n'
```
eslavich, line 328: Can we have the user edit the YAML content between the delimiters, then add them back automatically?

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    # 8 bytes used (on disk) size
+    # 8 bytes data size
+    # 16 bytes checksum
+    bmlen = len(constants.BLOCK_MAGIC)
+    min_header = bmlen + 2 + 48  
+    uidx = 22
+    bindex = [start]
+    k = 0
+    while len(asdf_blocks) - k > min_header :
+        hsz = struct.unpack(">H",asdf_blocks[k+bmlen:k+bmlen+2])[0]
```
eslavich, line 367: There is a [header struct-like thing in the Block class](https://github.com/asdf-format/asdf/blob/d0a1c0a055233d7b91b05e8e8930cb5fe8f804c2/asdf/block.py#L782) that we can reuse here.

asdf/commands/edit.py
eslavich, line 167: This function on the whole seems like it can be shared with asdf.py.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    
+def read_and_validate_yaml ( fd, fname ) :
+    """ Get the YAML text from an ASDF formatted file.
+
+    Parameters
+    ----------
+    fname : The input file name.
+    fd : GenericFile for fname.
+    """
+    YAML_TOKEN = b'%YAML'
```
eslavich, line 222: I mentioned this down below, but I think we should omit this from the temp file so that the user has fewer options for making a mistake.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+        delim = ldelim
+    else:
+        # Mabye raise exception
+        print("Unrecognized YAML delimiter ending the YAML text.")
+        print(f"It should be {wdelim} or {ldelim}, but the")
+        print(f"last {len(wdelim)} bytes are {edited_text[-len(wdelim):]}.")
+        sys.exit(1)
+
+    # May not be correct.  If on Windows use '\r\n'.
+    buffered_text = edited_text[:-len(delim)] + b'\n' + b' '*(diff-1) + delim
```
eslavich, line 321: Hmm, the ASDF Standard suggests spaces, but if I understand correctly, asdf.py pads with `0x00`.  Maybe we ought to use `0x00` here for consistency and file an issue to fix both of them in the future?

eslavich (commented):
This is a good start, in general I think we should take more opportunities for code sharing between edit.py and asdf.py/block.py.

Also, just a heads up, we're going to need to figure out some pytest tests for this, but that should wait until we're all agreed on the implementation.

asdf/commands/edit.py
eslavich: I mention this down below, but in this case there is a constant defined in constants.py that we can use.

asdf/commands/edit.py
eslavich: I agree with this.  Since available memory may well be an issue, we should copy the bytes from old to new in batches.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    index, asdf_blocks = compute_block_index_blocks(len(buffered_text),asdf_blocks)
+    out_bytes = buffered_text + asdf_blocks
+
+    # Write new file with edited text, buffer, and recomputed block index.
+    with open(tmp_oname,"w+b") as fd :
+        fd.write(out_bytes)
+        write_block_index(fd,index)
+
+    # Rename temp file.
+    os.rename(tmp_oname,oname)
```
eslavich, line 427: Good call!

asdf/commands/edit.py
kmacdonald-stsci: In the description of the tree at https://asdf-standard.readthedocs.io/en/1.5.0/file_layout.html#tree the buffer byte between the tree and blocks is recommended is 0x20.

asdf/commands/edit.py
eslavich: We discussed this over email and agreed to keep the entire YAML document editable.

asdf/commands/edit.py
eslavich: We discussed this over email and agreed to keep the entire YAML document editable.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    oname = os.path.join(tmpdir, fname)
+    if os.path.exists(oname):
+        os.remove(oname)
+    af = asdf.AsdfFile(tree,version=version)
+    af.write_to(oname)
+
+    return oname
+
+
+def create_edit_equal(base_yaml):
```
eslavich, line 60: These three helpers could be combined into a single function, something like this:

```python
import re

def create_edited_yaml(input, output, pattern, replacement):
    with open(input) as fd:
        content = fd.read()
    new_content = re.sub(pattern, replacement)
    with open(output, "w") as fd:
      fd.write(new_content)
```

Then in the test:

```python
# Edit with equal length
create_edited_yaml(yaml_base, edited_path, "foo: 41", "foo: 42")

# Edit with shorter length
create_edited_yaml(yaml_base, edited_path, "foo: 41", "foo: 2")

# Edit with longer length
create_edited_yaml(yaml_base, edited_path, "foo: 41", "foo: 41\nbar: 13")
```

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    with open(oname, "w") as fd:
+        for l in lines:
+            fd.write(l)
+            if "foo" in l:
+                print("bar: 13", file=fd)  # Add a line
+
+    return oname
+
+
+def copy_base_asdf_equal(base_asdf):
```
eslavich, line 130: Another good opportunity for code reuse here, we can just pass `oname` as a parameter to the function.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    oname = f"{base}_larger.asdf"
+    if os.path.exists(oname):
+        os.remove(oname)
+    shutil.copyfile(base_asdf, oname)
+
+    return oname
+
+
+@pytest.mark.parametrize("version", asdf.versioning.supported_versions)
+def test_edits(tmpdir,version):
```
eslavich, line 173: Missing space between `tmpdir,` and `version`

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    #        Test:
```
eslavich, line 174: Is this line leftover from something?

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    # Create base ASDF file for testing
+    tmpdir = "/Users/kmacdonald/tmp"
```
eslavich, line 176: Looks like debris left over from local testing?

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    asdf_base = create_base_asdf(tmpdir, version)
+
+    # Create base YAML file from base ASDF file
+    base, ext = os.path.splitext(asdf_base)
+    yaml_base = f"{base}.yaml"
+    # Run: asdftool edit -e -f {asdf_base} -o {yaml_base}
+    args = ["edit", "-e", "-f", f"{asdf_base}", "-o", f"{yaml_base}"]
+    main.main_from_args(args)
+
+    # Test smaller
```
eslavich, line 186: I would split each of these cases into separate parametrized tests.  That has the advantage of clearly identifying the failed case when something goes wrong.  As it stands, if the "smaller" case fails, we won't know if the other two would have succeeded since the test halts at the first error.

That structure would also save us from having to copy the "base" ASDF file, since we'd be only using it once per test anyway.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    asdf_smaller = copy_base_asdf_smaller(asdf_base)
+
+    # Create edited YAML file with larger number of characters
+    yaml_smaller = create_edit_smaller(yaml_base)
+
+    # Run: asdftool edit -s -f {yaml_larger} -o {asdf_larger}
+    args = ["edit", "-s", "-f", f"{yaml_smaller}", "-o", f"{asdf_smaller}"]
+    main.main_from_args(args)
+
+    af_smaller = asdf.open(asdf_smaller)
```
eslavich, line 197: Even though this is only a test, it's good to get in the habit of closing the AsdfFile after use, so that the underlying file pointer gets released.  The context manager makes this easy:

```python
with asdf.open(asdf_smaller) as af_smaller:
    assert af_smaller.tree["foo"] == 2
```

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+    Parameters
+    ----------
+    edited_text : the new YAML text to write out.
+    orig_text : the original YAML text to overwrite.
+    oname : the ASDF file to overwrite.
+    fname : the edit YAML to write to new file.
+    """
+
+    tmp_oname = oname + '.tmp'      # Save as a temp file, in case anything goes wrong.
+    buffer_size = 10 * 1000
```
eslavich, line 423: We chatted about switching to the asdf.py style of padding by a percentage of the current size.  I think that's still a good idea, to keep asdf.py and this module as closely aligned as possible.

eslavich (commented):
Left a smattering of comments, but it's shaping up nicely.  There were some issues that I should have noticed before but evidently I wasn't on my A-game.

asdf/commands/__init__.py
```diff
@@ -7,10 +7,11 @@
 from .tags import list_tags
 from .extension import find_extensions
 from .info import info
+from .edit import edit
 
 
-__all__ = ['implode', 'explode', 'to_yaml', 'defragment', 'diff', 'list_tags',
-    'find_extensions', 'info']
+__all__ = [ 'implode', 'explode', 'to_yaml', 'defragment', 'diff', 'list_tags',
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this space.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+        return False
+
+    # Validate input file is an ASDF file.
+    fd, asdf_text = open_and_check_asdf_header(fname)
+
+    # Read and validate the YAML of an ASDF file.
+    yaml_text = read_and_validate_yaml(fd, fname)
+
+    # Open a YAML file for the ASDF YAML.
+    if not is_yaml_file(oname):
```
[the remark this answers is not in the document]
kmacdonald-stsci: I included an error message.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+    delim = '*' * 70
+    print(f"\n{delim}")
+    print("ASDF formatting and YAML schema validated.")
+    print(f"The text portion of '{fname}' is written to:")
+    print(f"    '{oname}'")
+    print(f"The file '{oname}' can be edited using your favorite text editor.")
+    print("The edited text can then be saved to the ASDF file of your choice")
+    print("using 'asdftool edit -s -f <edited text file> -o <ASDF file>.")
+    print('-' * 70)
+    print("Note: This is meant to be a lightweight text editing tool of")
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this note.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+    edited_text - The text from the edited YAML file
+    orig_text - The text from the original ASDF file
+
+    Return
+    ------
+    The buffered text and the number of spaces added as buffer.
+    """
+    diff = len(orig_text) - len(edited_text)
+    if diff < 1:
+        print("Error: shouldn't be here.")
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this check.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+        delim = wdelim
+    elif edited_text[-len(ldelim) :]==ldelim:
+        delim = ldelim
+    else:
+        print("Unrecognized YAML delimiter ending the YAML text.")
+        print(f"It should be {wdelim} or {ldelim}, but the")
+        print(f"last {len(wdelim)} bytes are {edited_text[-len(wdelim):]}.")
+        sys.exit(1)
+
+    # May not be correct.  If on Windows use '\r\n'.
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed comment.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+        print(f"\n{msg_delim}")
+        print(f"The edited text in '{fname}' was written to '{oname}'")
+        print(f"{msg_delim}\n")
+    elif len(edited_text) < len(asdf_text):
+        buffered_text, diff = buffer_edited_text(edited_text, asdf_text)
+        with open(oname, "r+b") as fd:
+            fd.write(buffered_text)
+        print(f"\n{msg_delim}")
+        print(f"The edited text in '{fname}' was written to '{oname}'")
+        print(f"Added a  '\\n' and {diff} buffer of spaces between the YAML text and binary blocks.")
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this space.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+import pytest
+
+import asdf
+from asdf import AsdfFile
+from asdf.commands import main
+from ...tests.helpers import get_file_sizes, assert_tree_match
+
+
+
+"""
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this doc string.  It was for me to think about how to structure the code.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+Three tests are defined.  
+
+1. Run the command 'asdftool edit -e' to create a YAML file, simulating
+   the steps a user would make to start editing an ASDF file.
+2. Run the command 'asdftool edit -s' to save YAML edits such that the
+   edited YAML will have the same or fewer characters as the original
+   ASDF file, so will be overwritten in place with the space character
+   used as any buffer, to consume all the memory on disk the YAML takes
+   up in the ASDF file.
+2. Run the command 'asdftool edit -s' to save YAML edits such that the
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this doc string.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+   edited YAML will have the more characters than the original ASDF file.
+   This triggers a rewrite of the file, since there isn't enough 'room' 
+   on disk to accomadate the edited YAML.  The resultant YAML will be
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this doc string.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+   edited YAML will have the same or fewer characters as the original
+   ASDF file, so will be overwritten in place with the space character
+   used as any buffer, to consume all the memory on disk the YAML takes
+   up in the ASDF file.
+2. Run the command 'asdftool edit -s' to save YAML edits such that the
+   edited YAML will have the more characters than the original ASDF file.
+   This triggers a rewrite of the file, since there isn't enough 'room' 
+   on disk to accomadate the edited YAML.  The resultant YAML will be 
+   rewritten with a buffer (using the space character as buffer) to 
+   accomodate future edits.  If a block index existed in the original ASDF
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this doc string.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+    # Store the data in an arbitrarily nested dictionary
+    tree = {
+        "foo": 42,
+        "name": "Monty",
+        "sequence": seq,
+    }
+
+    fname = "test_edit_base.asdf"
+    oname = os.path.join(tmpdir, fname)
+    if os.path.exists(oname):
```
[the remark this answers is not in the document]
kmacdonald-stsci: This check is removed.

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,231 @@
...
+        os.remove(oname)
+    af = asdf.AsdfFile(tree,version=version)
```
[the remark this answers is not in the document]
kmacdonald-stsci: I ran black on this file to reformat style changes like this.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: I refactored the creation of the edited YAML using the `re` module.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: This has been refactored out of the test.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: I ran `black` on this file to reformat all style issues.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: I removed all such comments.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: This has been removed.  It was for local testing and debugging.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: The test has been refactored creating three separate tests.

asdf/asdf.py
```diff
@@ -6,6 +6,8 @@
 import warnings
 from pkg_resources import parse_version
 
+import ipdb
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this from the code.

asdf/commands/edit.py
kmacdonald-stsci: I changed this to rewrite in this case.

asdf/commands/edit.py
kmacdonald-stsci: I changed "validate" to "check".

asdf/commands/edit.py
kmacdonald-stsci: I changed this to use the string defined in constants.

asdf/commands/edit.py
kmacdonald-stsci: I changed "validate" to "check".

asdf/asdf.py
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this.

asdf/asdf.py
```diff
@@ -35,6 +37,49 @@
 
 from .tags.core import AsdfObject, Software, HistoryEntry, ExtensionMetadata
 
+def _get_asdf_version_in_comments( comments ):
+    """ From the initial comments line in an ASDF file, capture the ASDF version.
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this edit from `asdf.py`.

asdf/asdf.py
```diff
@@ -733,47 +778,22 @@ def _parse_header_line(cls, line):
...
-            raise ValueError("Does not appear to be a ASDF file.")
-
-        try:
-            version = versioning.AsdfVersion(parts[1].decode('ascii'))
-        except ValueError:
-            raise ValueError(
-                "Unparseable version in ASDF file: {0}".format(parts[1]))
-
-        return version
+        return _parse_asdf_header_line(line)
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this edit from `asdf.py`.

asdf/asdf.py
```diff
@@ -35,6 +37,49 @@
...
+                version = versioning.AsdfVersion(parts[1].decode('ascii'))
+            except ValueError:
+                pass
+            else:
+                return version
+
+    return None
+
+def _parse_asdf_comment_section( content ):
+    """ Parses the comment section, between the header line and the
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this edit from `asdf.py`.

asdf/asdf.py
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this edit from `asdf.py`.

asdf/asdf.py
```diff
@@ -35,6 +37,49 @@
...
+            try:
+                version = versioning.AsdfVersion(parts[1].decode('ascii'))
+            except ValueError:
+                pass
+            else:
+                return version
+
+    return None
+
+def _parse_asdf_comment_section( content ):
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this edit from `asdf.py`.

asdf/asdf.py
```diff
@@ -785,12 +805,19 @@ def _open_asdf(cls, self, fd, uri=None, mode='r',
                    ignore_missing_extensions=False,
                    **kwargs):
         """Attempt to populate AsdfFile data from file-like object"""
+        # Generally, I think this function should be made smaller.  There are
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this comment.

asdf/asdf.py
```diff
@@ -35,6 +37,49 @@
 
 from .tags.core import AsdfObject, Software, HistoryEntry, ExtensionMetadata
 
+def _get_asdf_version_in_comments( comments ):
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed my edits from `asdf.py`.

asdf/asdf.py
```diff
@@ -860,25 +892,31 @@ def _open_asdf(cls, self, fd, uri=None, mode='r',
...
             tree = yamlutil.custom_tree_to_tagged_tree(AsdfObject(), self)
 
+        # Function 6 Process binary blocks
         if has_blocks:
             self._blocks.read_internal_blocks(
                 fd, past_magic=True, validate_checksums=validate_checksums)
             self._blocks.read_block_index(fd, self)
 
+        # Function 7 References
+        ipdb.set_trace()
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed my edits from `asdf.py`.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+from .. import generic_io
+from .. import reference
+from .. import schema
+from .. import yamlutil
+
+from .main import Command
+
+__all__ = ['edit']
+
+#asdf_format_version = None
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this comment.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+        # Need an output file
+        parser.add_argument(
+            '--outfile', '-o', type=str, required=True, dest='oname',
+            help="Output file (YAML for -e option, ASDF for -s option")
+
+        # The edit is either being performed or saved
+        group = parser.add_mutually_exclusive_group(required=True)
+        group.add_argument(
+            '-s',action='store_true',dest='save',
+            help="Saves a YAML text file to its ASDF file.  Requires an ASDF input file.")
```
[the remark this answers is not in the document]
kmacdonald-stsci: I changed this help string.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+
+def validate_asdf_file ( fd ) :
+    """ Makes sure the header line is the expected one, as well
+    as getting the optional comment line.
+
+    Parameters
+    ----------
+    fd : GenericFile
+    """
+    #global asdf_format_version
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed them.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+    """ Makes sure the header line is the expected one, as well
+    as getting the optional comment line.
+
+    Parameters
+    ----------
+    fd : GenericFile
+    """
+    #global asdf_format_version 
+    #global asdf_standard_version 
+    ASDF_ID = b'#ASDF'
```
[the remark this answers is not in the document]
kmacdonald-stsci: I changed this to use constants defined in `constants.py`.

asdf/commands/edit.py
kmacdonald-stsci: I changed this to use the allocated size.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+
+    # Rename temp file.
+    os.rename(tmp_oname, oname)
+
+    # Output message to user.
+    delim = '*' * 70
+    print(f"\n{delim}")
+    print(f"The text in '{fname}' was too large to simply overwrite the")
+    print(f"text in '{oname}'.  The file '{oname}' was rewritten to")
+    print(f"accommodate the larger text size.")
```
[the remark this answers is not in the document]
kmacdonald-stsci: I ran `black` to change all formatting, then used `tox` as you defined above, to resolve these errors.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: I changed test checks to this format.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+    print('-' * 70)
+    print("Note: This is meant to be a lightweight text editing tool of")
+    print("      ASDF .If the edited text is larger than the YAML portion")
+    print("      of the ASDF file to be written to, the edits may not be")
+    print("      able to saved.")
+    print(f"{delim}\n")
+
+    return
+
+def buffer_edited_text(edited_text, orig_text):
```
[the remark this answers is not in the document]
kmacdonald-stsci: Changed "buffer" to "pad".

perrygreenfield (approved):
I don't think the request to bypass validation on opening should hold up acceptance.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+        return None
+    if not blk_header.startswith(constants.BLOCK_MAGIC):
+        return None
+    hsz = struct.unpack(">H", blk_header[4:6])[0]
+    header = fd.read(hsz)
+    return blk_header + header
+
+
+def rewrite_asdf_file(edited_text, orig_text, oname, fname):
+    """
```
eslavich, line 446: This docstring is missing the formatting updates.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    Parameters
+    ----------
+    edited_text : the new YAML text to write out.
+    orig_text : the original YAML text to overwrite.
+    oname : the ASDF file to overwrite.
+    fname : the edit YAML to write to new file.
+    """
+
+    tmp_oname = oname + ".tmp"  # Save as a temp file, in case anything goes wrong.
+    pad_size = 10 * 1000
```
eslavich, line 462: Didn't we decide to pad the file by a percentage like asdf.py does?

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    padded_text = add_pad_to_new_text(edited_text, pad_size)
+
+    ifd = open(oname, "r+b")  # Open old ASDF to get binary blocks
```
eslavich, line 465: The `+` opens this file for reading _and_ writing, which I don't think we want here.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    ifd.seek(len(orig_text))
+
+    ofd = open(tmp_oname, "w+b")  # Open temp file to write
+    ofd.write(padded_text)  # Write edited YAML
+
+    current_location = len(padded_text)
+    block_index = []
+    alloc_loc = 14  # 4 bytes of block ID, 2 blocks of size, 8 blocks into header
```
eslavich, line 473: In the case of a streaming block, this field is ignored and all bytes following the header should be included in the final block.  See https://asdf-standard.readthedocs.io/en/1.5.0/file_layout.html#flags

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+        while alloc >= block_chunk:
+            chunk = ifd.read(block_chunk)
+            ofd.write(chunk)
+            alloc -= block_chunk
+        if alloc > 0:
+            chunk = ifd.read(alloc)
+            ofd.write(chunk)
+    ifd.close()
+
+    write_block_index(ofd, block_index)
```
eslavich, line 498: I mentioned this in my previous review, but this is not correct for files that contain a streaming block.  Per the standard the block index should not be written when a file utilizes the streaming block.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    Checks to makes sure fname is a valid YAML file.
+    If the YAML text is smaller than the YAML text in the ASDF file
+        overwrite the YAML in the ASDF file.
+    If the YAML text is smaller than the YAML text in the ASDF file
+        If the file is small, then rewrite file.
+        If the file is large, ask if rewrite is desired.
+
+    Parameters
+    ----------
+    fname : The input YAML file.
```
eslavich, line 528: This docstring is missing the formatting updates.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    Parameters
+    ----------
+    fd - The output file to write the block index.
+    index - A list of locations for each block.
+    """
+    if len(index) < 1:
+        return
+
+    # TODO - this needs to be changed to use constants.py and pyyaml
+    bindex_hdr = b"#ASDF BLOCK INDEX\n%YAML 1.1\n---\n"
```
eslavich, line 403: I mentioned this in my previous two reviews, but we should use pyyaml here and `constants.INDEX_HEADER`.

eslavich (commented):
The streaming block case is a serious problem, and IMO we should still change that padding character.  Can you also look back over the comments from previous rounds of review?  GitHub marks them as "outdated" when the file changes but some of them haven't actually been resolved.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    fname : str
+        Input file name.
+
+    Return
+    ------
+    bool
+    """
+
+    base, ext = os.path.splitext(fname)
+    if ".yaml" != ext:
```
jdavies-st, line 108: Sometimes YAML files also use a `.yml` extension.  Don't know if we will ever come across one of these for this function?

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,97 @@
...
+
+def _create_edited_yaml(base_yaml, edited_yaml, pattern, replacement):
+    with open(base_yaml, "rb") as fd:
+        content = fd.read()
+        new_content = re.sub(pattern, replacement, content)
+        with open(edited_yaml, "wb") as fd:
+            fd.write(new_content)
+
+
+def _initialize_test(tmpdir, version, test_name):
```
jdavies-st, line 37: `test_name` is not needed here, as `tmpdir` always creates a new temporary directory under `basetemp` for each test (and each parameterization), so files will be segregated by directory based on the test name and parameter (version in this case).  Changing the filename is redundant.

jdavies-st (approved):
This looks really good.  Just a couple specific comments below.

And a general comment - it looks like there are a lot of `print()` statements here when handling errors, and this is the case with a bunch of the other `asdftool` subcommands as well.  It might be better to use the `logging` module I think, but perhaps that ship has sailed.  =)

asdf/commands/tests/test_edit.py
```diff
@@ -0,0 +1,97 @@
...
+    asdf_base = os.path.join(tmpdir, f"{test_name}_base.asdf")
+    yaml_base = os.path.join(tmpdir, f"{test_name}_base.yaml")
+    asdf_edit = os.path.join(tmpdir, f"{test_name}_edit.asdf")
+    yaml_edit = os.path.join(tmpdir, f"{test_name}_edit.yaml")
+
+    _create_base_asdf(version, asdf_base)
+    _create_base_asdf(version, asdf_edit)
```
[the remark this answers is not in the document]
kmacdonald-stsci: Yes.  For testing, I did not want to save edits to the original file from which the YAML was taken.  I want to be able to test that the edited file is the same size or larger, depending on the scenario, than the original file.  Originally, I simply copied the asdf_base file, but thought that might cause problems during Windows testing, so created two identical files at the same time.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+            "Allows for easy editing of the YAML in an ASDF file.  "
+            "For edit mode, the YAML portion of an ASDF file is"
+            "separated from the ASDF into a text file for easy"
+            "editing.  For save mode, the edited text file is written"
+            "to its ASDF file."
+        )
+
+        # Set up the parser
+        parser = subparsers.add_parser(
+            str("edit"),
```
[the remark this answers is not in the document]
kmacdonald-stsci: I simply copied the style from other examples, such as `exploded.py`.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+        )
+
+        # Need an input file
+        parser.add_argument(
+            "--infile",
+            "-f",
+            type=str,
+            required=True,
+            dest="fname",
+            help="Input file (ASDF for -e option, YAML for -s option",
```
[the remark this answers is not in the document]
kmacdonald-stsci: The close parenthesis is on line 51.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+        )
+
+        # Need an output file
+        parser.add_argument(
+            "--outfile",
+            "-o",
+            type=str,
+            required=True,
+            dest="oname",
+            help="Output file (YAML for -e option, ASDF for -s option",
```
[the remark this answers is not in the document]
kmacdonald-stsci: The close parenthesis is on line 61.  This formatting is the result of running `black` on the file.

asdf/commands/edit.py
kmacdonald-stsci: I realized that rereading the standard this week.  I will have to generate a test case for it as well.

asdf/commands/edit.py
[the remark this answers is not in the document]
eslavich: I mean, the close parenthesis within the help text string:

```
Input file (ASDF for -e option, YAML for -s option
```

asdf/commands/edit.py
kmacdonald-stsci: I removed it.

asdf/commands/edit.py
kmacdonald-stsci: I updated this.

asdf/commands/edit.py
kmacdonald-stsci: I added that extension as a check.

asdf/commands/tests/test_edit.py
kmacdonald-stsci: I added this due to the Windows problems I was having during Travis-CI testing.  I removed it.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+
+def is_valid_path_and_ext(fname, wanted_ext=None):
+    """
+    Validates the path exists and the extension is one wanted.
+
+    Parameters
+    ----------
+    fname : str
+        Input file name.
+    wanted_ext : List of extensions to check.
```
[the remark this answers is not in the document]
kmacdonald-stsci: Changed parameter string.

asdf/commands/edit.py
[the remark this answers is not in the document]
kmacdonald-stsci: I added the close parenthesis.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+__all__ = ["edit"]
+
+
+class Edit(Command):
+    @classmethod
+    def setup_arguments(cls, subparsers):
+        """
+        Set up a command line argument parser for the edit subcommand.
+        """
+        desc_string = (
```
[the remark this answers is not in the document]
kmacdonald-stsci: I added the correct spacing.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    Makes sure the header line is the expected one, as well
+    as getting the optional comment line.
+
+    Parameters
+    ----------
+    fd : GenericFile
+
+    Return
+    ------
+    The ASDF header line and the ASDF comment as bytes.
```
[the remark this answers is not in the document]
kmacdonald-stsci: Changed the Return string.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    that will be outputted to a YAML file.
+
+    Parameters
+    ----------
+    fname : str
+        Input file name
+
+    Return
+    ------
+    File descriptor for ASDF file and the ASDF header and ASDF comments as bytes.
```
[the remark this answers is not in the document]
kmacdonald-stsci: Changed the Return string.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+
+def read_and_validate_yaml(fd, fname):
+    """
+    Get the YAML text from an ASDF formatted file.
+
+    Parameters
+    ----------
+    fname : str
+        Input file name
+    fd : GenericFile for fname.
```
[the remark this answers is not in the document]
kmacdonald-stsci: Changed the Parameters string.

asdf/commands/edit.py
```diff
@@ -0,0 +1,589 @@
...
+    yaml_content = reader.read()
+
+    # Create a YAML tree to validate
+    # The YAML text must be converted to a stream.
+    tree = yamlutil.load_tree(io.BytesIO(yaml_content))
+    if tree is None:
+        print("Error: 'yamlutil.load_tree' failed to return a tree.")
+        sys.exist(1)
+
+    schema.validate(tree, None)  # Failure raises an exception.
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed `None`.

asdf/commands/edit.py
```diff
@@ -0,0 +1,529 @@
...
+        else :
+            break
+    return bindex, asdf_blocks[:k]
+
+    
+def write_block_index ( fd, index ) :
+    if len(index) < 1 :
+        return
+
+    bindex_hdr = b'#ASDF BLOCK INDEX\n%YAML 1.1\n---\n'
```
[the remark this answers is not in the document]
kmacdonald-stsci: I created a function that mirrors that method, using the `constants` module and pyyaml.

asdf/commands/edit.py
kmacdonald-stsci: I changed the functionality to add padding after the "..." delimiter and before the first block magic header.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+        return edit(args)
+
+
+def is_yaml_file(fname):
+    """
+    Determines if a file is a YAML file based only on the file extension.
+
+    Parameters
+    ----------
+    fname : The character string of the input file name.
```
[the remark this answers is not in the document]
kmacdonald-stsci: All docstrings have been changed to conform to this format.

asdf/asdf.py
```diff
@@ -763,6 +761,9 @@ def _parse_comment_section(cls, content):
 
     @classmethod
     def _find_asdf_version_in_comments(cls, comments):
+        """ From the initial comments line in an ASDF file, capture the ASDF
```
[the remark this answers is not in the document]
kmacdonald-stsci: I changed the formatting of all triple quotes to conform to this format.

asdf/commands/edit.py
```diff
@@ -0,0 +1,545 @@
...
+
+    # Get YAML reader and content
+    reader = fd.reader_until(constants.YAML_END_MARKER_REGEX,
+                             7,
+                             'End of YAML marker',
+                             include=True,
+                             initial_content=token)
+    yaml_content = reader.read()
+
+    # Create a YAML tree to validate
```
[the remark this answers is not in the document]
kmacdonald-stsci: I changed the functionality so the only time the YAML is validated is on the `save` option and only the edited YAML is validated to ensure only properly formatted YAML is saved.

asdf/commands/edit.py
kmacdonald-stsci: I have added the conditional branch to handle a streamed binary block and included a test for this condition.

asdf/commands/edit.py
kmacdonald-stsci: I have changed all docstrings to conform to this format.

asdf/commands/edit.py
kmacdonald-stsci: I have added the case for a streamed binary block and included a test for it.

asdf/commands/edit.py
```diff
@@ -0,0 +1,618 @@
...
+    fname : str
+        Input ASDF file name.
+
+    Return
+    ------
+    int
+        Location, in bytes, of the first binary block.
+    """
+    with generic_io.get_file(fname, mode="r") as fd:
+        reader = fd.reader_until(
```
eslavich, line 411: Is it safe to seek from the beginning of the file to the start of the block area in this way?  All the bytes in `constants.BLOCK_MAGIC` are in the [YAML valid ranges](https://yaml.org/spec/1.1/current.html#id868518).  In asdf.py we only start looking for the first block after we've consumed the YAML document.

asdf/commands/edit.py
```diff
@@ -0,0 +1,618 @@
...
+
+from .. import generic_io
+from .. import schema
+from .. import yamlutil
+
+from .main import Command
+
+__all__ = ["edit"]
+
+yaml_version = None
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed the use of this global variable.

asdf/commands/edit.py
kmacdonald-stsci: I changed the search to first find the YAML ending delimiter, then find the first block after that delimiter.

eslavich:
@kmacdonald-stsci I discovered a bug in testing the tool, kind of a silly case but I think we should handle it:

- Create an ASDF file with no binary blocks
- Use `asdftool edit -e` to split the file
- In all cases of editing the YAML (equal, shorter, or longer file size) `asdftool edit -s` raises the following exception:

```
  File "/Users/eslavich/anaconda3/envs/asdftool-edit/bin/asdftool", line 33, in <module>
    sys.exit(load_entry_point('asdf', 'console_scripts', 'asdftool')())
  File "/Users/eslavich/repos/asdf/asdf/commands/main.py", line 87, in main
    sys.exit(main_from_args(args))
  File "/Users/eslavich/repos/asdf/asdf/commands/main.py", line 70, in main_from_args
    result = args.func(args)
  File "/Users/eslavich/repos/asdf/asdf/commands/edit.py", line 91, in run
    return edit(args)
  File "/Users/eslavich/repos/asdf/asdf/commands/edit.py", line 649, in edit
    return save_func(args.fname, args.oname)
  File "/Users/eslavich/repos/asdf/asdf/commands/edit.py", line 633, in save_func
    loc = find_first_block(oname)
  File "/Users/eslavich/repos/asdf/asdf/commands/edit.py", line 432, in find_first_block
    reader.read()  # Read to the beginning of the first binary block.
  File "/Users/eslavich/repos/asdf/asdf/generic_io.py", line 224, in read
    raise ValueError("{0} not found".format(self._delimiter_name))
ValueError: First binary block not found
```

ASDF files without blocks will become more common after we change the ndarray auto-inline threshold to 100, which I think I heard is coming soon.

eslavich (approved):
I think this turned out great!  Thanks for bearing with all the review.  I suspect we'll want the next version (that launches an editor automatically) to be able to work with files without binary blocks, but the current solution is fine for now.

PR #874, opened 2020-09-04
Tweaks to new extension code

eslavich:
Some fixes for issues that were uncovered while testing converters in asdf-astropy:

- Sort plugins by package name + plugins from the asdf package at the end so that extension packages can override built-in plugins.
- Change `AsdfFile.extensions` property to return "user extensions" only, that is to say, extensions that were passed in to `AsdfFile.__init__` or `asdf.open`.
- Update docstrings to note where `ExtensionProxy` is returned.
- Remove some features that allowed referring to extensions by URI.  I think I went a little overboard with implementing those before we had a real use case, they can always be revived later.
- Include the extension URI in the `ExtensionProxy` repr to aid in debugging.

### DIFF
--- a/asdf/extension/_extension.py
+++ b/asdf/extension/_extension.py
@@ -311,10 +311,14 @@ class ExtensionProxy(Extension, AsdfExtension):
         if self.package_name is None:
             package_description = "(none)"
         else:
-            package_description = "{}=={}".format(self.package_name, self.package_version)
+            package_description = f"{self.package_name}=={self.package_version}"
 
-        return "<ExtensionProxy class: {} package: {} legacy: {}>".format(
-            self.class_name,
-            package_description,
-            self.legacy,
+        if self.extension_uri is None:
+            uri_description = "(none)"
+        else:
+            uri_description = self.extension_uri
+
+        return (
+            f"<ExtensionProxy URI: {uri_description} class: {self.class_name} "
+            f"package: {package_description} legacy: {self.legacy}>"
         )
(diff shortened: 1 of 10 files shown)


asdf/extension/_extension.py
```diff
@@ -313,7 +313,13 @@ def __repr__(self):
...
+        if self.extension_uri is None:
+            uri_description = "(none)"
+        else:
+            uri_description = self.extension_uri
+
+        return "<ExtensionProxy URI: {} class: {} package: {} legacy: {}>".format(
+            uri_description,
             self.class_name,
             package_description,
             self.legacy,
```
jdavies-st, line 325: Hey, since we're now python 3.6+, we can use f-strings here.

asdf/fits_embed.py
```diff
@@ -199,9 +199,8 @@ def open(cls, fd, uri=None, validate_checksums=False, extensions=None,
         extensions : object, optional
             Additional extensions to use when reading and writing the file.
             May be any of the following: `asdf.extension.AsdfExtension`,
-            `asdf.extension.Extension`, `str` extension URI,
-            `asdf.extension.AsdfExtensionList` or a `list` of URIs and/or
-            extensions.
+            `asdf.extension.Extension`, `asdf.extension.AsdfExtensionList`
+            or a `list` extensions.
```
jdavies-st, line 203:
```suggestion
            or a `list` of extensions.
```

jdavies-st (approved):
Looks good.  👍

asdf/extension/_extension.py
eslavich: f-f-f-f-fantastic!

PR #877, opened 2020-09-21
Add bandit to CI jobs and fix or # nosec failed lines

eslavich:
This PR adds bandit to to the Travis jobs and fixes or marks `# nosec` failing lines in the non-test (and non-extern) code.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -109,4 +109,4 @@ basepython= python3.8
 deps=
     bandit
 commands=
-    bandit -lll -r -x asdf/tests,asdf/commands/tests,asdf/tags/core/tests,asdf/extern asdf
+    bandit -r -x asdf/tests,asdf/commands/tests,asdf/tags/core/tests,asdf/extern asdf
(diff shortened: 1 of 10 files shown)


tox.ini
```diff
@@ -103,3 +103,10 @@ deps=
 extras= all,tests
 commands=
     pytest compatibility_tests/ --remote-data
+
+[testenv:bandit]
+basepython= python3.8
+deps=
+    bandit
+commands=
+    bandit -lll -r -x asdf/tests,asdf/commands/tests,asdf/tags/core/tests,asdf/extern asdf
```
jdavies-st, line 112: What happens if `bandit -ll` and you report MEDIUM severity issues?  Are there a lot of those?

eslavich: This PR fixes or explains away all issues at every severity level, so this would pass at `-l` or `-ll`.

jdavies-st (approved):
Awesome.  Looks good.

And we shouldn't worry about the decrease in coverage, as it wasn't tested before with the asserts either.

PR #885, opened 2020-11-11
Add user-friendly names to Travis CI jobs

jdavies-st:
- Add names to Travis CI jobs
- ~Add Python 3.9 to the matrix, and fix the bug found~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,43 @@
 os: linux
 dist: xenial
 language: python
-python: 3.8
+python: 3.8.6
 
 jobs:
   include:
     # Python 3.8, stable dependencies
-    - env: TOXENV=py38
+    - name: Python 3.8
+      env: TOXENV=py38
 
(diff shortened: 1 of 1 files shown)


eslavich (approved):
I am a user and I do indeed find this friendly!

As an aside, I wonder when those numpy dev failures are going to bite us...

jdavies-st:
`numpy` never even compiled in that job.  I wonder what the issue is in compiling it.  I see that the Astropy folks instead pull a binary version that is built nightly by the anaconda folks.

https://github.com/astropy/astropy/blob/3627a7473e7c73f2e46e74157eaa4f078ccb8574/tox.ini#L12-L13

jdavies-st:
Well, this PR went off the rails pretty quickly.

I'll back out the addition of a Python 3.9 job and deal with that in a separate PR.  Clearly it is not trivial.

And clearly allowing `pytest` version to float is also not trivial.  Separate PR for that as well.

PR #886, opened 2020-11-13
Make asdf compatible with Python 3.9

jdavies-st:
- Fix `importlib.abc.Traversable` bug
- Add Python 3.9 job to Travis CI matrix
- Update setup.cfg with `project_urls`
- Fix dev urls in `tox.ini`
- Update `_docdir` pytest autouse fixture to work with Pytest 6.0 and the latest `pytest-doctestplus`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ python: 3.8.6
 
 jobs:
   include:
+    # Python 3.9, stable dependencies
+    - name: Python 3.9
+      env: TOXENV=py39
+      python: 3.9.0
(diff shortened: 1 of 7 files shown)


jdavies-st:
A note here that only a subset of the tests get run in `pytest --pyargs asdf` mode, due to the way we package the installed code.  I don't think any of the doctests get run, as `setup.cfg` configures that, and it is not included in the installed package.

eslavich (approved):
Super duper, thanks!

Issue #887, opened 2020-11-24
[2.7.1] Test Errors with Python 3.9


olebole:
Hi,

on Debian, we are currently migrating to Python 3.9. With the released version, we get now errors when running the tests:
```
============================= test session starts ==============================
platform linux -- Python 3.9.0+, pytest-4.6.11, py-1.9.0, pluggy-0.13.0
rootdir: /tmp/autopkgtest-lxc.dhe2dv4q/downtmp/autopkgtest_tmp
plugins: openfiles-0.5.0, arraydiff-0.3, hypothesis-5.41.2, remotedata-0.3.2,
asdf-2.7.1, astropy-header-0.1.2, doctestplus-0.7.0, filter-subpackage-0.1.1,
cov-2.8.1
collected 462 items / 2 errors / 460 selected

==================================== ERRORS ====================================
_______________ ERROR collecting tags/core/tests/test_ndarray.py _______________
/usr/lib/python3/dist-packages/asdf/tags/core/tests/test_ndarray.py:25: in <module>
    TEST_DATA_PATH = helpers.get_test_data_path('', module=test_data)
/usr/lib/python3/dist-packages/asdf/tests/helpers.py:68: in get_test_data_path
    with resources.path(module, name) as path:
/usr/lib/python3.9/contextlib.py:117: in __enter__
    return next(self.gen)
/usr/lib/python3.9/importlib/resources.py:175: in _path_from_reader
    opener_reader = reader.open_resource(norm_resource)
<frozen importlib._bootstrap_external>:995: in open_resource
    ???
E   IsADirectoryError: [Errno 21] Is a directory:
'/usr/lib/python3/dist-packages/asdf/tags/core/tests/data'
_____________________ ERROR collecting tests/test_types.py _____________________
/usr/lib/python3/dist-packages/asdf/tests/test_types.py:20: in <module>
    TEST_DATA_PATH = str(helpers.get_test_data_path(''))
/usr/lib/python3/dist-packages/asdf/tests/helpers.py:68: in get_test_data_path
    with resources.path(module, name) as path:
/usr/lib/python3.9/contextlib.py:117: in __enter__
    return next(self.gen)
/usr/lib/python3.9/importlib/resources.py:175: in _path_from_reader
    opener_reader = reader.open_resource(norm_resource)
<frozen importlib._bootstrap_external>:995: in open_resource
    ???
E   IsADirectoryError: [Errno 21] Is a directory:
[...]

PR #889, opened 2020-11-24
Do not request a directory from importlib.resources.path

eslavich:
This change should fix the test failure reported in #887.  `importlib.resources.path` is clearly documented as not supporting directories:

```
package is either a name or a module object which conforms to the Package requirements. resource
is the name of the resource to open within package; it may not contain path separators and it may not
have sub-resources (i.e. it cannot be a directory).
```

In our Travis jobs `importlib.resources.path(some_module, "")` returns the path containing the module's `__init__.py`, but evidently that is not the case in all environments.

I ended up removing use of `importlib.resources` here since we're making another mistake by returning the context manager's path.  In some cases that path will point to a temporary file that is deleted on `__exit__`, so it's not appropriate to return it for use outside of the context manager.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,7 +43,12 @@
 - Add ``edit`` subcommand to asdftool for efficient editing of
   the YAML portion of an ASDF file.  [#873]
 
-2.7.1 (unreleased)
+2.7.2 (unreleased)
+------------------
+
+- Fix bug causing test collection failures in some environments. [#889]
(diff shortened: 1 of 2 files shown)


olebole:
I can confirm that this PR fixes the issue on Debian.

jdavies-st (approved):
LGTM.  This should probably be backported to the `2.7x` branch too?

PR #894, opened 2020-12-07
Literal bit limit changed from 52 bits to 64 bits

### LINKED ISSUE #891: auto inline for large integers
During testing against the current asdf master we noticed some test failures due to the change introduced in #882 .

Since previous asdf versions defaulted to store arrays in blocks, it was "safe" to write arrays with large (>2^51) integer values using default settings.
On the current version, the default settings might fail to write the asdf file because the default value for `auto_inline` tries to inline smaller arrays that may contain large integer values (e.g. ns timestamps since 1970)

Here is a simple example that works on 2.7.1 but fails with current `auto_inline`
```python
import asdf
import numpy as np

with asdf.AsdfFile(tree={"test": np.array([1, 2**51])}) as af:
    af.write_to("test.asdf")
```

I think it would be good for asdf to allow this special case with default parameters.
One solution might be to specifically check integer arrays that should get inlined because of the current `auto_inline` setting (should be "cheap" since they are short) and store them as binary block if large values are encountered. Maybe issue a warning in this case.
I have tested a quick implementation here: https://github.com/asdf-format/asdf/compare/master...CagtayFabry:fix_large_integer_inline

If the user specifically requests "inline" for a specific block or the whole file, I think it is reasonable to throw the error.

What do you think @eslavich @kmacdonald-stsci ?

kmacdonald-stsci:
Removal of the 52 bit limit for numbers in an ndarray (originally imposed for anticipated use with javascript), which allows usage for larger numbers.  The `ndarray` is still limited by 64 bits by the `dtype`.  In addition to removing this validation check, the `pytest` tests used to test the original limit have been removed as well.

Resolves #891

### DIFF
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -630,7 +630,6 @@ def validate(instance, ctx=None, schema={}, validators=None, reading=False,
                               *args, **kwargs)
     validator.validate(instance, _schema=(schema or None))
 
-    # additional_validators = [_validate_large_literals]
     additional_validators = []
     if ctx.version >= versioning.RESTRICTED_KEYS_MIN_VERSION:
         additional_validators.append(_validate_mapping_keys)
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -567,39 +567,6 @@ def get_validator(schema={}, ctx=None, validators=None, url_mapping=None,
     return validator
 
 
-def _validate_large_literals(instance, reading):
-    """
-    Validate that the tree has no large numeric literals.
-    """
-    def _validate(value):
-        bits = 63
-        max_integer_limit = (1 << bits) - 1
-        min_integer_limit = -((1 << bits) - 2)
-
-        if value <= max_integer_limit and value >= min_integer_limit:
-            return
-
-        if reading:
-            warnings.warn(
-                f"Invalid integer literal value {value} detected while reading file. "
-                "The value has been read safely, but the file should be "
-                "fixed.",
-                AsdfWarning
-            )
-        else:
-            raise ValidationError(
-                f"Integer value {value} is too large to safely represent as a "
-                "literal in ASDF"
-            )
-
-    if isinstance(instance, Integral):
-        _validate(instance)
-    elif isinstance(instance, Mapping):
-        for key in instance:
-            if isinstance(key, Integral):
-                _validate(key)
-
-
 def _validate_mapping_keys(instance, reading):
     """
     Validate that mappings do not contain illegal key types
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -573,7 +573,10 @@ def _validate_large_literals(instance, reading):
     """
     def _validate(value):
         # We can count on 52 bits of precision
-        if value <= ((1 << 51) - 1) and value >= -((1 << 51) - 2):
+        bits = 63
+        max_integer_limit = (1 << bits) - 1
+        min_integer_limit = -((1 << bits) - 2)
+        if value <= max_integer_limit and value >= min_integer_limit:
             return
 
         if reading:
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -572,10 +572,10 @@ def _validate_large_literals(instance, reading):
     Validate that the tree has no large numeric literals.
     """
     def _validate(value):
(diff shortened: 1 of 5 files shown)


asdf/schema.py
```diff
@@ -567,36 +567,6 @@ def get_validator(schema={}, ctx=None, validators=None, url_mapping=None,
     return validator
 
 
-def _validate_large_literals(instance, reading):
-    """
-    Validate that the tree has no large numeric literals.
-    """
-    def _validate(value):
-        # We can count on 52 bits of precision
-        if value <= ((1 << 51) - 1) and value >= -((1 << 51) - 2):
```
eslavich, line 576: @perrygreenfield do we still want to limit integers in some way (within range of 64-bit signed, for example)?  Or should we allow trees like this:
```python
with asdf.AsdfFile({"value": 2**70}) as af:
  af.write_to("test.asdf")
```

kmacdonald-stsci: The max bit level check has been added back, limiting an integer to 64 bits.

asdf/tests/test_schema.py
```diff
@@ -617,51 +619,59 @@ def test_schema_resolved_via_entry_points():
...
-    ff.set_array_storage(ff.tree['large_array'], 'inline')
-    buff = io.BytesIO()
+@pytest.mark.parametrize("num, ttype", [
+    (constants.MAX_NUMBER+1, "val"),
+    (constants.MAX_NUMBER+1, "list"),
+    (constants.MAX_NUMBER+1, "key"),
+    (constants.MIN_NUMBER-1, "val"),
+    (constants.MIN_NUMBER-1, "list"),
+    (constants.MIN_NUMBER-1, "key")
+])
```
jdavies-st, line 654: Since each `ttype` string is tested with each value of `num`, one can doubly-parametrize like so:

```python
@pytest.mark.parametrize("num", [constants.MAX_NUMBER+1, constants.MIN_NUMBER-1])
@pytest.mark.parametrize("ttype", ["val", "list", "key"])
def test_max_min_literals_write(num, ttype, tmpdir):
    ...
```

jdavies-st (approved):
Looks good.  Just a minor comment and question below.  🚀

asdf/tests/test_api.py
```diff
@@ -401,22 +401,16 @@ def test_auto_inline_masked_array(tmpdir):
...
-    outfile = str(tmpdir.join('test.asdf'))
-    arr = np.array([2**52 + 1] + [k for k in range(constants.DEFAULT_AUTO_INLINE)])
-    tree = {"array": arr}
-
-    with asdf.AsdfFile(tree) as af:
-        af.write_to(outfile)
-        assert len(list(af.blocks.inline_blocks)) == 0
-        assert len(list(af.blocks.internal_blocks)) == 1
+@pytest.mark.parametrize("val", [constants.MAX_NUMBER+1, constants.MIN_NUMBER-1])
+def test_auto_inline_large_value(val):
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this test.  The original test made sure that for short arrays, but large numbers the 52 limit was enforced.  Since that limit is removed, this test no longer applies.

asdf/tests/test_schema.py
kmacdonald-stsci: I rewrote the parametrization to do this double parametrization.

eslavich:
What should we do about uint64 arrays that are small enough to be auto inlined?  For example this:

```python
import numpy as np
import asdf

arr = np.array([2**64 - 1], dtype=np.uint64)

with asdf.AsdfFile({"arr": arr}) as af:
  af.write_to("test.asdf")
```

results in the following error:

```
ValidationError: Integer value 18446744073709551615 is too large to safely represent as a literal in ASDF
```

asdf/tests/test_api.py
```diff
@@ -401,24 +401,6 @@ def test_auto_inline_masked_array(tmpdir):
         assert len(list(af.blocks.internal_blocks)) == 2
 
 
-def test_auto_inline_large_value(tmpdir):
```
eslavich, line 404: Should we keep this test and rework it to cover the np.uint64 case?

Issue #818, opened 2020-07-08
Replace Travis with GitHub actions


eslavich:
Some dissatisfying aspects of Travis:

- The s390x build has always been flaky, but lately it's been failing consistently (timeouts during install)
- GitHub integration has degenerated in recent months, leaving the build status icons out of sync
- I just burned a bunch of time troubleshooting the MacOS build after Travis made an unannounced change to the container

Enough reason to give GitHub actions a try -- since we already have tox set up, it should be straightforward.

Cadair:
Github actions is great, I would also say that Azure pipelines is nice, and there is a bunch of work Tom and I did to make it easy in https://github.com/OpenAstronomy/azure-pipelines-templates

In reality there isn't a huge amount to choose between them.

jdavies-st:
Here's an action for s390x jobs.

https://github.com/marketplace/actions/run-on-architecture

PR #896, opened 2020-12-18
Create github actions workflow

kmacdonald-stsci:
For this PR, I deleted the Travis CI script and added a basic github actions work flow.  This basic github action needs to be expanded to cover all desired cases, but this works as a place to start from to get some testing done while transitioning from Travis CI.

Addresses #818


.github/workflows/asdf_ci.yml
```diff
@@ -0,0 +1,91 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  CRDS_SERVER_URL: "https://jwst-crds.stsci.edu"
+  CRDS_PATH: "/tmp/crds_cache"
```
jdavies-st, line 11: We don't use `CRDS` with `asdf`, so these are probably not necessary.

.github/workflows/asdf_ci.yml
```diff
@@ -0,0 +1,91 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8
+    - name: Lint with flake8
+      run: |
+        flake8 asdf --count --show-source --statistics
```
jdavies-st, line 29: Basically, all the steps are the same for each job, except for the install command and the tox command.  I'd recommend something like

https://github.com/astropy/astropy/blob/9c79f08d72cdd7982ea990d9b01553ac9c4d657b/.github/workflows/ci_workflows.yml#L67-L88

So that when we have 10 jobs in the matrix, you're not writing 10 sets of steps.  It will be more concise and definitely easier to maintain.

asdf/extern/atomicfile.py
```diff
@@ -71,13 +71,13 @@ def atomic_rename(src, dst):
...
                 raise
-            old = "%s-%08x" % (dst, random.randint(0, sys.maxsize))
+            old = "%s-%08x" % (dst, random.randint(0, sys.maxsize))  # nosec
             os.rename(dst, old)
             os.rename(src, dst)
             try:
                 os.unlink(old)
             except Exception:
-                pass
+                print(f"'os.unlink({old})' failed")
```
jdavies-st, line 80: We shouldn't be doing any `bandit` checks on the `extern` code.  Use the tox env already set up for bandit to do this check, as it already has the paths that need to be skipped.

jdavies-st:
Also, you may want to submit PR adding a dummy file inside the `.github/workflows/` subdir first, which we can then merge.  Github actions will then pick this up as a change and run them in this PR so we can see the checks running.

At least that's my understanding of the same issue that we had with `romancal`?

jdavies-st:
Finally, we should be deleting `.travis.yml` yet.  We still have to port over the rest of the matrix there before we delete.

PR #900, opened 2021-01-07
GitHub actions, no big endian test

### LINKED ISSUE #818: Replace Travis with GitHub actions

kmacdonald-stsci:
I added all Travis CI tests to github actions, with the exception of the big endian test.  There is no straight forward support for big endian architecture in github actions, so will be done in a separate workflow.  There are two known failures, the Numpy Dev and Pre-Release Dependencies tests.

Resolves #818

### DIFF
--- a/.github/workflows/asdf_ci.yml
+++ b/.github/workflows/asdf_ci.yml
@@ -7,28 +7,132 @@ on:
     branches: [ master ]
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
+  tox:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Code Coverage with Python 3.9
+            os: ubuntu-latest
+            python-version: 3.9
+            toxenv: coverage
 
+          - name: Python 3.8 Testing
+            os: ubuntu-latest
+            python-version: 3.8
+            toxenv: py38
+
+          - name: Python 3.7 Testing
+            os: ubuntu-latest
+            python-version: 3.7.9
+            toxenv: py37
+
+          - name: Python 3.6 Testing
+            os: ubuntu-latest
+            python-version: 3.6.10
+            toxenv: py36
+
+          - name: Documentation Build
+            os: ubuntu-latest
+            python-version: 3.8
+            toxenv: docbuild
+
+          - name: Mac OS Latest
+            os: macos-latest
+            python-version: 3.9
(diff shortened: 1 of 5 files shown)


.github/workflows/asdf_ci.yml
```diff
@@ -7,41 +7,144 @@ on:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+      - name: Run tox
+        run: tox -e ${{ matrix.toxenv }}
 
-  build:
+  windows_build:
+    name: Windows with Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
```
jdavies-st, line 132: Is this correct?  Ubuntu is Linux, and this is supposed to be Windows?

tox.ini
```diff
@@ -70,7 +70,7 @@ commands=
     sphinx-build -W docs build/docs
 
 [testenv:checkdocs]
-basepython= python3.8
+basepython= python3.9
```
jdavies-st, line 73: Setting `basepython` in the `testenv` here is not necessary.  These should just be removed.  If not specified, it will use whatever version is available, or you can do `py38-checkdocs` or `py39-coverage` to get the combo you want.  Hard-coding in the env makes the env less flexible.

jdavies-st (commented):
Why is the Windows build separate from all the others, and from what I can see, it doesn't look like its running on Windows.

.github/workflows/asdf_ci.yml
jdavies-st: I think this should be:

```yaml
runs-on: windows-latest
```

And perhaps it can be part of the main matrix?

jdavies-st:
For the Windows test, it looks like the test is expecting an `IOError` but it is actually raising a `ValueError`.  Strange.

jdavies-st (approved):
This looks great.  ~Other than figuring out the errors in the 3 jobs that are failing, it's good.~

EDIT: Ah right, the numpy dev tests are failing because `numpy 1.20` shows errors. That's a separate issue.

jdavies-st (approved):
LGTM.  Ready to merge.

PR #901, opened 2021-01-12
Fix errors and warnings with numpy 1.20

eslavich:
This fixes an issue with numpy 1.20 due to code in compression.py assigning an empty bytes object to an empty ndarray range.  I also changed instances of np.int and np.float to int and float, respectively, since the numpy aliases have been marked as deprecated in 1.20.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,8 @@
 
 - Fix bug causing test collection failures in some environments. [#889]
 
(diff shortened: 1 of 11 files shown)


kmacdonald-stsci (approved):
Everything looks good.

jdavies-st (approved):
🚀

PR #908, opened 2021-01-14
Add s390x workflow

eslavich:
This adds a new actions workflow that runs the tests on s390x architecture.  It's scheduled to run when a new tag is created (presumably a release candidate) and Mondays at 6am UTC.

### DIFF
--- a/.github/workflows/s390x.yml
+++ b/.github/workflows/s390x.yml
@@ -1,0 +1,42 @@
+name: s390x
+
+on:
+  push:
+    tags:
+      - '*'
+  schedule:
+    # Run every Monday at 6am UTC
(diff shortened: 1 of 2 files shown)


eslavich:
Well that was arduous.  Here's a link to a passing s390x job: https://github.com/asdf-format/asdf/runs/1706314754

kmacdonald-stsci (approved):

jdavies-st (approved):
Looks good.

Is there a way to run the tests on this runner as a non-root user?

jdavies-st:
Also, 47min is not too bad on this.  But agree that having it as a cron is better than per-PR CI.

PR #909, opened 2021-01-15
Fix buffer assignment in compression.py

eslavich:
I missed that we were setting `buffer[...]` here and not `buffer.data[...]`!  That's the real source of the problem.  This is the difference between them:

```python
In [10]: buffer = np.zeros((10,), np.uint8)
    ...: buffer.data[0:5] = b'12345'
    ...: buffer
Out[10]: array([49, 50, 51, 52, 53,  0,  0,  0,  0,  0], dtype=uint8)

In [11]: buffer = np.zeros((10,), np.uint8)
    ...: buffer[0:5] = b'12345'
    ...: buffer
Out[11]: array([57, 57, 57, 57, 57,  0,  0,  0,  0,  0], dtype=uint8)
```

(where 57 is 12345 % 2**8)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,7 +48,7 @@
 
 - Fix bug causing test collection failures in some environments. [#889]
 
-- Fix bug when decompressing arrays with numpy 1.20.  [#901]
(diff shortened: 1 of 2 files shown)


jdavies-st (approved):
Looks good.

PR #914, opened 2021-01-15
Add workflow that tests downstream packages

eslavich:
This adds a workflow that tests packages that depend on asdf.  We can trigger it manually before a release and it's also scheduled to run weekly.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,83 @@
+name: Downstream Packages
+
+on:
+  workflow_dispatch:
+    inputs:
+      asdf_ref:
+        description: asdf ref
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
Looks good.  What's the reason for separating out the `astropy` matrix?  Is it because `asdf` testing depends on `astropy`?

eslavich:
> What's the reason for separating out the astropy matrix? Is it because asdf testing depends on astropy?

astropy is fussy about how you run the tests when the package is installed (leads to import errors when it finds e.g. conftest.py for the same module in two places).  I tried adding extra variables to the matrix to make it work, but the result was ugly.

PR #922, opened 2021-02-02
Automatically launch editor when user invokes asdftool edit

eslavich:
This updates the new "asdftool edit" subcommand to automatically generate a temporary file name and open an editor.  There are heavier changes than might be expected because I discovered/created some opportunities to share more code with asdf.py and block.py.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,7 +41,7 @@
   than defaulting to storing all arrays internally. [#882]
 
 - Add ``edit`` subcommand to asdftool for efficient editing of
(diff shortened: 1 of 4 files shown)


eslavich:
I manually tested the command on both Windows 10 and macOS.

eslavich:
@kmacdonald-stsci found a serious bug in this branch where blocks were being lost when the YAML area was expanded.  The latest commit fixes the problem and improves test coverage of the case when blocks are present.

kmacdonald-stsci (approved):

PR #929, opened 2021-02-22
Fix tests impacted by numpy 1.20 deprecations

eslavich:
This fixes tests that are currently busted in master due to numpy deprecations and a warning out of astropy.  More detail below.

### DIFF
--- a/asdf/tags/core/tests/test_ndarray.py
+++ b/asdf/tags/core/tests/test_ndarray.py
@@ -368,7 +368,7 @@ def test_string(tmpdir):
 
 def test_string_table(tmpdir):
     tree = {
-        'table': np.array([(b'foo', 'სამეცნიერო', 42, 53.0)])
+        'table': np.array([(b'foo', 'სამეცნიერო', '42', '53.0')])
         }
 
     helpers.assert_roundtrip_tree(tree, tmpdir)
--- a/pytest_asdf/plugin.py
+++ b/pytest_asdf/plugin.py
@@ -2,6 +2,7 @@ import io
 import os
 from importlib.util import find_spec
 from pkg_resources import parse_version
+import pathlib
 
 import yaml
 import pytest
@@ -16,6 +17,12 @@ def pytest_addoption(parser):
         "asdf_schema_root", "Root path indicating where schemas are stored")
     parser.addini(
         "asdf_schema_skip_names", "Base names of files to skip in schema tests")
+    parser.addini(
+        "asdf_schema_skip_tests",
+        "List of tests to skip, one per line, in format <schema path suffix>::<test name>")
+    parser.addini(
+        "asdf_schema_xfail_tests",
+        "List of tests to xfail, one per line, in format <schema path suffix>::<test name>")
     parser.addini(
         "asdf_schema_skip_examples",
         "Base names of schemas whose examples should not be tested")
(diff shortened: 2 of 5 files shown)


asdf/tags/core/tests/test_ndarray.py
```diff
@@ -368,7 +368,7 @@ def test_string(tmpdir):
 
 def test_string_table(tmpdir):
     tree = {
-        'table': np.array([(b'foo', 'სამეცნიერო', 42, 53.0)])
+        'table': np.array([(b'foo', 'სამეცნიერო', '42', '53.0')])
```
eslavich, line 371: numpy is deprecating the behavior that converts these non-string values to strings

pytest_asdf/plugin.py
```diff
@@ -16,6 +17,12 @@ def pytest_addoption(parser):
         "asdf_schema_root", "Root path indicating where schemas are stored")
     parser.addini(
         "asdf_schema_skip_names", "Base names of files to skip in schema tests")
+    parser.addini(
```
eslavich, line 20: Added a feature here that allows individual tests to be skipped or xfailed via config options

setup.cfg
```diff
@@ -83,22 +83,27 @@ open_files_ignore = test.fits asdf.fits
...
+    stsci.edu/asdf/wcs/frame-1.0.0.yaml
+    stsci.edu/asdf/wcs/frame-1.1.0.yaml
+    stsci.edu/asdf/wcs/spectral_frame-1.1.0.yaml
+    stsci.edu/asdf/wcs/step-1.1.0.yaml
+    stsci.edu/asdf/wcs/step-1.2.0.yaml
+    stsci.edu/asdf/wcs/wcs-1.1.0.yaml
+    stsci.edu/asdf/wcs/wcs-1.2.0.yaml
+    stsci.edu/yaml-schema/draft-01.yaml
+asdf_schema_xfail_tests =
+    stsci.edu/asdf/core/ndarray-1.0.0.yaml::test_example_2
```
eslavich, line 106: This example is unsupported by this library.  I suspect we'll end up removing it but for now xfail seems reasonable.

jdavies-st, line 106: Is there a reason to change the config name from `asdf_schema_skip_names` to `asdf_schema_skip_tests`?  I suspect so that the full path can be used and not break existing usage?

tox.ini
```diff
@@ -39,7 +39,10 @@ pip_pre= true
 [testenv:warnings]
 basepython= python3.8
 commands=
-    pytest --remote-data -W error -W ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes
+    pytest --remote-data -W error \
+      -W ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes \
+      -W 'ignore:numpy.ndarray size changed:astropy.utils.exceptions.AstropyWarning' \
+      -W 'ignore:numpy.ndarray size changed:RuntimeWarning'
```
jdavies-st, line 45: If you would like, you can add these to `pytest` config (such as in setup.cfg) like so:

```ini
[tool:pytest]
filterwarnings =
    error
    ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes
    ignore:numpy.ndarray size changed:astropy.utils.exceptions.AstropyWarning
    ignore:numpy.ndarray size changed:RuntimeWarning
```
But perhaps we don't want to set these generally for all tests.

jdavies-st (approved):
Looks good!  Just a couple minor comments below.

setup.cfg
eslavich: Yeah, exactly.  I didn't like using only the schema basename but also didn't want to break anything.

tox.ini
eslavich: I already added these to setup.cfg in this PR, but the tox job needs them specified explicitly.  I think the command-line `-W error` must override the config.

jdavies-st: Right, it resets it.  Makes sense to have it here too then.

PR #930, opened 2021-02-22
Fix strides issues over FITS arrays and base arrays with negative strides

### LINKED ISSUE #916: Scipy 1.6.0 generated ndarray can not be restored
Take a look at the following script:

~~~ python
import numpy as np
from asdf import AsdfFile
from scipy.spatial.transform import Rotation

original_array = np.array([0.2, 1.3, 3.14])
modified_array = Rotation.from_euler(
    seq="xyz", angles=original_array, degrees=False
).as_euler("xyz", degrees=False)

# putting this one into the tree works
copy_of_modified_array = np.array(modified_array)

tree_write = {"array": modified_array}
file_write = AsdfFile(tree=tree_write)
file_write.write_to("test.yaml")

file_read = AsdfFile.open("test.yaml")
tree_read = file_read.tree

np.array(tree_read["array"])
~~~

The last line causes the error:

~~~
  File ".../bug.py", line 19, in <module>
    np.array(tree_read["array"])
  File "...\lib\site-packages\asdf\tags\core\ndarray.py", line 291, in __array__
    return self._make_array()
  File "...\lib\site-packages\asdf\tags\core\ndarray.py", line 269, in _make_array
    self._offset, self._strides, self._order)
ValueError: strides is incompatible with shape of requested array and size of buffer
~~~

This happens with the scipy version 1.6.0. Previous releases do not trigger the error.

I can create a new array from the scipy generated one without problems and putting the copy into the tree removes the error. So `modified_array` must have some special property that lets the asdf deserialization fail, but I couldn't figure out what exactly is causing the error. I initially thought the negative stride (see file content below) is the problem, but that was already the case in previous scipy versions.

I am not sure if this is a scipy problem or if there is an error in the deserialization routine of asdf. A related scipy issue might be [this one](https://github.com/scipy/scipy/issues/13358). 

Here is the file content produced by the script:

~~~ yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
  name: asdf, version: 2.7.2}
history:
  extensions:
    software: !core/software-1.0.0 {name: asdf, version: 2.7.2}
array: !core/ndarray-1.0.0
  source: 0
  datatype: float64
  byteorder: little
  shape: [3]
  strides: [-8]
...
�BLK 0                             ��]��i�+��8 ��������?�������?��Q�	@#ASDF BLOCK INDEX
%YAML 1.1
---
- 538
...
~~~

eslavich:
This PR fixes bugs in serializing views over FITS arrays and base arrays with negative strides.

The astropy.io.fits library always writes arrays in C-contiguous order, but this library has been serializing views as though the arrays were to be written in the same order that they are stored in memory.  This works fine for arrays that are already C-contiguous but corrupts views over F-contiguous or non-contiguous arrays.  The fix here will handle the most common case where the view and the array have the same memory layout.  It raises an error when they do not, which is better than writing unreadable files.  I ran the jwst regression tests on this branch and all tests passed except for 5 known and unrelated failures.

In the case of a base array with negative strides, we've been serializing views relative to the buffer that the base array is striding over, but writing the data in the memory order of the array itself.  The fix here is to write the array in the same order as the underlying buffer which keeps the views valid.  I don't have a test for this case because I haven't been able to create a base array with negative strides using numpy alone.  I did run the script from #916 with scipy 1.6.1 and the array read back in correctly.  If anyone knows how to create such an array do please let me know!

Resolves #916 and spacetelescope/jwst#5699

### DIFF
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -46,13 +46,6 @@ class _FitsBlock:
     def array_storage(self):
         return 'fits'
 
-    def override_byteorder(self, byteorder):
-        # FITS data is always stored in big-endian byte order.
-        # The data array may not report big-endian, but we want
-        # the value written to the tree to match the actual
-        # byte order on disk.
-        return 'big'
-
     @property
     def trust_data_dtype(self):
         # astropy.io.fits returns arrays in native byte order
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -92,7 +92,7 @@ size : integer
 """
 
 
-def _array_tofile_chunked(write, array, chunksize):  # pragma: no cover
+def _array_tofile_chunked(write, array, chunksize):
     array = array.view(np.uint8)
     for i in range(0, array.nbytes, chunksize):
         write(array[i:i + chunksize].data)
@@ -101,8 +101,7 @@ def _array_tofile_chunked(write, array, chunksize):  # pragma: no cover
 def _array_tofile_simple(fd, write, array):
     return write(array.data)
 
-
-if sys.platform == 'darwin':  # pragma: no cover
+if sys.platform == 'darwin':
     def _array_tofile(fd, write, array):
         # This value is currently set as a workaround for a known bug in Python
         # on OSX. Individual writes must be less than 2GB, which necessitates
@@ -112,7 +111,7 @@ if sys.platform == 'darwin':  # pragma: no cover
         if fd is None or array.nbytes >= OSX_WRITE_LIMIT and array.nbytes % 4096 == 0:
             return _array_tofile_chunked(write, array, OSX_WRITE_LIMIT)
         return _array_tofile_simple(fd, write, array)
-elif sys.platform.startswith('win'):  # pragma: no cover
+elif sys.platform.startswith('win'):
     def _array_tofile(fd, write, array):
         WIN_WRITE_LIMIT = 2 ** 30
         return _array_tofile_chunked(write, array, WIN_WRITE_LIMIT)
@@ -370,7 +369,20 @@ class GenericFile(metaclass=util.InheritDocstrings):
     """
 
     def write_array(self, array):
-        _array_tofile(None, self.write, array.ravel(order='K'))
+        """
+        Write array content to the file.  Array must be 1D contiguous
+        so that this method can avoid making assumptions about the
+        intended memory layout.  Endianness is preserved.
+
+        Parameters
+        ----------
+        array : np.ndarray
+            Must be 1D contiguous.
+        """
+        if len(array.shape) != 1 or not array.flags.contiguous:
+            raise ValueError("Requires 1D contiguous array.")
+
+        _array_tofile(None, self.write, array)
 
     def seek(self, offset, whence=0):
         """
@@ -749,7 +761,10 @@ class RealFile(RandomAccessFile):
             arr.flush()
             self.fast_forward(len(arr.data))
         else:
-            _array_tofile(self._fd, self._fd.write, arr.ravel(order='K'))
+            if len(arr.shape) != 1 or not arr.flags.contiguous:
+                raise ValueError("Requires 1D contiguous array.")
+
+            _array_tofile(self._fd, self._fd.write, arr)
 
     def can_memmap(self):
         return True
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -81,7 +81,10 @@ def asdf_datatype_to_numpy_dtype(datatype, byteorder=None):
     raise ValueError("Unknown datatype {0}".format(datatype))
 
 
-def numpy_byteorder_to_asdf_byteorder(byteorder):
+def numpy_byteorder_to_asdf_byteorder(byteorder, override=None):
+    if override is not None:
+        return override
+
     if byteorder == '=':
         return sys.byteorder
     elif byteorder == '<':
@@ -90,7 +93,7 @@ def numpy_byteorder_to_asdf_byteorder(byteorder):
         return 'big'
 
 
-def numpy_dtype_to_asdf_datatype(dtype, include_byteorder=True):
+def numpy_dtype_to_asdf_datatype(dtype, include_byteorder=True, override_byteorder=None):
     dtype = np.dtype(dtype)
     if dtype.names is not None:
         fields = []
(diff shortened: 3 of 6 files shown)


asdf/generic_io.py
```diff
@@ -112,7 +111,7 @@ def _array_tofile(fd, write, array):
         if fd is None or array.nbytes >= OSX_WRITE_LIMIT and array.nbytes % 4096 == 0:
             return _array_tofile_chunked(write, array, OSX_WRITE_LIMIT)
         return _array_tofile_simple(fd, write, array)
-elif sys.platform.startswith('win'):  # pragma: no cover
+elif sys.platform.startswith('win'):
     def _array_tofile(fd, write, array):
         WIN_WRITE_LIMIT = 2 ** 30
         return _array_tofile_chunked(write, array, WIN_WRITE_LIMIT)
```
jdavies-st, line 117: `OSX_WRITE_LIMIT` and `WIN_WRITE_LIMIT` are not really global variables here, so should probably be lower case.  Or you could make them global variables.  They're identical.

asdf/generic_io.py
```diff
@@ -101,8 +101,7 @@ def _array_tofile_chunked(write, array, chunksize):  # pragma: no cover
 def _array_tofile_simple(fd, write, array):
     return write(array.data)
 
-
-if sys.platform == 'darwin':  # pragma: no cover
+if sys.platform == 'darwin':
```
jdavies-st, line 104: Odd to have this logic at the module level.  Wouldn't it be clearer to have the logic in the calling function?  Or are there more than one calling functions?

asdf/generic_io.py
```diff
@@ -370,7 +369,20 @@ def write(self, content):
...
+        ----------
+        array : np.ndarray
+            Must be 1D contiguous.
+        """
+        if len(array.shape) != 1 or not array.flags.contiguous:
+            raise ValueError("Requires 1D contiguous array.")
+
+        _array_tofile(None, self.write, array)
 
     def seek(self, offset, whence=0):
```
jdavies-st, line 387: This is the first time I've seen `whence` as a kwarg.  I like.  Though it's not really clear from the docstring what it does.

asdf/tags/core/ndarray.py
```diff
@@ -408,32 +411,56 @@ def reserve_blocks(cls, data, ctx):
 
     @classmethod
     def to_tree(cls, data, ctx):
+        # The ndarray-1.0.0 schema does not permit 0 valued strides.
+        # Perhaps we'll want to allow this someday, to efficiently
+        # represent an array of all the same value.
```
jdavies-st, line 416: Isn't the `ndarray-1.0.0` invalid?  What about `ndarray-1.1.0`? I.e. `test_example2`.

asdf/tags/core/ndarray.py
```diff
@@ -408,32 +411,56 @@ def reserve_blocks(cls, data, ctx):
...
+        if block.array_storage == "fits":
+            # Views over arrays stored in FITS files have some idiosyncracies.
+            # astropy.io.fits always writes arrays C-contiguous with big-endian
+            # byte order, whereas asdf preserves the "contiguousity" and byte order
+            # of the base array.
+            if (block.data.shape != data.shape or
+                block.data.dtype.itemsize != data.dtype.itemsize or
+                block.data.ctypes.data != data.ctypes.data or
+                block.data.strides != data.strides):
+                raise ValueError("Views over FITS arrays must have identical memory layout")
```
jdavies-st, line 433: For a user who doesn't know what `memory layout` means, it might be good to say something like "Quit trying to serialize a numpy view!"  in the error message, with a link to the numpy docs about the differences between arrays and views of an array.

https://numpy.org/doc/stable/glossary.html#term-view

jdavies-st (approved):
Looks good.  I didn't follow everything, but the tests are clear.  And they pass.  And the `jwst` regtests that use `numpy_dtype_to_asdf_datatype()` in datamodels pass too.  So 👍 

Just some minor comments and questions below.

asdf/fits_embed.py
```diff
@@ -46,13 +46,6 @@ def readonly(self):
     def array_storage(self):
         return 'fits'
 
-    def override_byteorder(self, byteorder):
-        # FITS data is always stored in big-endian byte order.
-        # The data array may not report big-endian, but we want
-        # the value written to the tree to match the actual
-        # byte order on disk.
-        return 'big'
```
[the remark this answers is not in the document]
eslavich: FITS always writes arrays in big-endian byte order, so we needed a mechanism to tell NDArrayType to ignore the byte order reported by the base array.  This logic has been moved inside NDArrayType itself.

asdf/generic_io.py
eslavich: I don't know for sure, but I suspect whoever wrote this was trying to be efficient by doing the check once at import time instead of on every call.

asdf/generic_io.py
eslavich: That's funny, I didn't notice that.  Seems to be a [bit of a tradition](https://www.gnu.org/software/libc/manual/html_node/File-Positioning.html#index-fseek).

asdf/generic_io.py
eslavich: That is a bit strange.  I'll move those outside the functions.

asdf/tags/core/ndarray.py
eslavich: We haven't created `ndarray-1.1.0` yet, have we?  I don't see it in asdf-standard...

asdf/tags/core/ndarray.py
eslavich: Good point, that message is a bit cryptic.  I'll rework it.

asdf/generic_io.py
eslavich: Actually, I'm going to tackle this in a separate PR.  These workarounds seem to no longer be necessary, but this PR will be included in an asdf 2.7.3 and I'd rather drop them in 2.8.

eslavich: https://github.com/asdf-format/asdf/issues/932

asdf/generic_io.py
jdavies-st: 😅 Shows you how much I paid attention in my C programming!

asdf/tags/core/ndarray.py
jdavies-st: Doh!  Nevermind.

I guess this would be a candidate feature of 1.1.0.  =)

asdf/fits_embed.py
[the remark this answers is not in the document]
jdavies-st: Glad to see the logic moved inside the class itself.

kmacdonald-stsci (approved):

Issue #926, opened 2021-02-10
ASDF extensions for custom compression types?


lgarrison:
I have ASDF files where the binary data is compressed with a custom compression, identified by the `'blsc'` tag in the binary header.  The current way I'm recommending my users read these files is by downloading my fork of ASDF that knows how to do this decompression.  But this is definitely a fragile solution; if a user accidentally installs mainline ASDF, perhaps because my fork doesn't satisfy some ASDF version requirement for another project, then they can't read my files and have to go find my fork again.

I started reading about the ASDF extension mechanism, and I was wondering if that might be a path to a more robust solution.  In particular, I see that extension classes can be registered with setuptools entry points to inform ASDF how to process custom data types, and I was wondering if custom compression types might be registered in the same way.  Does that make sense in the context of ASDF extensions, or would that be orthogonal to their purpose?

Thanks!

eslavich:
I think it's a great idea to have a entry points plugin for additional compression types.  That could be added to the existing Extension class or a new entry point dedicated to compression providers.

lgarrison:
Thanks; I'd be happy to work on a PR for this.  My particular use-case is a little strange, though, because the files are already written without any reference to such an extension, so there's no `history: extensions: - lgarrison_blosc` or whatever in the YAML portion of the files.  The files are too numerous to rewrite, unfortunately.  Perhaps a way forward would be to provide a runtime flag to opt-in to "eager" or "unsafe" use of compression extensions that just rely on the compression name.  For example, in my Python package through which users will read my files, I could write:
```python
import asdf
asdf.compression.allow_unchecked_extensions()
```

Then, the future file loads will be checked against the list of compression types registered via the entry points. Any matching names will be decompressed via that extension.  If `allow_unchecked_extensions()` is not used, then any decompression must be done by an extension that is explicitly listed in the YAML of that file (which I think is how the type extension mechanism works, if I understand correctly).

eslavich:
Maybe it would be generally safe to rely on the compression name if there aren't any conflicts among the installed extensions?  If not, we do have a new [configuration feature](https://github.com/asdf-format/asdf/blob/master/asdf/config.py#L208) that would be an appropriate place to add an `allow_unchecked_extensions` flag.

[...]

PR #931, opened 2021-02-23
Support for compression extensions

lgarrison:
This PR adds support for externally-defined compression via the new-style ASDF Extension mechanism, as discussed in #926.  Third-party packages can register an extension with a setuptools entry point as usual, and include `compressors` and `decompressors` properties to reference `Compressor` and `Decompressor` subclasses.  I'll use the example of an extension that wraps Python's lzma compression (also used in the tests).

In `setup.py`, nothing is changed:
```python
setup(
   ...
    entry_points={'asdf.extensions':['mypkg = mypkg.asdf:LzmaExtension']}
)
```
The `LzmaExtension` subclasses `asdf.extension.Extension` with the `compressors` and `decompressors` properties:
```python
from asdf.extension import Extension

class LzmaExtension(Extension):
    @property
    def extension_uri(self):
        return "http://mypkg.org/extensions/lzma-1.0"
    
    @property
    def compressors(self):
        return [LzmaCompressor]
    
    @property
    def decompressors(self):
        return [LzmaDecompressor]
 ```
The compressor and decompressor are defined by subclassing `asdf.extension.Compressor/Decompressor`:

```python
import lzma

from asdf.extension import Compressor, Decompressor

class LzmaCompressor(Compressor):
    def __init__(self, preset=None):
        self._compressor = lzma.LZMACompressor(preset=preset)
        
    def compress(self, data):
        comp = self._compressor.compress(data)
        return comp
    
    def flush(self):
        return self._compressor.flush()
    
    @property
    def labels(self):
        return ['lzma']

class LzmaDecompressor(Decompressor):
    def __init__(self):
        self._decompressor = lzma.LZMADecompressor()
    
    def decompress(self, data):
        decomp = self._decompressor.decompress(data)
        return decomp
    
    @property
    def labels(self):
        return ['lzma']
```

Files that use a compression extension to compress any of their blocks will have the extension recorded in the history:
```yaml
history:
  extensions:
    compression_labels: [lzma]
    extension_class: mypkg.asdf.LzmaExtension
    extension_uri: "http://mypkg.org/extensions/lzma-1.0"
    software: !core/software-1.0.0 {name: mypkg, version: 0.0.1}
```
And the 4-byte `'lzma'` string will be used in the binary block header.

Compression/decompression options, like the compression level (called `preset` by lzma), can be set via `asdf.config`, and will be passed to the `Compressor`/`Decompressor` constructor:
```python
config = asdf.config.get_config()
config.compression_options['lzma'] = {'preset':lzma.PRESET_DEFAULT}
...
class LzmaCompressor(Compressor):
    def __init__(self, preset=None):
...
```
This also works for built-in compression.

Decompressors can supply a `decompress_into(data, out)` method that takes a pre-allocated array into which to write the decompressed data.  Doing so avoids an extra memcpy for the concatenation of decompressed blocks, which especially helps fast decompression.  For the built-in lz4, this improves the speed 50% from 420 MB/s to 625 MB/s on my platform.

I made quite a few guesses about the best way to implement things in the ASDF framework; using `asdf.config` to set compression options is a good example.  I'm happy to iterate on the design!  Thanks for your help thus far.


lgarrison:
As discussed with @eslavich, one of the failures of this implementation is that it doesn't hint what extension might be needed to load an unknown compression type, even if that extension in in the YAML.  Here's a minimal reproducer.  It's probably premature to drill down on the best solution to this specific problem, but we should keep it in mind for any larger refactoring that needs to be done.

<details>

```python
import lzma

import asdf
import numpy as np

from asdf.extension import Extension, Compressor, Decompressor
from asdf.config import config_context

class LzmaCompressor(Compressor):
    def __init__(self, preset=None):
        self._compressor = lzma.LZMACompressor(preset=preset)

    def compress(self, data):
        comp = self._compressor.compress(data)
        return comp

    def flush(self):
        return self._compressor.flush()

    @property
    def labels(self):
        return ['lzma']

class LzmaDecompressor(Decompressor):
    def __init__(self):
        self._decompressor = lzma.LZMADecompressor()

    def decompress(self, data):
        decomp = self._decompressor.decompress(data)
        return decomp

    @property
    def labels(self):
        return ['lzma']

class LzmaExtension(Extension):
    @property
    def extension_uri(self):
        return "http://somewhere.org/extensions/lzma-1.0"

    @property
    def compressors(self):
        return [LzmaCompressor]

    @property
    def decompressors(self):
        return [LzmaDecompressor]


tree = {'data':np.random.rand(128,128)}

with config_context() as config:
    config.add_extension(LzmaExtension())

    asdf.config.get_config().compression_options['lzma'] = {'preset':lzma.PRESET_DEFAULT}
    with asdf.AsdfFile(tree=tree) as af:
        af.write_to('test.asdf', all_array_compression='lzma')
        
    asdf.open('test.asdf')  # works
asdf.open('test.asdf')  # "ValueError: Supported compression types are: ['blsc', 'zlib', 'bzp2', 'lz4', 'input'], not 'lzma'"
# No warning about missing extension
```
</details>

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,84 @@
...
+
+Note that this interface has similar patterns to Converter.  This
+interface is designed for compression and decompression of ASDF
+binary array blocks, while Converter is designed for serialization
+of custom Python types into the YAML tree.
+"""
+
+import abc
+
+class Compressor(abc.ABC):
```
eslavich, line 15: The `asdf.compression` module also splits these into encoder vs decoder, but I wonder if that's necessary?  Maybe there are rare cases where someone would want to implement just one of decompress or compress, but that could be handled by implementing e.g. a compress method that raises `NotImplementedError`.

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,84 @@
...
+    @classmethod
+    def __subclasshook__(cls, C):
+        if cls is Compressor:
+            return (hasattr(C, "labels") and
+                    hasattr(C, "compress"))
+        return NotImplemented # pragma: no cover
+
+
+    @abc.abstractproperty
+    def labels(self):
```
eslavich, line 33: Do you have a use case in mind for multiple labels?  If that's uncommon, it might be reasonable to force users to create one instance of the class per label.

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,84 @@
...
+    and the `decompress()` or `decompress_into()`
+    methods.  May also provide a constructor that takes kwargs that are
+    set via `asdf.get_config().decompression_options[label] = kwargs`.
+    """
+    @classmethod
+    def __subclasshook__(cls, C):
+        # TODO: is this the right way to enforce that a subclass define `decompress()` or `decompress_into()`?
+        if cls is Compressor:
+            return (hasattr(C, "labels") and
+                    ( hasattr(C, "decompress") or
```
eslavich, line 68: Another option (instead of separate decompress/decompress_into methods) is to mimic the [numcodecs interface](https://numcodecs.readthedocs.io/en/latest/abc.html#numcodecs.abc.Codec.decode) and make the destination array an optional keyword argument to decompress.  That has the advantage of allowing us to make decompress an `@abc.abstractmethod`.

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,84 @@
...
+        Returns
+        -------
+        iterable of str
+            str labels handled by this class
+        """
+        pass # pragma: no cover
+
+
+    @abc.abstractmethod
+    def compress(self, data):
```
eslavich, line 47: I like the idea of configuring the compressors using`AsdfConfig`, but it would be nice if users in a console could interactively discover the available options without having to dig into the details of the `Compressor` subclass.  Maybe we should add a method here like `get_default_config` that returns the keys and default values (and maybe doc strings) for the compressor's config options.  Then we can use that method in AsdfConfig to generate structure that can be tab-completed in ipython.

asdf/extension/_compressor.py
eslavich: Speaking of instances, did you have in mind that the `Extension` instance would return instances of `Compressor` subclasses or the subclasses themselves?  The `Extension.converters` property returns instances and I prefer that they work the same way if possible.  Do you foresee a need for `Compressor` instances that maintain some kind of internal state between calls to `compress` or `decompress`?

eslavich (commented):
I've just looked at the new ABCs so far, left some questions and suggestions

asdf/extension/_compressor.py
lgarrison: Yeah, I initially started the implementation with a unified `Compressor` with `compress` and `decompress` methods.  I ran into problems in two places: the `Compressor` constructor, and methods that could be shared between the two, like `flush`.  In the constructor, you have to jump through hoops to split arguments into those passed to the underlying compression object and those passed to the decompression object (like `lzma.LZMA[De]Compressor`).  It's possible, but not particularly elegant.

I started doing that, but then I encountered the problem where `flush` can be present in both encoders and decoders, so you'd probably have to split that method into `flush_compress` and `flush_decompress`.  Again, it's possible, but I started to smell a pattern and so decided to try implementing them separately like the built-in compression had done.  I think it worked well for the `compression` module, and it also lets an `Extension` know what labels it can compress versus decompress, in the rare cases where those diverge.

But the extra surface area of `decompressors` in the `Extension` and elsewhere in the code is one of my least favorite parts of this PR.  I'm happy to work through the above issues if we think a unified `Compressor` would be better.

asdf/extension/_compressor.py
lgarrison: My thought was for label aliases.  I think @perrygreenfield had mentioned such aliasing, but that was probably at the YAML level (or maybe in the new JSON block header, if that's still the plan).  Otherwise, I agree restricting a `Compressor` to one label might be preferable.

I think I'm okay with the `Extension` returning instances, although the `Compressor` objects are currently designed to take arguments from the config in their constructors.  So that logic would have to be moved to the `Extension` and either implemented by the user writing the extension, or wrapped in some way.  I don't foresee any particular need to maintain state between calls, except perhaps for tracking performance metrics, which can be done by other means.

asdf/extension/_compressor.py
lgarrison: I did think about that, but the presence of `decompress_into` lets the `decompress` function know whether to provide an `out` arg and whether the return value is bytes to be concatenated or a number of bytes written.  It's possible to require that compressors handle the `out` arg with simple concatenation if they don't natively support writing to a user-supplied array, but then every implementation needs to repeat that logic.  I guess we could detect whether the return type is an integer and assume successful writing in-place if so.  Another option would be a field like `Decompressor.supports_append`.

asdf/extension/_compressor.py
eslavich: I think there's something that I don't understand yet, why does there need to be a `flush` method?  Wouldn't `compress` and `decompress` carry through the operation from start to finish, including any necessary flushes?

I was imagining this library would pass the entirety of each array to `compress` and receive back a new array containing the entirety of the compressed data.  Maybe that doesn't work for memory management reasons?

asdf/extension/_compressor.py
eslavich: This may be another symptom of my more general misunderstanding of how these methods will be used, but does it work to always return the decompressed array?  Something like this:

```python
def decompress(self, data, out=None):
  """
  Decompress block data.
  
  Parameters
  ----------
  data : np.ndarray
    Compressed bytes.
  out : np.ndarray, optional
    Optional pre-allocated output array.  If present, must be exactly the size required
    to store the decompressed data.

  Returns
  -------
  np.ndarray
    Decompressed data.  This will be the same array object as `out` if that parameter
    was provided.
  """
  pass
```
Then if an implementation uses a compressor that doesn't natively support writing to a user-supplied array, it just has to do something like this:
```python
def decompress(self, data, out=None):
  uncooperative_compressor = UncooperativeCompressor()
  result = uncooperative_compressor.decompress(data)
  if out is not None:
    out[:] = result
    return out
  else:
    return result
```

lgarrison: That almost works, but you need a way to return the number of decompressed bytes, because you don't know the size of the decompressed block.  All you know is the total size of all decompressed blocks, because that's the logical array size that ASDF tracks in the tree.  One way to return the decompressed size is to advance the `out` "pointer", e.g. `return out[len(result):]`; another is to just return an int.  So it can work, it just requires that each decompressor implement this logic.

Unfortunately, this would also require wrapping the `zlib` and `bz2` built-in compressors.  Currently, the API is simple enough that it's compatible with `zlib.decompressobj` and `bz2.BZ2Decompressor`, but neither of them support the `out` parameter.

asdf/extension/_compressor.py
lgarrison: I like the tab-complete idea a lot.  Did you have an idea of how to register the tab-complete fields of a `Compressor` with `AsdfConfig`?  If `Compressor.get_default_config()` returns a key->(val,docstring) dict, then I guess when an `Extension` is loaded, it needs to register this dict with the config somehow.  Then the config can build an object with these fields and store it under `config.compression_options` such that
```
lzma_conf = config.compression_options['lzma']
lzma_conf.<TAB>
```
will prompt tab-completion.

Is `config.add_extension()` a good place to hook this logic?  Is it called for all extensions?

asdf/extension/_compressor.py
eslavich: The ASDF block header actually does include the number of decompressed bytes in a field, see `data_size` here:

https://asdf-standard.readthedocs.io/en/1.5.0/file_layout.html#block-header

I'd like to wrap zlib and bz2 in this interface anyway, so that everything is nice an uniform.  I don't see that as much of a downside since the wrapper will be thin.

asdf/extension/_compressor.py
eslavich: Hmm... good question.  One option is to dynamically generate config classes for each compressor:

```python
label = "lzma"
config_dict = {
    "preset": (6, "LZMA compression preset.  Integer from 0-9."),
    "foo": ("bar", "Some other LZMA config parameter.")
}

cls = type(
    f"{label.upper()}CompressorConfig",
    (object,),
    {"__doc__": f"Configuration options for the {label.upper()} compressor."}
)

for name, (default_value, docstring) in config_dict.items():
    setattr(cls, name, property((lambda attr: lambda self: getattr(self, attr))(f"_{name}")))
    setattr(cls, name, getattr(cls, name).setter((lambda attr: lambda self, value: setattr(self, attr, value))(f"_{name}")))
    getattr(cls, name).__doc__ = docstring

def generate_init(config_dict):
    def init(self):
        for name, (default_value, docstring) in config_dict.items():
            setattr(self, f"_{name}", default_value)
    return init
cls.__init__ = generate_init(config_dict)
```
An instance of that class seems to behave like we'd want:
```python
In [82]: config = cls()

In [83]: config.
                 foo
                 preset
 
In [84]: ?config.preset
Type:        property
String form: <property object at 0x7ff4f60e2cc0>
Docstring:   LZMA compression preset.  Integer from 0-9.

In [85]: config.preset
Out[85]: 6

In [86]: config.preset = 9

In [87]: config.preset
Out[87]: 9
```

But now I'm wondering if instead we should ask the author of the compressor/extension provide an implementation of a config class.  That way they could e.g. perform validation in the setter, automatically adjust parameters with tied values, etc.  I can see that being a valuable general feature of `Extension`, not just for compression config options.

asdf/extension/_compressor.py
lgarrison: Yes, that's a good point.  On the compression side, there's probably not much use for `flush` since we always have the whole array data.  Even if we didn't (e.g. data was streaming in), we could do some buffering before compression.  The built-in compressors rely on `flush` being called, but since we're going to wrap those anyway, we can force a `flush` after each `compress` call, which shouldn't be too bad because our block size is big.

On the decompression side, the algorithms using block decompression (like lz4) should be fine without `flush`, as they will self-flush at the end of a block.  However, I think this will force us to read the entire array data before calling decompression like `zlib` that is not wrapped in a blocking scheme.  We can't call [`flush`](https://docs.python.org/3/library/zlib.html#zlib.Decompress.flush) after every file block because that renders the decompression object unusable.  So, we could permit a final `flush` call, but just for decompression instead of compression.  Or, we could force non-blocked decompression schemes to read the full array data, at the cost of some extra memory.

asdf/extension/_compressor.py
lgarrison: Sorry, I was referring to compression blocks, not ASDF blocks.  There can be multiple compression blocks per ASDF block (compression blocks are the ones sized to `DEFAULT_BLOCK_SIZE = 1 << 22` in `compression.py`), and the decompressed size of each of those isn't stored anywhere.  All you have is the total size (the `data_size`), which you can use to pre-allocate an array that the decompressed compression blocks can fill.

But if we're wrapping the built-in compressions anyway, then I think the int-return scheme is perfectly acceptable, as long as we document this API for extension implementers.

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Returns compressed data.
+        """
+        pass # pragma: no cover
```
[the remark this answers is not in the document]
lgarrison: This `compress` method is designed to be called from `compression.compress`, which chunks the data into compression blocks and feeds them one at a time into this function, writing the result out to disk (or to `fd`, a `GenericIO` object).  So it should support mmap'd huge arrays without additional effort (worth checking though).  I can still see a use for an `out` parameter, to avoid a memcpy for compressors that support it, but my sense is that this particular optimization is not likely to be the bottleneck for compression, which is generally much slower than decompression.

asdf/extension/_compressor.py
eslavich: Oh, ha, I didn't even know we did that.  Does that seem strange to you, that ASDF is only compressing 4 MB at a time?  I would think that we'd want to let the individual compression algorithms tune themselves.  The LZ4 compressor is writing out a header with each block, which sounds more like N compressed files cat'ed together?

lgarrison: It's a little strange, but there are some pros, I think.  A pragmatic one is that in the case of Blosc and lz4 (at least the `lz4.block` API), there are 32-bit size limits that require some external blocking to deal with arrays larger than 2 GB.  But even in the absence of that problem, I think blocking enables efficient overlap of decompression and IO in a very general way.  One thread can read blocks and pass them to a decompression thread, which can always decompress the data because it is completely self-contained within that 4 MB.  The compression doesn't need to have any special "streaming" capability.  It's obviously not something being used right now, but might be very nifty in the future.

asdf/extension/_compressor.py
lgarrison: I think I'm still wrestling with whether an extra layer of abstraction—a Config object, either automatically generated or supplied by the user—is likely to be helpful, or just extra boilerplate.  The tab-completion is nice, for sure, but this requires an extension to explicitly implement every arg supported by the underlying compression that an end user might want to use.  It's likely to be a lot of pass-through getters and setters, and copy-pasted docstrings from the underlying compressor.  It just seems verbose and fragile.

If we did go the dict-only approach, there's still some chance for validation in the `Compressor` constructor. You lose tab-completion and individual docstrings for the settings, but you also don't have to explicitly enumerate all settings and can point the user to the docs of the underlying compressor.

eslavich: Hmm... good point.  Trying to keep up with multiple versions of the same compression package with different options would make us crazy.  I do still like the idea of a general `Extension` feature that allows the extension to register a config class, but maybe in this case it would be something simple like this:

```python
class LzmaExtensionConfig:
  def __init__(self):
    self._compressor_options = {}
  
  @property
  def compressor_options(self):
    return self._compressor_options
    
  @compressor_options.setter
  def compressor_options(self, value):
     """
     Additional keyword arguments passed to `lzma.LZMACompressor`.  See
     https://docs.python.org/3/library/lzma.html for details.
     """
     self._compressor_options = value

class LzmaExtension:
  config_class = LzmaExtensionConfig
```

Then we add that to `AsdfConfig` somewhere reasonable so that users can do something like:

```python
asdf.get_config().extensions.<something>.compressor_options = {"preset": 9}
```

How does that strike you?  I don't know what `<something>` should be, the only reliably unique key to an extension is its URI, but that's a lot to type.

lgarrison: I like that.  For the `<something>`, what about assigning a "short name" to each extension from the name of the entry point registration?  E.g. `lzma_ext` in this example:
```python
setup(
   ...
    entry_points={'asdf.extensions':['lzma_ext = mypkg.asdf:LzmaExtension']}
)
```
Short-name conflicts could be warned about and require a URI lookup.

But just to check, `asdf.get_config().extensions.lzma_ext` is a `LzmaExtension` object, not the `LzmaExtensionConfig`, right?  So I guess setting the config would require another level of indirection, like `asdf.get_config().extensions.lzma_ext.config`.  That's getting pretty verbose.  Maybe it could be shortened to `asdf.get_extension_config('lzma_ext')`?

Similarly, maybe the casual user won't care about which extension the compression came from, and should just be able to type: `asdf.get_compression_config('lzma')`.  I think it's just a question of API surface area versus verbosity.

lgarrison: And formally, if a whole extension has one `compressor_config` associated with it, then we probably need to limit extensions to one compressor.  So packages that want to define multiple compressions need to define and register multiple extensions.  That's not the end of the world, just something worth thinking about.

eslavich: Oops, I forgot that `extensions` was already taken.  That hasn't been released yet, and was really only intended for use within the library.  We could change it to a `get_extensions` method instead to make access to the extension configs more convenient.

I was thinking the extension's config would be free-form.  It could have foo_compressor_config and bar_compressor_config properties if it wanted to, or any other name.  Then inside of the `FooCompressor` class the implementation would call `asdf.get_config().extensions.lzma.foo_compressor_config` to get the settings.

asdf/extension/_compressor.py
eslavich: My sentiment is that we don't want to miss out on possible optimizations that the underlying compression library may implement (screaming fast parallel compression in a C extension, for example).  IMO it's worth trading additional complexity in the `asdf.extension.Compressor` implementations to get that.  @perrygreenfield and @jdavies-st what do you think?

lgarrison: This PR does allow tuning of the compression block size via the compression options mechanism we've been working on.  Maybe a value of `compression_block_size=None` or `NO_BLOCK` can force the use of one block?

lgarrison:
Here's a work-in-progress commit showing some of the changes we talked about.  I removed `Decompressors`, thus merging compression and decompression into the `compress` and `decompress` methods of `Compressor`.  The `flush` method now only applies to decompression; compression must be stateless (or at least self-flushing).  The `ZlibCompressor` in `compression.py` is a good example of how this works; note that the object never constructs a persistent compressor like it does a decompressor.

The calling functions, `asdf.compression.compress/decompress`, now assume `out` kwarg support and leave any blocking to the `Compressor`, so they're a bit simpler.  I did not reimplement blocking for the built-in compressors, although I will for lz4, since it's required for that to work with large arrays.

I haven't tackled the configuration options yet; still thinking on that.  If the options are only accessed through extensions, then I'm not sure what the best way to pass options to built-in compressors is, unless they should have an extension too.

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,59 @@
...
```
jdavies-st, line 51: I assume the `compress` method is not require, hence `pass`?  If it is required, maybe `raise NotImplementedError` might be appropriate?

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,59 @@
...
+    
+    
+    @abc.abstractmethod
+    def decompress(self, data, out=None):
+        """
+        Returns decompressed data.
+        """
+        pass # pragma: no cover
```
jdavies-st, line 59: Same comment here.  Is `decompress()` required to be defined for a Compression extension?

asdf/tests/test_compression.py
```diff
@@ -67,7 +70,7 @@ def _roundtrip(tmpdir, tree, compression=None,
     with asdf.open(generic_io.InputStream(buff), **read_options) as ff:
         helpers.assert_tree_match(tree, ff.tree)
 
-    return ff
+    return tmpfile
```
jdavies-st, line 73: Is there a reason this now returns the path to the file instead of the tree itself?  I see that does not affect the legacy tests as this `_roundtrip` function is essentially a wrapper for the `assert_tree_match` assertion helper.

Perhaps it shouldn't return anything and be formalized as an assertion helper itself?

lgarrison: I think this version of the tests just uses the filename to open the file again to check that the compression extension shows up in the history (`test_compression_with_extension()`).  The old `return ff` was actually returning a closed ASDF object, so the history couldn't be queried.  I didn't want to return an open ASDF object; we could instead return the history, or nothing, and make the history check separate.

jdavies-st: Ah, I see now.  Thanks.  Yes, that's an improvement!  😄

lgarrison:
Here's some work on the configuration side of things, using the `config_class` in the `Extension` as suggested.  `test_compression_with_extension()` has an example of how such a `config_class` is set up, and the user accesses the options dict through `get_config().[de]compression(label)`.  `compression.py` accesses the options in the same way.  The `AsdfConfig` initializes the built-in extension configs in `_init_builtin_compression_options()`.

This doesn't offer much in the way of tab-completion, but we may be able to massage that if we like this overall structure.

asdf/extension/_compressor.py
eslavich: I've been thinking about decompression a little more, and I think it may be reasonable to read the entirety of the compressed data before passing it to the plugin's `decompress` method.  This library memory-maps arrays by default, so under normal conditions we could do this in `asdf.compression.decompress`:

```python
compressed_data = fd.read_into_array(used_size)
```

without requiring any additional program memory.

lgarrison: I think I am always suspicious of memory-mapped IO in performance-sensitive contexts because it defers the IO into the bowels of the computation, so the computation is constantly being interrupted waiting for IO.  And the IO is only being triggered sporadically when the computation needs the next section of data—except for read-ahead, which very well may eliminate this concern.  I can do some tests.

lgarrison: I did a standalone experiment where I wrote 1 GB of Blosc-compressed random-float data to a network file system from one node, then went to another node and timed the IO + decompression of the data using memmap.  Then I wrote another 1 GB in a new file from the first node, and went back to the second node and timed IO + decompression using explicit IO, no memmap.  The latter was 2.5x faster than the former.

In detail, here's what I was timing:
with memmap:
```python
%timeit -n1 -r1 mcomp = np.memmap('/mnt/home/lgarrison/ceph/test1.dat'); decomp = blosc.decompress(mcomp)
# 6.47 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```
no memmap:
```python
%timeit -n1 -r1 comp = open('/mnt/home/lgarrison/ceph/test2.dat', 'rb').read(); decomp = blosc.decompress(comp)
# 2.28 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

I repeated the experiment a dozen times, with consistent results.  Interestingly, when I increased the number of Blosc threads to 2, the memmap time decreased to 4 seconds; at 4 threads, 2.5 seconds; at 8 threads, 1.6 seconds (faster than explicit IO).  So I assume memmap is effectively parallelizing the IO, which results in a speedup, at least on a fast network file system.  The actual decompression (no IO) takes 0.8 seconds at 1 threads and 0.1 seconds at 8 threads, so we're IO dominated in all these tests.

So... I think these results are consistent with memmap's IO pattern being suboptimal for compression.  But on the flip side, using many threads can be even faster than explicit IO.  The explicit IO could be parallelized too, although memmap gives it for "free".  Overall, these results suggest to me that we should not rely on memmapped IO here, although I don't want to extrapolate to all systems based on my platform.

Here's the writing code if you want to repeat the test yourself.
<details>

```python
import asdf
import numpy as np
import blosc

def w(fn):
    rng = np.random.default_rng()
    a = rng.random(1*(1<<30)//4, dtype=np.float32)
    comp = blosc.compress(a.tobytes(), typesize=4, cname='zstd', clevel=1, shuffle=blosc.BITSHUFFLE)
    with open(f'/mnt/home/lgarrison/ceph/{fn}', 'wb') as fp:
        fp.write(comp)
```
</details>

The reason I used one node to write and one node to read was to avoid the file still being in the page cache from the writing.  I'm not on machines where I have root, so I can't manually drop the page cache.  It's possible that memmap performance is different somehow on a network system than on a local disk, but I also think a cold read from a network file system is a common ASDF use case.

perrygreenfield: That's interesting. Thanks for doing this. Let us mull this over for a day or so. But I think it shows memmap has its issues.

eslavich: Wow, okay, yeah, that seems conclusive.  The two other options I can think of for avoiding multiple calls to decompress are 1) pass in a file-like object with a `read` method or 2) pass in a generator.  The file interface has the advantage of allowing the compressor to select its own block size, but we'd need to either wrap the real file so that it stops returning bytes once the (ASDF) block is consumed, or pass the limit as an additional argument.  Do either of those options strike you as a good idea?

Also let me know if you don't think a single call to decompress is a virtue worth pursuing.  I like the simplicity of a stateless compressor but I realize that's an aesthetic statement about an interface that end users will likely never see.

asdf/extension/_compressor.py
eslavich: @perrygreenfield has a use case for tweaking compression parameters on an individual array basis.  His example is lossy compression where the appropriate parameter values depend on the level of noise in a given array.  Given that, I wonder if our best option is to make the config an argument to the `AsdfFile.set_array_compression` method.  Something would also need to be done for the `all_array_compression` argument to `AsdfFile.write_to` and `AsdfFile.update`.

lgarrison: Now that you mention it, I have the same use-case for Blosc, as one might want to manually set the typewidth or bit-shuffle versus byte-shuffle, depending on the individual array.  I agree that a `config` argument to `set_array_compression` is probably the way to go.

Do we need a way to pass options for decompression, like number of threads?  For blosc, most of the time just a global `import blosc; blosc.set_nthreads(n)` would be acceptable, so maybe we don't need a full decompression options mechanism.

eslavich: That's a good question, but yeah, maybe we defer decompression config until someone needs it?

lgarrison: I think that makes sense.  I'll put together a version that removes the `AsdfConfig` additions and uses `set_array_compression` args instead.

lgarrison:
As discussed, this version doesn't use `AsdfConfig` for the compression options, or get into the weeds of extension configuration options.  Instead, the syntax is one of the following:
```python
af.set_array_compression(arr, 'zlib', level=1)  # first way
af.write_to(fn, all_array_compression='zlib', compression_kwargs=dict(level=1))  # second way
```

I did change the decompression to accept the `read_blocks` generator, obviating the need for `flush`.  My sense is that matching the IO block size exactly to the compression block size probably doesn't matter as much as just having a big block size, so I added an `io_block_size` to the `AsdfConfig` to allow this tuning.  This did involve some semi-invasive changes to `generic_io.py`, so let me know if anything is awry.

I also wouldn't mind passing a file handle to the compressors to allow matching the exact block size, but I didn't see a particularly elegant way to achieve this in the `generic_io` framework.

Note there's no default config or tab-completion mechanism in this version, but that may be an acceptable tradeoff, at least in the interest of finishing this PR!

lgarrison:
Hmmm... CI seems to be hanging.  Probably something to do with the block size changes.  Doesn't hang on my system, making debugging a little harder...

eslavich:
I'm able to reproduce the hang on my laptop, I'll see if I can figure out what's going on

eslavich:
The get_config() function hasn't been imported when called here:

https://github.com/lgarrison/asdf/blob/compression-extension/asdf/generic_io.py#L1083

That is causing asdf.open to raise an error here:

https://github.com/lgarrison/asdf/blob/compression-extension/asdf/tests/helpers.py#L231

and for some reason under those conditions the statement in the finally block hangs.  I wonder if we're being irresponsible about our http connection in generic_io.py.

lgarrison:
Great, thanks, that saved me a lot of time!  I guess I needed `--remote-data` to see the hang locally.

Looking at the way `block_size` is used in that function, I think it's overkill to use the config there.  I'll just revert the change.

More generally, `HTTPConnection` does need a block size, but my refactoring of the block size logic into `GenericFile` doesn't get inherited because `HTTPConnection` doesn't call the super constructor.  I guess that's intentional?  I'll probably just let it use `io.DEFAULT_BUFFER_SIZE` for now.

lgarrison:
Now when I run with `--remote-data`, I get a hang locally, but a different one.  It seems to be from this test: https://github.com/lgarrison/asdf/blob/compression-extension/asdf/tests/test_file_format.py#L131

This eventually calls something like the following, which just hangs on my system:
```python
import http.client
sock = http.client.HTTPConnection('0.42.42.42')
sock.connect()
```
Does this timeout on CI?  Or fail immediately from the "bad" IP address?  Maybe my local network handles that address in an odd manner.

eslavich:
> Does this timeout on CI? Or fail immediately from the "bad" IP address? Maybe my local network handles that address in an odd manner.

It does seem to fail immediately on our network and in the CI container.  We've had trouble with that test in the past though, I wonder if we should just remove the bad IP address case.

asdf/compression.py
```diff
@@ -30,118 +34,232 @@ def validate(compression):
...
-                "lz4 library in not installed in your Python environment, "
-                "therefore the block in this ASDF file "
-                "can not be compressed.")
-        return Lz4Compressor(lz4.block)
+        # compression
+        self._mode= compression_kwargs.get('mode','default')
+        self._compression_block_size = compression_kwargs.get('compression_block_size',1<<22)
+
+        # decompression
+        self._size = 0
```
eslavich, line 78: Can these variables be local to the decompress method now that it's handling the entire input in one call?

asdf/compression.py
```diff
@@ -30,118 +34,232 @@ def validate(compression):
...
+    def __init__(self, compression_kwargs=None, decompression_kwargs=None):
+        if compression_kwargs is None:
+            compression_kwargs = {}
+        if decompression_kwargs is None:
+            decompression_kwargs = {}
+            
+        self.compression_kwargs = compression_kwargs.copy()
+        self.decompression_kwargs = decompression_kwargs.copy()
+        
+    def compress(self, data, out=None):
```
eslavich, line 160: Did we drop this `out` argument from compress?

eslavich: (or is the ABC and LZ4Compressor missing an `out`)

asdf/compression.py
```diff
@@ -30,118 +34,232 @@ def validate(compression):
...
+    that can handle this type of compression.
+    `return_extension` can be used to return corresponding extension for bookkeeping purposes.
+    Returns None if no match found.
+    '''
+    # TODO: in ASDF 3, this will be done by the ExtensionManager
+    extensions = get_config().extensions
+
+    for ext in extensions:
+        for comp in ext.compressors:
+            # TODO: slightly unfortunate to have to construct the object to get the labels
```
eslavich, line 218: I mentioned this up above, but we might want to return the label from a class method

eslavich (commented):
I'm about halfway through reviewing, but I need to take a break so I'm going to yield my first block of comments (groan, sorry).

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @abc.abstractproperty
+    def labels(self):
+        """
+        Get the string labels that this Compressor
+        is able to compress.
+
+        Returns
+        -------
+        iterable of str
```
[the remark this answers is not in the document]
lgarrison: I made the `str` -> `bytes` change, but I did have to sprinkle a few extra `decode('ascii')` calls throughout the code in places where we are printing the compression label or comparing it to user input.  It's not too bad, though.

asdf/extension/_compressor.py
```diff
@@ -0,0 +1,59 @@
...
+    @abc.abstractmethod
+    def compress(self, data):
+        """
+        Returns compressed data.
+        """
+        raise NotImplementedError
+    
+    
+    @abc.abstractmethod
+    def decompress(self, data, out=None):
```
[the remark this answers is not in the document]
lgarrison: Yeah, I agree, let's make `out` a required parameter.

asdf/compression.py
lgarrison: We could go either way, but let's drop `out` for now.

lgarrison:
I am realizing that in the `Compressor` we could do `def compress(self, data, out, **kwargs)` (same for `decompress`), instead of `def __init__(self, compression_kwargs=None, decompression_kwargs=None)`, with the associated constructor boilerplate.  In other words, the `kwargs` could be passed straight to the methods, no need to even construct an object.  Except `lz4` uses the object construction as a chance to `import lz4`, and maybe it's better to do any kwarg parsing once rather than every call to `compress`?  Not sure; simplifying/removing the constructor could be nice.

eslavich:
> I am realizing that in the Compressor we could do def compress(self, data, out, **kwargs)

I like that!  We'd then be able to return Compressor instances from the Extension property and sidestep the problem with label being an instance property.  Or turn everything into class methods I suppose, but in the case of Converter it's been useful to allow extension developers to return instances and might be here as well.

Calling import every time doesn't seem like much of a hit:

```
In [1]: import lz4

In [2]: %timeit import lz4
81.8 ns ± 1.13 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I don't know if argument parsing is expensive for any of the compressors, but if it is they could always construct the object in a method decorated with `@functools.lru_cache`.

lgarrison:
For the `import lz4`, I actually wasn't thinking of performance, but rather checking if `lz4` is installed in a user's environment, i.e. this code: https://github.com/lgarrison/asdf/blob/compression-extension/asdf/compression.py#L63-L69

But if we're using `Compressor` instances and not making everything class methods, then there's no problem keeping that logic in the constructor.

eslavich:
Ah, I understand now.  I imagine too we're going to eventually move lz4 support out to an optional extension, and since the lz4 package will be listed as a requirement in that package's setup.cfg, we won't end up needing to worry about it being installed.

asdf/compression.py
```diff
@@ -221,26 +327,33 @@ def compress(fd, data, compression, block_size=DEFAULT_BLOCK_SIZE):
...
+    encoder = _get_compressor(compression, compression_kwargs=config)
+
+    # Get a contiguous, 1D memoryview of the underlying data, preserving data.itemsize
+    # - contiguous: because we may not want to assume that all compressors can handle arbitrary strides
+    # - 1D: so that len(data) works, not just data.nbytes
+    # - itemsize: should preserve data.itemsize for compressors that want to use the record size
+    # - memoryview: don't incur the expense of a memcpy, such as with tobytes()
+    data = memoryview(data)
+    if not data.contiguous:
+        data = memoryview(data.tobytes())  # make a contiguous copy
```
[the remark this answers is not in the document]
lgarrison: Is this just a statement about non-contiguous memory-mapped arrays?  I don't think a `memoryview` should inherently trigger a read of the full array, just if we force it with `tobytes()`.

The requirement for contiguity is for interoperability with C-style libraries that will assume this under the hood.  It's not strictly a requirement for Python-based libraries, or for particularly clever C libraries.  But it seems safest to deal with it at this level.

asdf/compression.py
```diff
@@ -221,26 +327,33 @@ def compress(fd, data, compression, block_size=DEFAULT_BLOCK_SIZE):
...
-    if hasattr(encoder, "flush"):
-        fd.write(encoder.flush())
+    encoder = _get_compressor(compression, compression_kwargs=config)
+
+    # Get a contiguous, 1D memoryview of the underlying data, preserving data.itemsize
+    # - contiguous: because we may not want to assume that all compressors can handle arbitrary strides
+    # - 1D: so that len(data) works, not just data.nbytes
+    # - itemsize: should preserve data.itemsize for compressors that want to use the record size
+    # - memoryview: don't incur the expense of a memcpy, such as with tobytes()
+    data = memoryview(data)
```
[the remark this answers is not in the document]
lgarrison: I go back and forth on this; I think there are some surprising aspects of `memoryview` that are more transparent with Numpy arrays, like the fact that they can be non-contiguous, multi-dimensional, and `len(mview) != mview.nbytes`.  Some of the casting rules are weird, too, as seen below.

asdf/compression.py
```diff
@@ -221,26 +327,33 @@ def compress(fd, data, compression, block_size=DEFAULT_BLOCK_SIZE):
...
+    if not data.contiguous:
+        data = memoryview(data.tobytes())  # make a contiguous copy
+    data = data.cast('c').cast(data.format)  # we get a 1D array by a cast to byte, then a cast to data.format
```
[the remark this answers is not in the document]
lgarrison: It's just weird casting rules:
```python
data = memoryview(np.arange(16).reshape(4,4))
print(len(data), data.itemsize)  # 4 8
#data = data.cast(data.format)  # TypeError: memoryview: cannot cast between two non-byte formats
data = data.cast('c').cast(data.format)
print(len(data), data.itemsize)  # 16 8  << want 16 8
```

lgarrison:
Is the Compatibility test failure real?  Unfortunately, I can't seem to run it locally due to a strange `BlockingIOError` that occurs during the virtualenv setup.

lgarrison:
The Compatibility test issues a warning due to `compression_labels` in the history (traceback below).  However, the actual error is an `AttributeError`, so there may be something more complicated going on.

<details>

```console
(base) lgarriso@cori16:~/asdf$ /tmp/pytest-of-lgarriso/pytest-4/asdf-2.3.3-env/bin/python ./compatibility_tests/assert_file_correct.py /tmp/pytest-of-lgarriso/pytest-4/test_file_compatibility_asdf_v0/test-current.asdf
/tmp/pytest-of-lgarriso/pytest-4/asdf-2.3.3-env/lib/python3.8/site-packages/asdf/yamlutil.py:279: AsdfConversionWarning: Failed to convert tag:stsci.edu:asdf/core/extension_metadata-1.0.0 to custom type (detail: __init__() got an unexpected keyword argument 'compression_labels'). Using raw Python data structure instead
  warnings.warn("Failed to convert {} to custom type (detail: {}). "
  File "./compatibility_tests/assert_file_correct.py", line 22, in <module>
    main()
  File "./compatibility_tests/assert_file_correct.py", line 18, in main
    assert_file_correct(path)
  File "/global/u1/l/lgarriso/asdf/compatibility_tests/common.py", line 17, in assert_file_correct
    with asdf.open(str(path)) as af:
  File "/tmp/pytest-of-lgarriso/pytest-4/asdf-2.3.3-env/lib/python3.8/site-packages/asdf/asdf.py", line 1345, in open_asdf
    return AsdfFile._open_impl(instance,
  File "/tmp/pytest-of-lgarriso/pytest-4/asdf-2.3.3-env/lib/python3.8/site-packages/asdf/asdf.py", line 704, in _open_impl
    return cls._open_asdf(self, fd, uri=uri, mode=mode,
  File "/tmp/pytest-of-lgarriso/pytest-4/asdf-2.3.3-env/lib/python3.8/site-packages/asdf/asdf.py", line 665, in _open_asdf
    self._check_extensions(tree, strict=strict_extension_check)
  File "/tmp/pytest-of-lgarriso/pytest-4/asdf-2.3.3-env/lib/python3.8/site-packages/asdf/asdf.py", line 172, in _check_extensions
    if extension.extension_class not in self._extension_metadata:
AttributeError: 'TaggedDict' object has no attribute 'extension_class'
```

eslavich:
I think I see what's going on.  Up through asdf 2.6, the deserializer for `ExtensionMetadata` constructed the instance like this:

https://github.com/asdf-format/asdf/blob/2.6.0/asdf/tags/core/__init__.py#L45

which doesn't play nicely with the schema which allows arbitrary additional properties:

https://github.com/asdf-format/asdf-standard/blob/master/schemas/stsci.edu/asdf/core/extension_metadata-1.0.0.yaml

In 2.7.0 we fixed that problem:

https://github.com/asdf-format/asdf/blob/2.7.0/asdf/tags/core/__init__.py#L46

We're not maintaining support for asdf 2.6.x, so I think we should just update the MIN_VERSION constant in test_file_compatibility.py to 2.7.0.  @jdavies-st @perrygreenfield does that seem reasonable?

eslavich:
Alright @lgarrison if you merge in master it should fix the tests

lgarrison:
Great, thanks.  I think I am going to remove `compression_labels` from the history if it is empty though, as it will be for the vast majority of extensions.  It just feels like extra noise.  Probably going to rename it `supported_compression` too, seems more readable.

lgarrison:
I think we're getting close, except ASDF doesn't warn about missing compression extensions.  The reason appears to be that the compression validation happens [here](https://github.com/lgarrison/asdf/blob/compression-extension/asdf/asdf.py#L889-L890) before the extension validation [here](https://github.com/lgarrison/asdf/blob/compression-extension/asdf/asdf.py#L907-L908).

Reproducer:
<details>

```python
import lzma

import asdf
import numpy as np

from asdf.extension import Extension, Compressor
from asdf.config import config_context

class LzmaCompressor(Compressor):
    def compress(self, data, **kwargs):
        comp = lzma.compress(data, **kwargs)
        yield comp

    def decompress(self, blocks, out, **kwargs):
        decompressor = lzma.LZMADecompressor(**kwargs)
        i = 0
        for block in blocks:
            decomp = decompressor.decompress(block)
            out[i:i+len(decomp)] = decomp
            i += len(decomp)
        return i

    @property
    def label(self):
        return b'lzma'

class LzmaExtension(Extension):
    @property
    def extension_uri(self):
        return "asdf://somewhere.org/extensions/lzma-1.0"

    @property
    def compressors(self):
        return [LzmaCompressor()]

tree = {'data':np.random.rand(128,128)}

with config_context() as config:
    config.add_extension(LzmaExtension())

    with asdf.AsdfFile(tree=tree) as af:
        af.write_to('test.asdf', all_array_compression='lzma')
        
    asdf.open('test.asdf')  # works
asdf.open('test.asdf')  # no warning about missing extension
```
</details>

eslavich:
> ASDF doesn't warn about missing compression extensions

Drat!  The ValueError message is helpful enough for now I think, we can file an issue to look into suggesting an extension later.  I'll make a note to do that after this PR is merged.

asdf/tests/test_extension.py
```diff
@@ -110,6 +117,14 @@ def select_tag(self, obj, tags, ctx):
         return "select_tag result"
 
 
+class MinimalCompressor(Compressor):
```
eslavich, line 120: Haha

eslavich (commented):
I noticed a couple of things that we might be able to drop, otherwise this looks ready to go!

asdf/compression.py
```diff
@@ -30,118 +33,192 @@ def validate(compression):
...
+        compression_block_size = kwargs.pop('compression_block_size', 1<<22)
+
+        nelem = compression_block_size // data.itemsize
+        for i in range(0,len(data),nelem):
+            _output = self._api.compress(data[i:i+nelem], **kwargs)
+            header = struct.pack('!I', len(_output))
+            yield header + _output
+
+
+    def decompress(self, blocks, out, **kwargs):
```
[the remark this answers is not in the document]
lgarrison: It's true... a lot of the complexity comes from being fully flexible with regards to the file block size.  I am not sure what can be done though; even letting the decompressor take control of the `fd.read(N)` doesn't guarantee that N bytes will be read; N is just an upper limit.  So I think we're stuck with this for now.

asdf/tests/test_extension.py
lgarrison: `def compress(data): return b''` may be lossy, but it has a great compression ratio!

lgarrison:
Without much evidence, I'm going to blame the test failure on the 5-hour-old Astropy release.  There's some kind of FITS parsing error, kind of hard to make sense of...

eslavich:
Looks like some of our test data we have checked into the repo needs to be regenerated.  I'll file an issue to address that separately.

Time to merge I think!  Thanks a lot for contributing this and for all the work that went into it.  I'm going to be writing up documentation for the new extension API and will include a section on compression, I'll be sure to mention you when I get that PR open.

lgarrison:
Thanks for all your help on this!  I'm happy to help with the high-level documentation as that comes together.  I'm also going to be porting my projects to use this mechanism, so we'll get some "real world" experience from that over the next few months.

CagtayFabry:
Thanks @lgarrison @eslavich for all the ahrd work and adding this !

will certainly try it out soon as well 🚀

PR #936, opened 2021-02-26
Fix FITS array view tests

eslavich:
@kmacdonald-stsci pointed out a flaw in the tests I added to cover views over FITS arrays -- the base array was full of zeros, so the assertions were not checking what I thought they were.  Correcting the tests revealed a new bug (present in 2.7.2, so not caused by the 2.7.3 changes) where views over FITS arrays with a different dtype are not deserialized correctly.  This is an uncommon case but we should fix it eventually, so I added an xfail'ed test and will file an issue as well.

### DIFF
--- a/asdf/tests/test_fits_embed.py
+++ b/asdf/tests/test_fits_embed.py
@@ -12,6 +12,7 @@ from astropy.table import Table
 from jsonschema.exceptions import ValidationError
 
 import asdf
+from asdf.constants import DEFAULT_AUTO_INLINE
 from asdf import get_config
 from asdf import fits_embed
 from asdf import open as asdf_open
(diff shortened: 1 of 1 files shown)


eslavich:
Issue added here: https://github.com/asdf-format/asdf/issues/937

asdf/tests/test_fits_embed.py
```diff
@@ -486,22 +488,47 @@ def test_array_view_compatible_layout(tmp_path):
     """
     file_path = tmp_path / "test.fits"
 
-    data = np.zeros((10, 10), dtype=np.float64)
-    data_view = data[:, :5]
+    data = np.arange(400, dtype=np.float64).reshape(20, 20)
```
kmacdonald-stsci, line 491: If you want to guarantee a size greater than `asdf.constants.DEFAULT_AUTO_INLINE` it might be worth defining the dimensions in terms of the constant.

eslavich: Yeah, that's a better way, it'll maintain itself.  I'll do that.

kmacdonald-stsci (approved):

jdavies-st (approved):

PR #945, opened 2021-03-22
Fix test failures on s390x and add Docker image for future troubleshooting

### LINKED ISSUE #928: s390x asdftool edit test failures
See test failures here.

https://github.com/asdf-format/asdf/runs/1901035329?check_suite_focus=true#step:3:1975

eslavich:
This fixes an asdftool edit test that has been failing on s390x architecture.  The difference there is that the initial byteorder string is "big" and not "little", so replacing it with "medium" causes an extra prompt to confirm that it's okay to rewrite the file.

I also checked in a Dockerfile that I used to troubleshoot the problem with some brief instructions on how to use it.

Resolves #928

### DIFF
--- a/.github/workflows/s390x.yml
+++ b/.github/workflows/s390x.yml
@@ -1,6 +1,7 @@
 name: s390x
 
 on:
+  workflow_dispatch:
   push:
     tags:
       - '*'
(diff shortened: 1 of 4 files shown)


jdavies-st (approved):
🎉 

And thanks for the Dockerfile and documentation to ease future maintenance.

PR #949, opened 2021-03-22
Fix for base arrays with negative strides

### LINKED ISSUE #916: Scipy 1.6.0 generated ndarray can not be restored

eslavich:
Here's my second attempt at a fix for #916.  I noticed that we have a precedent where we copy the view when it has strides unsupported by ASDF (in the existing case, 0 strides).  Since re-computing the view against the base array as serialized is hard, I propose that we do the same here and simply copy the view when its base is non-contiguous.  I also added scipy to our test dependencies so that we can test this using the example in the issue description.

Fixes #916

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,13 @@
 
 - Remove the ``asdf.test`` method and ``asdf.__githash__`` attribute. [#943]
 
+2.7.4 (unreleased)
+------------------
+
+- Fix pytest plugin failure under older versions of pytest. [#934]
+
+- Copy array views when the base array is non-contiguous. [#949]
+
 2.7.3 (2021-02-25)
 ------------------
 
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -417,6 +417,14 @@ class NDArrayType(AsdfType):
         if any(stride == 0 for stride in data.strides):
             data = np.ascontiguousarray(data)
 
+        # The view computations that follow assume that the base array
+        # is contiguous.  If not, we need to make a copy to avoid
+        # writing a nonsense view.
+        base = util.get_array_base(data)
+        if not base.flags.contiguous:
+            data = np.ascontiguousarray(data)
+            base = util.get_array_base(data)
+
         shape = data.shape
 
         block = ctx.blocks.find_or_create_block_for_array(data, ctx)
@@ -447,8 +455,6 @@ class NDArrayType(AsdfType):
         else:
             # Compute the offset relative to the base array and not the
             # block data, in case the block is compressed.
-            base = util.get_array_base(data)
-
             offset = data.ctypes.data - base.ctypes.data
 
             if data.flags.c_contiguous:


CagtayFabry:
IIRC @marscher already committed a fix for the unusual stride layout created by `Rotation` that should make it to the next scipy release (targeted for 1.7)

In that case I am not sure if the test based on `Rotation` can reliably be used to induce the desired behavior in the future.

Do you agree @marscher?

marscher:
The fix in scipy will also be back ported to oder versions, so you're right that it shouldn't be used as a regression test. These types of arrays simply can be produced by creating a reverse view on the memory, e.g. [::-1]

eslavich:
> These types of arrays simply can be produced by creating a reverse view on the memory, e.g. [::-1]

Unfortunately (or rather, fortunately!) existing releases of asdf handle this array just fine:

```python
import asdf
import numpy as np
from numpy.testing import assert_array_equal

array = np.arange(1000)[::-1]
with asdf.AsdfFile({"array": array}) as af:
    af.write_to("test.asdf")

with asdf.open("test.asdf") as af:
    assert_array_equal(af["array"], array)
```

That gives a view with negative strides over a contiguous base array.  The scipy array is special because the base array itself has negative strides, which means that the memory allocated for the base is laid out in reverse order from what the base array needs.  As far as I can tell, that's not possible to achieve in pure Python.

In any case, thanks for the warning about the scipy fix.  Do you think I should remove the test, or forego this PR entirely?  If this situation is considered to be a scipy bug, then maybe it's not something that asdf needs to handle.

marscher:
I think this edge case can only be enforced by using Cythons memoryviews and should be super rare. As long as you do not want to require Cython for your testsuite I suggest it is kind of overkill to support it.

eslavich:
Ok, I'll remove the test, but keep the fix in place.

jdavies-st (approved):
Looks good, but without a test, I can't see what it is fixing.

PR #950, opened 2021-03-23
Revive changelog check

### LINKED ISSUE #820: Revive the changelog bot
It seems to have gotten lost during the transition to asdf-format

eslavich:
This adds a workflow that checks for a CHANGES.rst entry for the current PR.  It also responds to the same label that we used before, `no-changelog-entry-needed`.

Resolves #820

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,19 @@
+name: Changelog
+
+on:
+  pull_request:
+    types: [labeled, unlabeled, opened, synchronize, reopened]
+
+jobs:
+  changelog:
+    name: Confirm changelog entry
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: true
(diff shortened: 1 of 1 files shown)


CagtayFabry:
do you mind if I copy this little gem over to weldx? @eslavich

eslavich:
> do you mind if I copy this little gem over to weldx? @eslavich

Fine by me!

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,19 @@
+name: Changelog
+
+on:
+  pull_request:
+    types: [labeled, unlabeled, opened, synchronize, reopened]
```
jdavies-st, line 5: Clever!  I can change labels, which triggers this job.  🚀

jdavies-st (approved):
Very cool!  We should be using this everywhere.  `jwst` would be a start.  @hbushouse would be happy.

hbushouse:
> Very cool! We should be using this everywhere. `jwst` would be a start. @hbushouse would be happy.

Would save all of us the trouble of having to manually request a change log entry when doing PR reviews. Of course it is now part of the PR template checklist, so that helps too.

PR #951, opened 2021-03-23
Update LICENSE

eslavich:
I noticed that our license included a reference to "Astropy Team" and the year needed updating.  I replaced what we had with the text from https://spdx.org/licenses/BSD-3-Clause.html, filled in the year and copyright holder at the top, and added some extra whitespace.

If this is deemed suitable I'll go ahead and copy it out to the other repos as well.


jdavies-st (commented):
I think we want it to be copyright AURA.  See

https://github.com/spacetelescope/style-guides/blob/master/guides/github-repositories.md#license

Is the DSMO style guide up-to-date @nden?

Otherwise looks good.

jdavies-st (approved):

PR #953, opened 2021-03-23
Use temporary working directory for all tests

### LINKED ISSUE #942: Test suite leaves behind untracked files
The docstring tests create files in the repo root:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	largeval.asdf
	test.asdf
```

Perhaps we can create those in a temporary directory instead?

eslavich:
We apparently have a solution for this that used to work but something changed in the meantime.  It doesn't seem to hurt anything to use a temporary working directory for all tests, so I changed the fixture to do that.

Resolves #942

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -7,9 +7,9 @@ import pytest
 @pytest.fixture(scope="session", autouse=True)
 def temp_cwd():
     original_cwd = os.getcwd()
-    try:
-        with tempfile.TemporaryDirectory() as tmp:
+    with tempfile.TemporaryDirectory() as tmp:
+        try:
             os.chdir(tmp)
             yield
-    finally:
-        os.chdir(original_cwd)
+        finally:
+            os.chdir(original_cwd)
--- a/conftest.py
+++ b/conftest.py
@@ -1,20 +1,15 @@
+import os
+import tempfile
+
 import pytest
 
-@pytest.fixture(autouse=True)
-def _docdir(request):
-    """
-    Make sure that doctests run in a temporary directory so that any files that
-    are created as part of the test get removed automatically.
-    """
-    # Trigger ONLY for doctestplus.
+
+@pytest.fixture(scope="session", autouse=True)
+def temp_cwd():
+    original_cwd = os.getcwd()
     try:
-        doctest_plugin = request.config.pluginmanager.getplugin("doctestplus")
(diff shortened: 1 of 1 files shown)


eslavich:
Hmm, Windows is angry...

jdavies-st:
Using a tempdir for all tests seems like overkill, and also means tests that need input data will now have a hard time finding it because the relative path to the test data might have changed?  Not sure if that is the issue with the Windows tests.

I wrote this pytest autouse fixture, and I suspect pytest broke it in their latest releases.  `astropy` has updated the version there, as have we in `jwst`.

https://github.com/astropy/astropy/blob/aed5138bc5d22a89caa956bcdff641673b43dede/docs/conftest.py#L27-L41

eslavich:
This seems to describe the problem on Windows:

https://bugs.python.org/issue42796

I'll give that a try.  I don't think it's an issue with paths to data, since our tests seem to all reference the data directory relative to `__file__`.

conftest.py
```diff
@@ -1,20 +1,15 @@
...
-        if isinstance(request.node.parent, doctest_plugin._doctest_textfile_item_cls):
-            tmpdir = request.getfixturevalue('tmpdir')
-            with tmpdir.as_cwd():
-                yield
-        else:
+
+@pytest.fixture(scope="session", autouse=True)
+def temp_cwd():
+    original_cwd = os.getcwd()
+    with tempfile.TemporaryDirectory() as tmp:
```
jdavies-st, line 10: By using Python's `tempfile` and not the built-in feature of Pytest, how does this respond to

```
pytest --basetemp=/tmp/scratch
```

in directing where the tempdirs get stored?

Generally, the Pytest temp dirs do not get deleted when the Python process gets completed, so that the results of the test can be examined, and the Windows issue probably avoided.

eslavich: There's no impact on files stored in a pytest temp directory, only to files that are saved relative to the current working directory.  We don't do that in the regular test suite because it would leave junk behind, but it's hard to avoid in the docstring tests where we want the examples to work when copied and pasted into the user's console.

eslavich: (we chatted about this on Slack and I learned about the tmpdir_factory fixture)

jdavies-st (approved):
New version looks good!

PR #955, opened 2021-03-23
Drop workarounds for bugs in older versions of Python on macOS and Windows

### LINKED ISSUE #932: Drop unnecessary workarounds in generic_io.py
We have a few workarounds for OS-specific bugs in generic_io.py that we probably don't need to maintain anymore:

- [Workaround](https://github.com/asdf-format/asdf/blob/c446d9040de2adb2c05c215855837e948339d1b4/asdf/generic_io.py#L66) for array reads in OS X Mavericks, which is well past end of life
- [Workaround](https://github.com/asdf-format/asdf/blob/c446d9040de2adb2c05c215855837e948339d1b4/asdf/generic_io.py#L105) for a bug when writing arrays > 2GB on a Mac, which has been fixed in all currently maintained versions of Python.
- [Workaround](https://github.com/asdf-format/asdf/blob/c446d9040de2adb2c05c215855837e948339d1b4/asdf/generic_io.py#L115) for a similar bug in Windows.  This one needs a little more research.

eslavich:
This removes some workarounds from bugs that no longer affect us.

- The macOS read/write issue was resolved in Python 3.6+: https://bugs.python.org/issue24658
- I haven't found documentation on the Windows issue, but I tested on Python 3.6.13, 3.7.10, and 3.8.5 on Windows and was able to write a 4 GB array using this branch.

Resolves #932

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,8 @@
 
 - Remove unnecessary ``.tree`` from search result paths. [#954]
 
+- Drop support for bugs in older operating systems and Python versions. [#955]
+
 2.7.4 (unreleased)
 ------------------
 
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -63,80 +63,6 @@ def _check_bytes(fd, mode):
     return True
 
 
-if (sys.platform == 'darwin' and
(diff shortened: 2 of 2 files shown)


jdavies-st (approved):
🎉 deleted code!

PR #956, opened 2021-03-25
Add --ignore-query flag to asdftool diff

eslavich:
This adds a `--ignore-query` (or `-i`) flag to the asdftool diff command.  It accepts a JMESPath expression and ignores matching nodes when computing the diff.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,7 +49,7 @@
 
 - Drop support for bugs in older operating systems and Python versions. [#955]
 
-- Add argument to ``asdftool diff`` that ignores tree nodes based on
+- Add argument to ``asdftool diff`` that ignores tree nodes that match
   a JMESPath expression. [#956]
 
 2.7.4 (unreleased)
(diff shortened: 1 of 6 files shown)


jdavies-st (commented):
Cool!  This is very simple, and I like it.

That said, it's not clear how to use it, so a couple simple examples in the docs and a docstring somewhere would be useful.

Also, I know `jmespath` calls these queries, but from a top-level `asdf` user standpoint, when using `asdftool`, it might be clearer to use:

```
$ asdftool diff --ignore=meta.date file1.asdf file2.asdf
$ asdftool diff --ignore="meta.[date, filename]" file1.asdf file2.asdf
```

as we are using it not so much as a query at the top level but as a filter, even though the string is a query.

nden (approved):
I agree with James' comments. `ignore` or `ignore_attrs` sounds better to me.
Other than that it looks good and works well.

eslavich:
I changed the argument name to `--ignore`, added an epilogue with examples, and added a docstring.  Let me know if you think that's sufficient, documentation-wise.

jdavies-st (approved):
This looks good!

Issue #946, opened 2021-03-22
Add weldx to downstream workflow?


eslavich:
@CagtayFabry we have an actions workflow in this repository that runs ASDF-related tests in a handful of our downstream projects.  We run this workflow on a regular schedule and also immediately prior to release, to make sure we're not about to break something.  Are you interested in having weldx added to the list?  If so, can you confirm that this is a reasonable test script:

```
git clone https://github.com/BAMWelDX/weldx.git
cd weldx
conda env create -f environment.yml
conda activate weldx
pytest tests/asdf_tests
```

CagtayFabry:
Thanks for the notification and good to know about downstream tests @eslavich 💯 
definitely interested, I will get back to you in more detail in a few days

We will have to check if running `pytest tests/asdf_tests` is sufficient (it should be).

we will probably remove `environment.yml` soon, I think something like the following should work (and be faster)

```yaml
runs-on: ubuntu-latest
    steps:
      - name: Set up Python 3.9
        uses: actions/setup-python@v2
        with:
          python-version: 3.9
      - name: Install asdf
        run: pip install git+https://github.com/asdf-format/asdf@${{ github.event.inputs.asdf_ref || 'master' }}
      - name: Install pytest
        run: pip install pytest pytest-cov pytest-xdist setuptools_scm
      - name: Install weldx
        run: | 
          git clone https://github.com/BAMWelDX/weldx.git
          cd weldx
          pip install -e .
      - name: Run weldx tests
        run: pytest tests/asdf_tests --asdf-tests
```

Maybe it would be possible to test the current `asdf` master against both, the current `weldx` master (for our development) as well as our latest pypi release (for stability).

also 🔔  @vhirtham @marscher

PR #960, opened 2021-03-30
add weldx tests to downstream.yml

CagtayFabry:
add downstream test against `weldx` master to as mentioned #946

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -81,3 +81,21 @@ jobs:
         run: pip install git+https://github.com/astropy/astropy@${{ github.event.inputs.astropy_ref || 'master' }}#egg=astropy[test]
       - name: Run astropy tests
         run: pytest --pyargs astropy
+
+  weldx:
+    name: weldx@master unit tests
+    runs-on: ubuntu-latest
+      steps:
+          - name: Set up Python 3.9
+            uses: actions/setup-python@v2
+            with:
+              python-version: 3.9
+              - name: Install asdf
+                run: pip install git+https://github.com/asdf-format/asdf@${{ github.event.inputs.asdf_ref || 'master' }}
+              - name: Install weldx
+                run: |
+                  git clone https://github.com/BAMWelDX/weldx.git
+                  cd weldx
+                  pip install -e .[test]
+              - name: Run weldx tests
+                run: pytest tests/asdf_tests --asdf-tests


CagtayFabry:
can you trigger this manually to see if it works @eslavich ?

eslavich:
GitHub isn't allowing me to run the workflow from this branch, I think there may be a syntax error somewhere?

eslavich:
After `python-version: 3.9` the subsequent steps are indented too far

CagtayFabry:
thanks, that got messed up copying over here

eslavich:
Here it goes...

https://github.com/asdf-format/asdf/runs/2229659222?check_suite_focus=true

.github/workflows/downstream.yml
```diff
@@ -81,3 +81,21 @@ jobs:
...
+        with:
+          python-version: 3.9
+      - name: Install asdf
+        run: pip install git+https://github.com/asdf-format/asdf@${{ github.event.inputs.asdf_ref || 'master' }}
+      - name: Install weldx
+        run: |
+          git clone https://github.com/BAMWelDX/weldx.git
+          cd weldx
+          pip install -e .[test]
+      - name: Run weldx tests
```
eslavich, line 100: This failed due to missing tests/asdf_tests, it must be that each step resets the working directory before it runs?  Up above we're using `actions/checkout@v2` to fetch the repository which seems to take care of the working directory automatically.

eslavich: Actually now that I say that I see that the weldx steps almost fit into the matrix job.  Maybe we should add a new variable to the matrix for additional pytest arguments?

CagtayFabry:
matrix looking good, thanks for the test run !
We will probably have to set the default branch to `asdf_2.8` for the time beeing

CagtayFabry:
I am so sorry, no idea why my indentation keeps failing...

eslavich:
What do you think of this?

https://github.com/asdf-format/asdf/compare/test_weldx_downstream

I added a `test_command` variable so that matrix jobs can customize the pytest command.

CagtayFabry:
> 
> 
> What do you think of this?
> 
> https://github.com/asdf-format/asdf/compare/test_weldx_downstream
> 
> I added a `test_command` variable so that matrix jobs can customize the pytest command.

that looks very nice 👍 
Just need to set the default ref to `asdf_2.8` instead of `master` for weldx

eslavich:
Hooray, success!

https://github.com/asdf-format/asdf/runs/2230976642?check_suite_focus=true

I'll change the branch on this PR to the other one.

eslavich:
Oops, I can't do that, so here's a new PR: https://github.com/asdf-format/asdf/pull/961

eslavich:
I merged #961 so I think this one can be closed.  Thanks @CagtayFabry !

PR #961, opened 2021-03-30
Test weldx downstream

### LINKED ISSUE #946: Add weldx to downstream workflow?

eslavich:
This is similar to #960 but adds a new parameter to the job matrix to allow custom test commands.

Resolves #946

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -59,7 +59,7 @@ jobs:
             test_command: pytest
           - package_name: weldx
             repository: BAMWelDX/weldx
-            ref: ${{ github.event.inputs.weldx_ref || 'master' }}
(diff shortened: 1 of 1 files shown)


CagtayFabry:
Thanks @eslavich, I will keep this updated should we make any changes

Issue #890, opened 2020-11-25
adding custom %TAG entries to file header


CagtayFabry:
I think it would be nice to provide a way for custom extensions to add additional `%TAG` directives to the top YAML header to shorten tags provided by custom extensions that might lead to long tag strings. ASDF already does this by introducing `%TAG ! tag:stsci.edu:asdf/` .

Consider the following example:
```yaml
#ASDF 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
# asdf headers
fraction: !<tag:nowhere.org:custom/fraction-1.0.0> [10, 3]
...
```
With adding  `%TAG !foo! tag:nowhere.org:custom/` we get the shorter
```yaml
#ASDF 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
%TAG !foo! tag:nowhere.org:custom/
--- !core/asdf-1.1.0
# asdf headers
fraction: !foo!fraction-1.0.0 [10, 3]
...
```

The only thing needed to get this output is to add the `"!foo!": "tag:nowhere.org:custom/"` to the `tags` mapping in the final `    yaml.dump_all` call in `asdf.yamlutil.dump_tree` here:
https://github.com/asdf-format/asdf/blob/8ccee8b17a053d6e33ebca81a4ed3f3c3569aee1/asdf/yamlutil.py#L408-L413

Currently `tags` is (statically) build to simply consist of `{"!": "tag:stsci.edu:asdf/"}`

One possible way to implement this would be to introduce a new (optional) property for the ExtensionClass that provides the tags.
Example extending the `FractionExtension` with a new property `yaml_tag_context`:
```python
class FractionExtension(asdf.AsdfExtension):
    # ...

    @property
    def yaml_tag_context(self):
        return {"!foo!": "tag:nowhere.org:custom/"}
```
We can now query this property for registered extensions and add them to the list.

I tested this for simple read/write and it works, see this simple diff  https://github.com/asdf-format/asdf/compare/master...CagtayFabry:test_custom_yaml_tag

~Of course it would be nice to only load `%TAG` values from extensions that are actually used when creating the file (similar to listing them in the history)~ update

eslavich:
[...]

PR #963, opened 2021-03-31
add extension %TAG definitions

CagtayFabry:
this is the implementation discussed in #890

I was wondering if this is still of interest?
Maybe it could be included in the 2.8 release?

Here is an example implementation for the `WeldxExtension`:
```python
class WeldxExtension(AsdfExtension):
    """Extension class registering types with both tags and schemas defined by weldx."""

    # ... default properties

    @property
    def yaml_tag_handles(self):
        return {"!weldx!": "tag:weldx.bam.de:weldx/"}
```

output file:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
%TAG !weldx! tag:weldx.bam.de:weldx/
--- !core/asdf-1.1.0
  name: asdf, version: 2.8.0.dev223+g7916a92}
history:
  extensions:
    software: !core/software-1.0.0 {name: asdf, version: 2.8.0.dev223+g7916a92}
    extension_class: weldx.asdf.extension.WeldxAsdfExtension
    software: !core/software-1.0.0 {name: weldx, version: 0.3.4.dev10+gd41bb1a.d20210331}
    extension_class: weldx.asdf.extension.WeldxExtension
    software: !core/software-1.0.0 {name: weldx, version: 0.3.4.dev10+gd41bb1a.d20210331}
R: !weldx!core/transformations/rotation-1.0.0
  angles: !unit/quantity-1.1.0 {unit: degree, value: 20}
  sequence: x
...
```


CagtayFabry:
I am not sure what a good test for this could look like (maybe a roundtrip that uses a custom `%TAG` entry?)

What do you think @eslavich ?

eslavich:
Yes, I think we can get this into 2.8.  A roundtrip is a good idea for a test.  We should probably also write a file and search through the raw bytes to make sure the custom `%TAG` is used, since a roundtrip would succeed even if the custom tag was being completely ignored.

asdf/yamlutil.py
```diff
@@ -405,6 +405,14 @@ def dump_tree(tree, fd, ctx, tree_finalizer=None, _serialization_context=None):
     yaml_version = tuple(
         int(x) for x in ctx.version_map['YAML_VERSION'].split('.'))
 
+    # add yaml %TAG definitions from extensions
+    for ext in _serialization_context._extensions_used:
+        ext_tags = getattr(ext.delegate, "yaml_tag_context", None)
```
eslavich, line 410: In the [YAML spec](https://yaml.org/spec/1.2/spec.html#tag/shorthand/) this is referred to as a "tag handle".  Can we call it that here too, instead of "context"?

eslavich: Also, this property should be added to both `Extension` and `ExtensionProxy` in `asdf/extension/_extension.py`.

CagtayFabry: sure that sounds grood, I will look into it

CagtayFabry: I am not sure how to correctly implement the `Extension` properties, if you would be so kind to take another look @eslavich

asdf/yamlutil.py
```diff
@@ -405,6 +405,15 @@ def dump_tree(tree, fd, ctx, tree_finalizer=None, _serialization_context=None):
     yaml_version = tuple(
         int(x) for x in ctx.version_map['YAML_VERSION'].split('.'))
 
+    # add yaml %TAG definitions from extensions
+    if _serialization_context:
+        for ext in _serialization_context._extensions_used:
+            ext_tags = getattr(ext.delegate, "yaml_tag_handle", None)
```
eslavich, line 411: With the changes I recommended above, we should be able to simplify to this:

```python
for ext in _serialization_context._extensions_used:
    for key, val in ext.yaml_tag_handles.items():
        if key not in tags:
            tags[key] = val
```

eslavich: Up above it looks like it's possible for `tags` to still be None when we get here.  Maybe we should initialize it with an empty dict so we don't risk a problem?

eslavich:
Can you add a CHANGES.rst entry for this work?  I'll take care of the documentation when I write up docs for the new extension API in general.

asdf/extension/_extension.py
```diff
@@ -298,6 +316,24 @@ def legacy(self):
         """
         return self._legacy
 
+    @property
+    def yaml_tag_handle(self):
```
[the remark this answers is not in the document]
CagtayFabry: > 
> 
> Can you add a CHANGES.rst entry for this work? I'll take care of the documentation when I write up docs for the new extension API in general.

yes I will clean this up and update `CHANGES.rst`
thanks for your input

CagtayFabry:
> 
> 
> Can you add a CHANGES.rst entry for this work? I'll take care of the documentation when I write up docs for the new extension API in general.

yes I will clean this up and update CHANGES.rst
thanks for your input

asdf/yamlutil.py
CagtayFabry: I wasn't sure in what context this part would not get executed
https://github.com/asdf-format/asdf/blob/62fe49a9032e0883c0b0f5ebe0f51c7b4f6e72de/asdf/yamlutil.py#L388-L398

but you are right, I will init with empty dict and change accordingly

eslavich: Hmm... you're right, and the [standard](https://asdf-standard.readthedocs.io/en/1.5.0/tree.html#tags) says:

> ... the tag of the root element of the tree must always be tag:stsci.edu:asdf/core/asdf-1.1.0

So if tree_type is `None`, that should probably be an error.  I'll file a separate issue to clean that up.

asdf/extension/_extension.py
```diff
@@ -298,6 +316,24 @@ def legacy(self):
...
+        See https://yaml.org/spec/1.2/spec.html#tag/shorthand/
+
+        Example: ``{"!foo!": "tag:nowhere.org:custom/"}``
+
+        Returns
+        -------
+        dict
+
+        """
+        return self._yaml_tag_handle
```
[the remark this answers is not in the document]
CagtayFabry: updated now

eslavich (approved):
This looks ready to go, thanks!  The documentation for the new extension API doesn't exist yet, but I'll be sure to add a section on this when I get to it.

CagtayFabry:
looking forward to it 🚀

PR #971, opened 2021-04-06
Support https and redirects

### LINKED ISSUE #965: HTTPS and 3xx redirect support
This library does not support opening files from https URLs, only http.  This line:

https://github.com/asdf-format/asdf/blob/2.7.3/asdf/generic_io.py#L1065

uses `http.client.HTTPConnection` for all URLs, when it should use HTTPSConnection for https.

We also don't follow redirects -- there's no built-in support for that in http.client.

See https://github.com/astropy/specutils/issues/801 for a traceback when an https URL is passed to asdf.open.

eslavich:
This PR drops support for lazy-loading arrays over http and adds support for https and 3xx redirects.  The lazy-loading is a nice idea, but not useful in its current state and something we should revisit later when we have the time.

Resolves #965

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -1763,7 +1763,7 @@ def is_asdf_file(fd):
         if parsed.scheme in ["http", "https"]:
             # We don't want to read URL content here because
             # that will cause the file to be downloaded twice.
-            return os.path.splitext(parsed.path)[1] == "asdf"
+            return os.path.splitext(parsed.path)[1] == ".asdf"
 
     to_close = False
     if isinstance(fd, AsdfFile):
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -30,7 +30,7 @@ from .extension import (
     get_cached_asdf_extension_list,
     get_cached_extension_manager,
 )
-from .util import NotSet
+from .util import NotSet, patched_urllib_parse
 from .search import AsdfSearchResult
 from ._helpers import validate_version
 
@@ -1744,7 +1744,10 @@ def is_asdf_file(fd):
     """
     Determine if fd is an ASDF file.
 
-    Reads the first five bytes and looks for the ``#ASDF`` string.
+    For most input, reads the first five bytes and looks
+    for the ``#ASDF`` string.
+
+    For URL input, looks for a .asdf extension.
 
     Parameters
     ----------
@@ -1754,6 +1757,13 @@ def is_asdf_file(fd):
     if isinstance(fd, generic_io.InputStream):
         # If it's an InputStream let ASDF deal with it.
         return True
+
+    if isinstance(fd, str):
+        parsed = patched_urllib_parse.urlparse(fd)
+        if parsed.scheme in ["http", "https"]:
+            # We don't want to read URL content here because
+            # that will cause the file to be downloaded twice.
+            return os.path.splitext(parsed.path)[1] == "asdf"
 
     to_close = False
     if isinstance(fd, AsdfFile):
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -1740,6 +1740,13 @@ def open_asdf(fd, uri=None, mode=None, validate_checksums=False, extensions=None
         **kwargs)
 
 
+# astropy.io.fits supports opening files that are externally
+# compressed with gzip or zip, and asdf does not, so we may as
+# well give those extensions to FITS.
+_FITS_EXTENSIONS = [".fits", ".gz", ".zip"]
+_ASDF_EXTENSIONS = [".asdf"]
+
+
 def is_asdf_file(fd):
     """
     Determine if fd is an ASDF file.
@@ -1747,7 +1754,8 @@ def is_asdf_file(fd):
     For most input, reads the first five bytes and looks
     for the ``#ASDF`` string.
 
-    For URL input, looks for a .asdf extension.
+    For URL input, looks for an extension that should be passed
+    to AsdfInFits, otherwise assumes ASDF.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


asdf/asdf.py
```diff
@@ -1755,6 +1758,13 @@ def is_asdf_file(fd):
         # If it's an InputStream let ASDF deal with it.
         return True
 
+    if isinstance(fd, str):
```
eslavich, line 1761: Is it reasonable to detect ASDF vs ASDF-in-FITS using the URL's filename extension?  Ideally we would peek the file content here too, but doing that requires making multiple HTTP requests or passing a GenericFile object to astropy.io.fits.  astropy can handle GenericFile but doesn't treat it as a real file and I'm not sure what the consequences of that are.  Passing the URL to astropy also brings the benefit of cached downloads.

perrygreenfield: You mean something like http://somewhere.edu/somepath/data.fits[asdfextention]? (I don't actually know the right syntax for identifying the extension; by number only, extname, extname+extver, etc.)

eslavich: I just mean `filename.fits` vs `filename.asdf`.  The implication is that the library won't be able to open an obfuscated URL that serves ASDF, like `https://example.com/files/441ec2c4-a18c-42b5-b6b4-9bc92cfeb013`.

eslavich: Another option is to check for `.fits`, `.fits.gz`, `.asdf`, and use that if present but download the first few bytes of the file if not.

perrygreenfield: I'm not sure I understand the need to support asdf in fits implicitly. Is there a good use case?

perrygreenfield: Initially, I'd be happy with requiring the ".asdf" extension. We can always expand the possibilities later.

eslavich: > I'm not sure I understand the need to support asdf in fits implicitly. Is there a good use case?

The initial bug report was from a specutils user opening an ASDF-in-FITS over https.  Basically they're doing this:

```python
asdf.open("https://data.science.stsci.edu/redirect/JWST/jwst-data_analysis_tools/IFU_optimal_extraction/NRS00001-faintQSO-F100LP-G140H-01_1_491_SE_2020-08-25T12h15m00_s3d.fits")
```

This `is_asdf_file` function determines whether we open the file as ASDF or as FITS.  It's not the only check -- if we proceed with opening as ASDF, there will still be an error if the initial bytes are unrecognized, but at that point we will have lost our opportunity to open as FITS.

jdavies-st: I think this should work with `AsdfInFits` if possible, as that's our current use case.  Not sure the best implementation.  But yeah, I think if you're passing a `.fits` file to `asdf.open()`, then the assumption that it is an `AsdfInFits` file should be valid.

eslavich: Should we make ASDF the default?  The way I've coded it the library will open the file as FITS unless there's a `.asdf` extension present.  Might be more fitting to turn that around so that FITS only happens if the URL has a `.fits` or `.fits.gz` extension and assume everything else is ASDF.

perrygreenfield: Why not require the http string include some information for ASDF to know that it is ASDF in FITS? That would be much simpler to implement.

jdavies-st: I don't think it's a good idea to require a URL string to inform `asdf` whether it is a good file to open or not.  We don't have any control in how Box formats its URLs.  And it adds extra complexity for the user if they are linking to data on Box.  It adds an extra failure point that can very easily be dealt with in the `asdf` code.

I think if the user is trying to open a fits file with `asdf.open()`, the assumption should be it is `AsdfInFits`. If it doesn't work, a useful error is raised.

Now how we determine it's a FITS file could be `.fits` or `.fits.gz` extension, or some bytes at the beginning.  I don't have a preference.  But perhaps going the simple route of extension is the best for now.  In the end, we want this to work for URLs of JWST .fits data.

Agree that opening as `.asdf` should be the default and only trigger `AsdfInFits` if there's an extension indicating.

eslavich: We could move towards requiring ADSF-in-FITS users to open with `asdf.fits_embed.open` or something like that.  I'd prefer that over the current strategy of auto-detecting with "one open method to rule them all".  It would be nice too to keep the FITS support quarantined to that particular module instead of distributing it around the library.  Maybe someday we'll want to move it into a separate package.

perrygreenfield: It all depends on where this URL comes from doesn't it? Could I see an example of a blind URL (i.e., just copying and pasting one into an open function without any knowledge of the context)? If you know the context, you can explicitly provide that to the open function. Hey, this is an ASDF in FITS case, open it as such (using a keyword, or special open function). For a file on the local file system I can see trying to figure out what it is since that is low cost. But a remote http request is less so. You might as well copy the whole file and then do that.

jdavies-st: That could work, but that would involve some upstream breakage of existing code in `specutils`, `jwst`, `romancal`, etc., as currently datamodels, and most other packages use `asdf.open()`, and there's no reason for a package to use separate open function if it's going to be a URL, because a URL or a file are only known at runtime.  So it would always have to use `asdf.fits_embed.open` if it knew it was `AsdfInFits`.

eslavich: The "one open method" solution would look something like this:

- Request the URL and wait for the first 5 bytes to arrive.  Peek their values without consuming any of the stream.
- If `#ASDF`, pass on to AsdfFile
- If anything else, pass on to AsdfInFits

After digging around yesterday I'm of the opinion that this isn't feasible under the current structure of asdf.py (that module is a superfund site!), and it may not be wise to include a dramatic refactor in a 2.7.x release.

eslavich: (the ideal solution I mean, but if we want to get there then a band-aid based on the URL filename extension seems like a good short-term solution)

jdavies-st: Agree that the bandaid might be best for now, as we should not be doing breaking changes in a 2.7.4 release.  So for 2.7.x (and 2.8x I would argue), `asdf.open()` should open `AsdfInFits` files in the same way it always has.

Agree we should revisit this when we get a chance to refactor. I see there's already a #TODO in there already to fix `_open_impl`

https://github.com/asdf-format/asdf/blob/ba09e57d9344206016127d1a86d3319a2df9f2ec/asdf/asdf.py#L925-L929

asdf/generic_io.py
```diff
@@ -827,211 +825,41 @@ def fast_forward(self, size):
...
-    size = int(response.getheader('content-length'))
-    first_chunk = response.read(block_size)
-    response.close()
-    return HTTPConnection(connection, size, parsed.path, uri or init,
-                          first_chunk)
+    fd = tempfile.NamedTemporaryFile("w+b")
+    try:
+        # This method is only called with http and https schemes:
+        with urlopen(init) as response: # nosec
+            chunk = response.read(io.DEFAULT_BUFFER_SIZE)
```
eslavich, line 853: I should be using `asdf.get_config().io_block_size` here

asdf/asdf.py
eslavich: I kept the URL-based detection but made ASDF the default, and added a helpful warning when the library is forced to assume ASDF.

jdavies-st (approved):
Looks good!

eslavich:
Submitted an issue here for improving this: https://github.com/asdf-format/asdf/issues/973

PR #972, opened 2021-04-12
Astropy master to main in downstream workflow

jdavies-st:
This updates our downstream workflow to run astropy tests from the `main` branch instead of `master`.  The name changed this past week.  It was failing this morning:

https://github.com/asdf-format/asdf/runs/2321003190?check_suite_focus=true

Also, it seems this workflow is a bit fragile.  Generally, we should be running the _default_ branch of all these repos, assuming the default branch is the dev branch.  Is there a reason we need to even be naming this branch?

```
pip install git+https://github.com/org/repo
```

will install the project from the default branch without needing to specify it.  Maybe we should simplify this downstream workflow?

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -10,7 +10,7 @@ on:
       astropy_ref:
         description: astropy ref
         required: true
-        default: master
+        default: main
       gwcs_ref:
         description: gwcs ref
         required: true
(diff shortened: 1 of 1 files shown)


jdavies-st:
The downstream workflow run on this branch:

https://github.com/jdavies-st/asdf/actions/runs/741260219

eslavich:
> Also, it seems this workflow is a bit fragile. Generally, we should be running the default branch of all these repos, assuming the default branch is the dev branch. Is there a reason we need to even be naming this branch?

The idea behind the parameterized name was to allow us to override it and run the tests on a branch or release tag.  That might come up if we know that the dev branch is broken with a certain asdf release but there's an unmerged branch that fixes it.  It may also be a good idea to test with both dev and the latest release of each package so that we know if we're about to cause problems with what's available.

eslavich (approved):

jdavies-st:
Ah, ok.  That's a great use case, and I hadn't thought of that.  👍

PR #977, opened 2021-04-27
Prevent astropy warnings in tests

### LINKED ISSUE #966: Resolve astropy 4.2.1 FITS header warning
A handful of tests in the warnings-as-exceptions job began failing with the release of astropy 4.2.1:

```
E                       astropy.io.fits.verify.VerifyWarning: Error validating header for HDU #0 (note: Astropy uses zero-based indexing).
E                           Header size is not multiple of 2880: 1246
E                       There may be extra bytes after the last HDU or the file is corrupted.
```

It looks like some of the test data we have checked into the repo needs to be regenerated

eslavich:
astropy 4.2.1 started emitting an additional warning when opening bad files.  This PR marks tests that open known bad files to ignore that warning.

Resolves #966

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,8 @@
 
 - Add support for HTTPS URLs and following redirects. [#971]
 
+- Prevent astropy warnings in tests when opening known bad files. [#977]
+
 2.7.3 (2021-02-25)
 ------------------
 
--- a/asdf/commands/tests/test_diff.py
+++ b/asdf/commands/tests/test_diff.py
@@ -55,6 +55,7 @@ def test_diff_simple_inline_array():
     _assert_diffs_equal(filenames, result_file, minimal=False)
 
 
+@pytest.mark.filterwarnings('ignore::astropy.io.fits.verify.VerifyWarning')
 def test_file_not_found():
     # Try to open files that exist but are not valid asdf
     filenames = ['frames.diff', 'blocks.diff']
--- a/asdf/tags/core/tests/test_history.py
+++ b/asdf/tags/core/tests/test_history.py
@@ -207,6 +207,7 @@ history:
             pass
 
     # Make sure to test for incompatibility with ignore_missing_extensions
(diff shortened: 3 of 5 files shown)


jdavies-st:
Do we know why these are bad files?  Almost every test file that the `asdf` library has written now gets these warnings.  There are dozens in the `jwst` repo as well. 🤔

eslavich:
> Do we know why these are bad files?

In asdf at least they're intentionally bad, for example:

```python
def test_no_asdf_header(tmpdir):
    content = b"What? This ain't no ASDF file"
```

That content isn't recognized as ASDF, so it gets passed to `astropy.io.fits`, which emits the warning.  Do you have examples in `jwst` of files that should be valid ASDF or FITS but still result in a warning?

jdavies-st:
Yeah, a ton of them.  Start here and scroll up.

https://github.com/spacetelescope/jwst/runs/2422915561?check_suite_focus=true#step:10:845

That said, you bring up a good point.  Some are likely bad files, but I've looked at a few, and they are not even writing out files, which makes me think `AsdfInFits` is generating bad files?  Not sure.  Would be good to understand whether it's a bug in `astropy.io.fits` or a bug in `asdf.AsdfInFits` instead of sweeping it under the rug.

eslavich:
This commit seems responsible:

https://github.com/astropy/astropy/commit/070cd942a6ae9e18513c21c19efe0eb91a969972#diff-6ebde8ab98c018da8dd092e05ef30d9607dc9f998bfed0df91a6b67c671ca651

The warning is emitted from within that `_check_padding` function.  Previously they were raising "Header missing END card." for invalid files before `_check_padding`.  Now the order is reversed so the warnings are appearing before the exception is raised.

jdavies-st (approved):
This looks good.

jdavies-st:
And let's discuss any issues with `jwst` elsewhere, preferably in a `jwst` issue.

PR #978, opened 2021-04-28
Improve file type detection for streaming input

### LINKED ISSUE #973: Improve file type detection for URLs and InputStream objects
The `asdf.asdf.is_asdf_file` method reads the first few bytes of content for most input, but for URLs it relies on the filename suffix in the URL string.  This mean that FITS files at obfuscated URLs (or otherwise missing a recognizable suffix) fail to open with `asdf.open`.  Ideally we would pass the open stream to `is_asdf_file` and peek the first few bytes without consuming them.

Similarly, `InputStream` objects are always assumed to be ASDF.

### LINKED ISSUE #975: passed pure-read-only file wrapper makes parse_header fail
Hi there,
eventually I'm doing something completely stupid here, but please have a look at this test snippet:

```python
def test_readonly_file(tmpdir):
    fn = os.path.join(str(tmpdir), 'test.asdf')
    a = asdf.AsdfFile()
    a.tree["test"] = True
    a.write_to(fn)

    class FileWrapper:
        def __init__(self):
            self.fd_read_only = open(fn, 'rb')

        def read(self, n):
            return self.fd_read_only.read(n)

    wrapper = FileWrapper()
    assert isinstance(generic_io.get_file(wrapper, 'r'), generic_io.InputStream)

    asdf.open(wrapper, mode='r')
```
produces the output:
``` python
tests/test_generic_io.py:467: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
asdf.py:1733: in open_asdf
    return AsdfFile._open_impl(instance,
asdf.py:948: in _open_impl
    return cls._open_asdf(self, fd, uri=uri, mode=mode,
asdf.py:842: in _open_asdf
    self._file_format_version = cls._parse_header_line(header_line)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'asdf.asdf.AsdfFile'>, line = b' 1.0.0\n'

    @classmethod
    def _parse_header_line(cls, line):
        """
        Parses the header line in a ASDF file to obtain the ASDF version.
        """
        parts = line.split()
        if len(parts) != 2 or parts[0] != constants.ASDF_MAGIC:
>           raise ValueError("Does not appear to be a ASDF file.")
E           ValueError: Does not appear to be a ASDF file.

asdf.py:755: ValueError

```

This is happens for current master. The generic_io.InputStream wrapper is intended for pure read-only operations as far as I understood. So it should be possible to read the asdf file created within the test case, right?

eslavich:
This PR updates the code inside of `asdf.open` that detects the input file type.  It resolves some bugs in reading from URLs and non-seekable open files, but also introduces other changes that I should call out:

- It is no longer possible to call `asdf.open` on an `HDUList` instance; instead `asdf.fits_embed.AsdfInFits.open` must be called directly.  This is technically a breaking change but not a feature that is currently used by jwst, and I'm not aware of other users of AsdfInFits.

- It is no longer possible to call `asdf.open` with gzipped or zipped FITS input.  Instead `asdf.fits_embed.AsdfInFits.open` must be called directly.

- When `asdf.open` is called with FITS input, we're now passing an instance of `asdf.generic_io.GenericFile` to `astropy.io.fits.open`.  That shouldn't be a problem since GenericFile has all the standard file-like methods, but it is a difference.

Resolves #975
Resolves #973


asdf/asdf.py
```diff
@@ -797,7 +797,7 @@ def _find_asdf_version_in_comments(cls, comments):
         return None
 
     @classmethod
-    def _open_asdf(cls, self, fd, uri=None, mode='r',
```
kmacdonald-stsci, line 800: Can `uri` be removed as a parameter?  It's used at the call at https://github.com/asdf-format/asdf/blob/7e63794cd4a8e6b73e33f336dd421edc8b53c8e6/asdf/asdf.py#L950

kmacdonald-stsci (approved):

asdf/asdf.py
eslavich: Oops, thanks!  I guess it works due to the `**kwargs`.  I'll remove it from the call and make sure there aren't any other cases.

asdf/tests/test_fits_embed.py
```diff
@@ -26,7 +26,7 @@
 )
 
 
-TEST_DTYPES = ['<f8', '<f8', '<u4', '>u4', '<i4', '>i4']
+TEST_DTYPES = ['<f8', '>f8', '<u4', '>u4', '<i4', '>i4']
```
jdavies-st, line 29: Good catch!

asdf/tests/test_fits_embed.py
```diff
@@ -275,33 +275,6 @@ def test_asdf_in_fits_open(tmpdir, dtype):
...
-    asdf_in_fits.write_to(tmpfile)
-
-    # Test opening the file directly from the URI
-    with asdf_open(tmpfile) as ff:
-        compare_asdfs(asdf_in_fits, ff)
-
-    # Test open/close without context handler
-    ff = asdf_open(tmpfile)
-    compare_asdfs(asdf_in_fits, ff)
-    ff.close()
```
jdavies-st, line 293: Shouldn't these still be tested, as they are still supported?

We're only getting rid of reading `fits.gz`, open `HDUList` objects, right?  Also what about passing an open file handle of the FITS file?  That is not mentioned above as something that is being removed.  Is it?

asdf/asdf.py
```diff
@@ -1757,11 +1764,18 @@ def is_asdf_file(fd):
...
+
     Parameters
     ----------
     fd : str, `~asdf.generic_io.GenericFile`
 
     """
+    warnings.warn(
+        "The asdf.asdf.is_asdf_file method is deprecated.  Use asdf.util.get_file_type instead.",
+        AsdfDeprecationWarning,
+    )
```
jdavies-st, line 1777: The `is_asdf_file()` method is only used internally, right?  Or is it used elsewhere?

jdavies-st (approved):
Looks good.  I like the `Enum` subclass for file type.  Very slick.

Just a couple questions below.

asdf/asdf.py
eslavich: I'm not sure.  It's tested as though it were part of the public API, but it's hard to tell since historically we haven't been careful about clearly marking public vs internal.  GitHub search only returns hits for the asdf repo for the string "is_asdf_file".

eslavich: I'll remove it...

PR #987, opened 2021-05-03
Use all the Draft 4 components in schema validator

### LINKED ISSUE #976: Local references in metaschemas do not resolve correctly
When `asdf.schema.check_schema` is called on a schema with a custom metaschema with local references (`$ref: "#"` or the like), the references get resolved to https://json-schema.org/draft-04/schema intead of the custom metaschema.

eslavich:
This fixes a bug causing relative references in metaschemas to be resolved incorrectly (for example, the reference `#/definitions/schemaArray` was being resolved to `http://json-schema.org/draft-04/schema#/definitions/schemaArray` instead of the the host document).  The problem turned out to be that we have been using some Draft 7 components when building up our schema validator.

Resolves #976

### DIFF
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -744,7 +744,11 @@ def check_schema(schema, validate_default=True):
 
     resolver = _make_resolver(extension.get_default_resolver())
 
-    cls = mvalidators.create(meta_schema=meta_schema,
-                             validators=validators)
+    cls = mvalidators.create(
+        meta_schema=meta_schema,
+        validators=validators,
+        type_checker=mvalidators.Draft4Validator.TYPE_CHECKER,
+        id_of=mvalidators.Draft4Validator.ID_OF,
+    )
     validator = cls(meta_schema, resolver=resolver)
     validator.validate(schema, _schema=meta_schema)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -63,7 +63,10 @@
   filenames without recognizable extensions.  Remove the ``asdf.asdf.is_asdf_file``
   function. [#978]
 
-- Fix bug in ``asdf.schema.check_schema`` that caused relative references in
+2.7.5 (unreleased)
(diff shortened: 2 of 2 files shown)


eslavich:
Resolves #976

asdf/schema.py
```diff
@@ -744,7 +744,11 @@ def _validate_default(validator, default, instance, schema):
...
     resolver = _make_resolver(extension.get_default_resolver())
 
-    cls = mvalidators.create(meta_schema=meta_schema,
-                             validators=validators)
+    cls = mvalidators.create(
+        meta_schema=meta_schema,
+        validators=validators,
+        type_checker=mvalidators.Draft4Validator.TYPE_CHECKER,
+        id_of=mvalidators.Draft4Validator.ID_OF,
+    )
```
jdavies-st, line 752: Are these supposed to be different from the `type_checker` here

https://github.com/asdf-format/asdf/blob/4899ac53720ddced2f5f494dd52b5c3d609f6fe5/asdf/schema.py#L265-L269

and the `id_of` here (same thing)

https://github.com/asdf-format/asdf/blob/4899ac53720ddced2f5f494dd52b5c3d609f6fe5/asdf/schema.py#L270-L276

?

jdavies-st: And if not, a comment on why they are being redefined there but not here might be in order.  It's not clear to me.  I.e. a docstring in `_create_validator()` would help.

eslavich: This is a validator for checking a schema against its metaschema, which is a different beast from the validator that checks an ASDF file against a schema.  It isn't possible for `tuple`, `np.str_`, etc to appear in a schema and we also don't need to handle YAML tags.

jdavies-st: Ah, cool.  I understand now.  Thanks!

jdavies-st (approved):

PR #988, opened 2021-05-04
Update asdftool tags and asdftool extensions for new extension API

### LINKED ISSUE #959: asdftool does not yet account for the new extension API
The `asdftool extensions` and `asdftool tags` subcommands list support available to the library via plugins.  These do not yet include extensions and tag support provided by new extension API.

eslavich:
Long term we may want to drop these, the Python API has plenty of options for getting at this information.

Resolves #959

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -63,6 +63,9 @@
   filenames without recognizable extensions.  Remove the ``asdf.asdf.is_asdf_file``
   function. [#978]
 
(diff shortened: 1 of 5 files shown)


jdavies-st (approved):
Looks good.

These tools are actually pretty useful, though if standard built-in python command line tools make this simple as well, then we should deprecate this, or implement them using those built-in tools.

Issue #989, opened 2021-05-04
Multiple ASDF block indices in file?


lgarrison:
This is a bit of a long shot, but I am working with a collaborator whose pipeline appears to occasionally produce corrupted ASDF files, and I'm trying to figure out how this is occurring.  The symptom is that the files have multiple `#ASDF BLOCK INDEX` sections, and so ASDF fails to load the files (I'm pretty sure that having multiple block indices in a single file should never happen, but please correct me if I'm wrong!).  Has anyone seen this failure mode before?  I think it's likely that the ASDF library itself is not at fault, but instead it's some kind of bad interaction between software components and/or the file system.  The failure does not seem deterministic, but we are still debugging.

Here's a link to an example file; if you grep for `ASDF BLOCK`, you'll find there are two occurrences: https://drive.google.com/file/d/1L8X9V_1oiODgP1V-ZhKRJ-UcDmv6ZHR-/view?usp=sharing

eslavich:
Strange... I haven't seen this before.  The extra `#ASDF BLOCK INDEX` occurs in the middle of the first block's decompressed data:

```python
In [22]: b"ASDF BLOCK INDEX" in af.blocks.get_block(0).data.tobytes()
Out[22]: True
```

but it also occurs in the *compressed* data in the file?  So blosc must not be modifying those bytes?

The checksum of the decompressed data matches:

```python
In [24]: af.blocks.get_block(0).validate_checksum()
Out[24]: True
```

lgarrison:
Okay, I figured it out.  You put me on the right track by observing the extra `BLOCK INDEX` must have been in the data before the file was written, due to the checksum.  The bug occurs when an uninitialized Numpy array contains a recycled region of memory that used to hold a block index.  In this case, we were slicing the Numpy array up to the last initialized value before adding to the tree, but because ASDF writes the whole underlying array, not just the view, the uninitialized region ended up in the file too.  Like so:

```python
a = np.empty(10**5)
a[:10] = 1.  # initialize up to 10
af = asdf.AsdfFile(dict(arr=a[:10]))  # only first 10 elements are in the tree, but all 10^5 elements written
af.write_to('a.asdf')
```

[...]

PR #990, opened 2021-05-06
Search for block index starting from end of file

lgarrison:
An earth-shattering, one-character PR to use the last block index found in the file, in case a pathological data block also contains something that looks like a block index header.  This can especially happen if uninitialized arrays are written to disk, because the uninitialized regions could contain recycled memory that has a block index (which happened in #989).

I confirm this fixes the reading of this file, which exhibits the pathology: https://drive.google.com/file/d/1DJr2qkdC2Qm3tos1cigmbUK8ngFzha5K/view?usp=sharing

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,7 +70,7 @@
   search results. [#981]
 
 - Search for block index starting from end of file. Fixes rare bug when
-  a data block contains a block index. [#989]
+  a data block contains a block index. [#990]
 
 2.7.5 (unreleased)
 ------------------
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -69,6 +69,9 @@
 - Add ``AsdfSearchResult.replace`` method for assigning new values to
   search results. [#981]
 
+- Search for block index starting from end of file. Fixes rare bug when
+  a data block contains a block index. [#989]
```
eslavich, line 73: These are PR numbers and not issue numbers, so this just needs to be bumped up to `#990`

eslavich (approved):
The `r` heard 'round the world!

PR #994, opened 2021-05-10
Update downstream workflow to install jwst with --editable

eslavich:
This fixes the jwst job by installing with --editable, and also cleans up the workflow so that the asdf ref doesn't have to be specified twice.

Here's a test run of this version of the workflow: https://github.com/asdf-format/asdf/actions/runs/829268692

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -3,10 +3,6 @@ name: Downstream
 on:
   workflow_dispatch:
     inputs:
-      asdf_ref:
-        description: asdf ref
-        required: true
-        default: master
       astropy_ref:
         description: astropy ref
         required: true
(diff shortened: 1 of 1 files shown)


jdavies-st (commented):
Is there a reason we are checking out the downstream packages at all for this workflow?  Can't we just install them straight from Github via:

```
pip install "jwst[test] @ git+https://github.com/spactelescope/jwst"
```

I.e. shouldn't they all work in the same way as the downstream `astropy` one does?

There should not be a need for an editable install of `jwst` in order for `pytest --pyargs jwst` to work.  The tests should run if installed in site-packages.

eslavich:
This change was prompted by the fact that jwst tests failed in our downstream job last night:

https://github.com/asdf-format/asdf/runs/2542775069?check_suite_focus=true

I think I see what happened -- the package was installed into site-packages using `pip install .[test]`, but the tests were run from the jwst repo, so the installed package was ignored.

> I.e. shouldn't they all work in the same way as the downstream astropy one does?

No, we need the repository to test ASDF schemas and also for cases where the tests aren't included in the package.

jdavies-st:
If we're doing `--pyargs` builds in downstream packages, we should

1) not clone them, but instead just `pip install` from the repo URL

or if the repo is needed for some reason,

2a) run the tests without `--pyargs`

or

2b) run the tests with `--pyargs`, but make sure to `cd` to some neutral territory so that the installed version is not confused with the local checkout.

eslavich:
I'll give that a try, maybe that'll let me eliminate the special astropy job too...

eslavich:
Nice that seems to have worked: https://github.com/asdf-format/asdf/actions/runs/832918130

jdavies-st (approved):
🚀

Issue #999, opened 2021-06-02
New array overwrites existing unloaded array when block index is not present


eslavich:
This bug was discovered over in https://github.com/spacetelescope/jwst/issues/6039:

```python
import asdf
from numpy.random import random

af = asdf.AsdfFile()
af["arr0"] = 0
af["arr1"] = random(10)
af["arr2"] = random(10)
af.write_to("test.asdf", include_block_index=False)

af = asdf.open("test.asdf")
af["arr0"] = random(5)
af.write_to("test2.asdf")
```

```
TypeError: buffer is too small for requested array
```

Key elements: more than one array in the initial file, no block index present, lazy loading enabled, and the third array is added such that it gets serialized first.  The block manager does not check for more unread blocks when a new one is added, so the new block claims an index that existing views already refer to.

eslavich:
Resolved by #1000

PR #1000, opened 2021-06-09
Fix bug when adding a block to an existing file with no index

eslavich:
This addresses #999 (and therefore also https://github.com/spacetelescope/jwst/issues/6039).

The problem arises when adding a block to an existing file that has no block index.  Ordinarily BlockManager knows the location of each internal block right away due to the index, but when the index is missing it defers locating most of the blocks until they're requested by the user.  When a block is added under those conditions, the BlockManager needs to be careful to locate remaining blocks first, otherwise the new block steals a block number that's already been claimed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 2.9.0 (unreleased)
 ------------------
 
+2.8.1 (unreleased)
+------------------
+
+- Fix bug in block manager when a new block is added to an existing
+  file without a block index. [#1000]
+
 2.8.0 (2021-05-12)
 ------------------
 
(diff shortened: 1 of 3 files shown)


jdavies-st (approved):
Wow, that's a subtle bug.

Does this mean that if I wasn't doing any view fiddling, and a swap in block index occurred, it might have occurred unnoticed (and I would just get the wrong data) before this change?

eslavich:
> Does this mean that if I wasn't doing any view fiddling, and a swap in block index occurred, it might have occurred unnoticed (and I would just get the wrong data) before this change?

I think that's right, we got lucky and saw an error because the new array was too small for the old view.  If it hadn't been (like in the test, all the array have length 10), asdf would have cheerfully swapped in the new array without complaint.

PR #1002, opened 2021-06-09
Don't use pathlib in test

eslavich:


jdavies-st:
What's the problem using `pathlib`?

eslavich:
It doesn't work with asdf in Python 3.5, which the 2.7.x series still supports

PR #1010, opened 2021-08-04
Fix array compression for non-native byte order

### LINKED ISSUE #1009: Array compression fails for non-native byte order
Array compression fails for non-native byte order, for example:

```python
import asdf
import numpy as np

with asdf.AsdfFile() as af:
  af["array"] = np.arange(1024, dtype='>i2')
  af.write_to("test.asdf", all_array_compression='zlib')
```

results in this error:

```
ValueError: memoryview: destination format must be a native single character format prefixed with an optional '@'
```

lgarrison:
Use a round-trip to Numpy to create a 1D `memoryview` that has the right byte order, because `memoryview` can't cast to all types it can represent.

Closes #1009.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@
 
 - Update documentation to reflect new 2.8 features. [#998]
 
+- Fix array compression for non-native byte order [#1010]
+
 2.8.1 (2021-06-09)
 ------------------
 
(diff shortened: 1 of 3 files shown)


eslavich (approved):
Tidy fix, thanks a lot!

PR #1014, opened 2021-08-14
fix from_yaml_tree docstring parameter name

CagtayFabry:
found this little documentation discrepancy while going through the code


eslavich (approved):
Oops, yes, thanks for the fix!

Issue #1015, opened 2021-08-17
NDArrayType attributes when reading from tree


CagtayFabry:
The `NDArrayType` is a bit different from other asdf types in that the `from_tree` returns an instance of the `NDArrayType` (instead of the mapped python types)

This also means that you can access the classattributes `.name='core/ndarray'` and `.version='1.0.0'` from any instance. From my understanding, these attributes are only used for generating the appropriate tag mapping for the asdf extension?

Unfortunately this can have unexpected results when using the loaded array with other libraries that check for these attributes on class creation, see here https://github.com/BAMWelDX/weldx/issues/110

In the attributes are not needed on the instances returned by `from_tree` by asdf after creation, can they be set to `None` so that other code is less likely to pick them up? see https://github.com/CagtayFabry/asdf/commit/603b01664eda3d7f7305d184a859d719b8aad50c

eslavich:
> From my understanding, these attributes are only used for generating the appropriate tag mapping for the asdf extension?

Yes, that's right.  The instance of NDArrayType wraps np.ndarray, while the NDArrayType class methods implement the ExtensionType interface.  I find this terribly confusing.

Ideally I think we'd just separate NDArrayType into two different classes, one for the np.ndarray wrapper and another for the ExtensionType implementation.  I was planning to do that when switching the built-in types to the new converter API, but we can try to get it done earlier if it's causing problems for you.

CagtayFabry:
It's not really a pressing issue since it's rather easy to handle (once you know where to look)

maybe I'll open a PR with the small changes above and we can have a look if it breaks anything in the pipelines?

and totally, seperating the class with the new coverter should make the interfaces much easier to understand

PR #1016, opened 2021-08-17
override NDArrayType class attributes

CagtayFabry:
set instance attributes of `NDArrayType` to `None` when reading from tree as discussed in #1015


CagtayFabry:
could you take another look here @eslavich ?

eslavich:
I'll look later today, we should get this merged before I release 2.8.2.  Thanks for the reminder...

eslavich:
Closing in favor of #1031

PR #1020, opened 2021-09-14
fix `TaggedString` docstring

CagtayFabry:
The docstring doesn't seem 100% accurate since `TaggedString` doesn't have an `__init__` method and one cannot tag it directly upon creating (in contrast to `TaggedList`)


eslavich (approved):
LGTM, thanks!

PR #1023, opened 2021-11-03
update weldx downstream workflow

CagtayFabry:
we have changed the schema file location with a recent release, can you update this @eslavich ?


eslavich (approved):
Thanks for the update!

PR #1029, opened 2021-11-30
Fix schema URI resolving when the URI prefix is also claimed by a legacy extension

### LINKED ISSUE #1028: URIs in custom schemas do not resolve correctly
When resolving references in a custom schema, URIs from gwcs are mangled by the asdf default resolver because they start with the prefix "http://stsci.edu/schemas/".  The `asdf.schema.load_schema` function passes the URI through the resolver without first checking if the URI is available from a new-style extension.

This is a bug that showed up when testing dkist with asdf-astropy.

eslavich:
This fixes a bug where the schema loader looks for content on disk when it should be retrieving it from the ResourceManager instance.

Resolves #1028

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@
 
 - Fix use of 'python' where 'python3' is intended. [#1026]
 
+- Fix schema URI resolving when the URI prefix is also
+  claimed by a legacy extension. [#1029]
+
 2.8.1 (2021-06-09)
 ------------------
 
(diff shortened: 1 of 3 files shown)


eslavich:
This passes the jwst regression tests (one unrelated failure):

https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/143/

perrygreenfield (approved):
LGTM

PR #1031, opened 2021-12-02
Remove 'name' and 'version' attributes from NDArrayType instances

### LINKED ISSUE #1015: NDArrayType attributes when reading from tree

eslavich:
@CagtayFabry what do you think about this alternative fix?  My concern with #1016 is that we won't catch all the cases where an NDArrayType is created.

Resolves #1015

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@
 - Fix schema URI resolving when the URI prefix is also
   claimed by a legacy extension. [#1029]
 
+- Remove 'name' and 'version' attributes from NDArrayType
+  instances. [#1031]
+
 2.8.1 (2021-06-09)
 ------------------
 
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -368,6 +368,18 @@ class NDArrayType(AsdfType):
             self._array = None
             raise e from None
 
+    def __getattribute__(self, name):
+        # The presence of these attributes on an NDArrayType instance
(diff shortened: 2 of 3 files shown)


CagtayFabry:
> @CagtayFabry what do you think about this alternative fix? My concern with #1016 is that we won't catch all the cases where an NDArrayType is created.
> 
> Resolves #1015

this looks like a much cleaner solution, thanks!

perrygreenfield:
Just to clarify, the idea here is to effectively remove these attributes from the class instances, but not the class itself (which the tag machinery would use)?

eslavich:
> Just to clarify, the idea here is to effectively remove these attributes from the class instances, but not the class itself (which the tag machinery would use)?

Yes, just so!

perrygreenfield (approved):
LGTM

Issue #1018, opened 2021-08-25
[2.8.1] Tests depend on the "python" executable name


olebole:
When running the tests on a standard Debian machine, I get a lot of errors like

```
______________________ test_asdf_open_failure[version4] _______________________
tmp_path = PosixPath('/tmp/pytest-of-debci/pytest-0/test_asdf_open_failure_version4')
create_editor = <function create_editor.<locals>._create_editor at 0x7f502e72d040>
version = AsdfVersion('1.4.0')
mock_input = <function mock_input.<locals>._mock_input at 0x7f502e72dca0>
    def test_asdf_open_failure(tmp_path, create_editor, version, mock_input):
        file_path = str(tmp_path/"test.asdf")
    
        with asdf.AsdfFile(version=version) as af:
            af["foo"] = "bar"
            af.write_to(file_path)
    
        os.environ["EDITOR"] = create_editor(r"^#ASDF .*?$", "#HJKL 1.0.0")
    
        with file_not_modified(file_path):
            with mock_input(r"\(c\)ontinue editing or \(a\)bort\?", "a"):
>               assert main.main_from_args(["edit", file_path]) == 1
/usr/lib/python3/dist-packages/asdf/commands/tests/test_edit.py:311: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3/dist-packages/asdf/commands/main.py:70: in main_from_args
    result = args.func(args)
/usr/lib/python3/dist-packages/asdf/commands/edit.py:62: in run
    return edit(args.filename)
/usr/lib/python3/dist-packages/asdf/commands/edit.py:213: in edit
    open_editor(temp_file.name)
/usr/lib/python3/dist-packages/asdf/commands/edit.py:372: in open_editor
    subprocess.run(f"{editor} {path}", check=True, shell=True) # nosec
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
input = None, capture_output = False, timeout = None, check = True
popenargs = ('python /tmp/pytest-of-debci/pytest-0/test_asdf_open_failure_version4/editor.py /tmp/test-_h78phud.yaml',)
kwargs = {'shell': True}
[...]

PR #1033, opened 2021-12-13
Use python3 as executable instead of python in tests

olebole:
Even after #1026, there are still some appearances of executing "python" instead of "python3" self. This PR replaces them and so completes fixing #1018. 

This is tested with the Debian build of python-asdf.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 2.9.0 (unreleased)
 ------------------
 
+- Fix more use of 'python' where 'python3' is intended. [#1033]
+
 2.8.2 (2021-12-06)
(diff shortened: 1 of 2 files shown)


eslavich:
Oh no!  I'm sorry, I had assumed that only recently added tests had this problem.  I'll merge this and do another release this morning.

eslavich:
I investigated the gwcs dev failure and tracked it down to a problem in asdf-astropy and the compound transform schemas (not anything we can fix in asdf itself).

eslavich (approved):

eslavich:
This has been released in 2.8.3

PR #1034, opened 2021-12-13
Fix tag listing when extension is not fully implemented

eslavich:
This fixes a bug in the `asdftool tags` command when an extension does not implement all the tags in its manifest.

I'm also removing two tests that involve gwcs that haven't really been testing anything since support for WCS serialization was removed from this library.  These tests fail with gwcs master due to the removal of the gwcs.tags module.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+2.8.4 (unreleased)
+------------------
+
+- Fix tag listing when extension is not fully implemented. [#1034]
+
 2.8.3 (2021-12-13)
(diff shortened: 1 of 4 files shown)


nden (approved):
As a side question, did we ever remove the subclassing code form asdf?

eslavich:
> As a side question, did we ever remove the subclassing code form asdf?

It hasn't been removed, but I didn't implement the feature in the new extension code.  As people transition over we'll either hear complaints about that, or (hopefully) hear nothing.

PR #1035, opened 2021-12-13
Update asdf-standard pointer

eslavich:
This updates the asdf-standard pointer to include commits from https://github.com/asdf-format/asdf-standard/pull/289, which will resolve schema-related failures with gwcs dev.


eslavich:
The remaining two gwcs dev failures will be resolved by https://github.com/astropy/asdf-astropy/pull/36

Issue #1037, opened 2021-12-13
Update warnings-as-exceptions CI job to use Python 3.10


eslavich:
See https://github.com/asdf-format/asdf/pull/1036

PR #1038, opened 2021-12-13
Unpin Pytest < 6

bnavigator:
pytest-doctestplus has been compatible with pytest 6 since version 0.8

As Pytest 5 is not compatible with Python 3.10, this is a blocker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
           - name: Python 3.10 Testing
             os: ubuntu-latest
             python-version: 3.10.0
-            toxenv: py38
+            toxenv: py310
 
           - name: Code Coverage with Python 3.9
             os: ubuntu-latest
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,9 +44,7 @@ docs =
     matplotlib
     docutils
 tests =
-    # As of 2020-07-28, pytest-doctestplus is not compatible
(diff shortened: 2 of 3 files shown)


bnavigator:
Oh wow. That's quite some fallout with the changed importing mechanism in Pytest 6.

eslavich:
I'm not sure, but I think adding this to tox.ini under `[testenv]` will help:

```
usedevelop= true
```

bnavigator:
> I'm not sure, but I think adding this to tox.ini under `[testenv]` will help:
> 
> ```
> usedevelop= true
> ```

Correct! Thank you!

eslavich:
Hooray!  The `Twine` and `Test Against Installed Packaged` jobs should be overridden with usedevelop= false again.  I don't know what's going on with the "Warnings Treated as Exceptions` job, but I can investigate that in the morning if you don't beat me to it.

eslavich:
#1037 refers to updating these to use 3.10:

https://github.com/asdf-format/asdf/blob/master/.github/workflows/ci.yml#L120
https://github.com/asdf-format/asdf/blob/master/tox.ini#L41

eslavich (approved):
🎉

eslavich:
@bnavigator do you need a release of these fixes?

bnavigator:
No, I already worked around it in the rpm package. After all this is only testing, nothing facing the user.

PR #1041, opened 2021-12-30
MNT: Replace distutils with packaging

pllim:
`distutils` is deprecated. cc @eslavich

### DIFF
--- a/asdf/util.py
+++ b/asdf/util.py
@@ -321,7 +321,7 @@ def minversion(module, version, inclusive=True, version_path='__version__'):
 
     By default this uses `pkg_resources.parse_version` to do the version
     comparison if available.  Otherwise it falls back on
-    `distutils.version.LooseVersion`.
+    `packaging.version.Version`.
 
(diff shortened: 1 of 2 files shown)


pllim:
Hmm I don't understand the error.

`error: package directory 'asdf-standard/resources' does not exist`

eslavich:
> error: package directory 'asdf-standard/resources' does not exist

This is just the submodule giving us grief again... @WilliamJamieson has a plan to get rid of it and replace with a normal Python dependency, hopefully soon.

pllim:
As discussed offline, @WilliamJamieson will take over this PR. I think the compatibility test needs to be rewritten because `packaging.version.Version` is not a drop-in replacement for `distutils.StrictVersion`. Care needs to be given to return data type (unless backward compatibility for this module is not a problem) and comparison logic.

pllim:
Ah, I was sooo close. 😆 

Thanks for wrapping this up, @WilliamJamieson !

PR #1042, opened 2022-01-06
Move complex serialization support to Converter

eslavich:
This PR begins the process of moving serialization support for core objects to the Converter API.

- Add new-style Extensions for the core manifests
- Create a Converter for complex types
- Add some new test helpers that are compatible with Converter
- Remove or xfail tests that are not relevant to the new API


WilliamJamieson (approved):
LGTM

PR #1043, opened 2022-01-06
Drop support for Python 3.6 and 3.7

eslavich:
This PR drops support for Python 3.6 and Python 3.7 and removes tests and test environments that are only relevant to these versions.


WilliamJamieson (approved):
LGTM

PR #1044, opened 2022-01-10
Move large integer support to Converter

eslavich:
This moves support for large-magnitude integer literals to Converter, in the process dropping the IntegerType class and representing even large integers as regular builtins.int instead.  The user no longer needs to manually wrap large integers in a special class; the Converter handles this transparently as needed.


WilliamJamieson (approved):
LGTM

PR #1045, opened 2022-01-11
Drop support for core/subclass_metadata-1.0.0 and core/constant-1.0.0 tags

eslavich:
We removed the feature that used subclass metadata, and AFAIK the core/constant-1.0.0 tag has never been used.  If there are users out there with files containing these tags, they'll see a warning when opening the file but the object will still be available as a dict.


WilliamJamieson (approved):
LGTM

PR #1046, opened 2022-01-11
Move support for ExternalArrayReference to Converter

eslavich:
This moves support for serializing ExternalArrayReference to a Converter class, and also changes ExternalArrayReference itself from a dict to a dataclass.

@Cadair I think you are the primary user of this class, do these changes work for you?  The most significant change I think is that the class will no longer be available from `asdf.ExternalArrayReference`, it'll have to be imported from `asdf.core.ExternalArrayReference` instead.  This will only go into asdf 3.0 so you can pin to 2.x to avoid it.


WilliamJamieson (approved):
LGTM

PR #1047, opened 2022-01-12
Move ExtensionMetadata and friends to Converter

eslavich:
This moves ExtensionMetadata, HistoryEntry, and Software to the Converter API (and also makes them dataclasses instead of dicts).  Branched off of #1046 and relies upon https://github.com/asdf-format/asdf-standard/pull/290, so we'll need to merge those two first.


WilliamJamieson (approved):
LGTM

PR #1048, opened 2022-01-12
Move support for AsdfObject to Converter

eslavich:
This moves AsdfObject serialization to the Converter API.  There will probably be more work we want to do on AsdfObject for 3.0, but for now I'm just leaving it as a dict the way it was originally.


asdf/core/_converters/entities.py
```diff
@@ -22,3 +35,112 @@ def from_yaml_tree(self, node, tag, ctx):
             datatype=node["datatype"],
             shape=tuple(node["shape"]),
         )
+
+
+class SoftwareConverter(Converter):
+    tags = ["tag:stsci.edu:asdf/core/software-*"]
+    types = ["asdf.core._entities.Software"]
```
CagtayFabry, line 42: A quick (PR unrelated) question regarding the tag support here if I may @eslavich
Is supporting 'all' versions considered best practice for ASDF schemas/tags?

I guess this is somewhat pending with the 'revised' schema versioning guideline [here](https://github.com/asdf-format/asdf-standard/wiki/2.0.0-roadmap#review-schema-versioning-policy)

eslavich: That's a good point, I've been using the wildcard without thinking much about it.  Better think through a couple of scenarios here:

**Breaking changes to the structure of software-x.y.z**

This code will crash with KeyError trying to access properties that don't exist.  The file will not open.

**Add a new required property to software**

This code will open the file but discard the required property.  On update/write_to, the information will be lost, and the user will not be warned.

**Add an optional validated property to software**

Same story, the code will lose the new property.

Hmm!  There doesn't seem to be any advantage to using the wildcard here, do you agree?  I think your comment was a PR comment after all :)

CagtayFabry: Yes good points, I have been thinking about similar 'problems' as well regarding our weldx schemas
I think it is sensible to avoid support for **all** versions with `-*` precisely for the reasons you mentioned.

My personal approach right now for the upcoming weldx 1.0.0 release is somewhere along this line:
- specify/pin at least the supported major version in the converter tags so that in case some breaking changes are necessary there is at least one way to fail with the correct error message when old code encounters new files
- I think leaving only the patch version as wildcard is 'truest' to the schemantic versioning for converters (but can be quite some work when things change often)
- in schemas using the `tag:` validator I tend to wildcard minor and patch version like `-1.*` since I feel it is often more about the object reference than its specific implementation (handled by the converter)

eslavich: The answer is probably specific to the tag/type, for example the ExtensionMetadata class I added knows that additional properties are possible and will retain them and write them back out again.  In the case of that class a wildcard would be fine for minor versions since it won't lose any properties.  For the Software class though any additional properties will be lost so it's better to fail.

Now that I write that though I'm wondering if I should be more careful in general about additional properties, the schema technically allows them for software-1.0.0...

WilliamJamieson (approved):
LGTM

PR #1049, opened 2022-01-13
Remove wildcard from converter tags

eslavich:
@CagtayFabry's question [here](https://github.com/asdf-format/asdf/pull/1048#discussion_r783782316) revealed some flaws in the new Converters.  I've updated them to claim specific tag versions and also to retain additional properties in an "extra" dict where the schema allows them.


WilliamJamieson (approved):
LGTM

PR #1050, opened 2022-01-18
Add new plugin type for custom schema validators

### LINKED ISSUE #1012: adding custom validators using new extension API
What is the correct way to register custom validators now with the new extension API?

The documentation only mentions this in the deprecated old API section.
https://asdf.readthedocs.io/en/latest/asdf/extending/legacy.html#adding-custom-validators

eslavich:
This adds a new plugin type for custom schema validators.  We need this in order to implement support for ndarray using the Converter API.

Resolves #1012


eslavich:
@CagtayFabry will this work for your use case?  The URI patterns seem to nicely resolve the issues we were discussing around the scope of the custom validator.

CagtayFabry:
thank you for starting this @eslavich 

I will have to take a closer look but from a quick scan it looks good to me.
Regarding URI pattern association: in the example listed if I wanted to 'enable' (associate?) the `MaxAreaValidator` for all schemas listed unter `/tags/` I would do this?
```python
from asdf.extension import Validator
from asdf.exceptions import ValidationError

class MaxAreaValidator(Validator):
    schema_property = "max_area"
    tags = ["asdf://example.com/example-project/tags/**"]
```

eslavich:
> Regarding URI pattern association: in the example listed if I wanted to 'enable' (associate?) the MaxAreaValidator for all schemas listed unter /tags/ I would do this?

Yes, that's right!

CagtayFabry:
Great, with careful versioning syntax this should also effectively enable versioning of the validators :+1:

I'm quite busy these days but I will try a quick demo implementation with our weldx validators in the next 2 weeks to give feedback, would that work @eslavich ?

eslavich:
> I'm quite busy these days but I will try a quick demo implementation with our weldx validators in the next 2 weeks to give feedback, would that work @eslavich ?

Yes, that'd be super helpful, thank you!  We might merge this in the meantime, but it's only going into an epic branch and not into master, so there will still be plenty of time to adjust the code based on your feedback.

asdf/extension/_extension.py
```diff
@@ -6,6 +6,8 @@
 from ._tag import TagDefinition
 from ._legacy import AsdfExtension
 from ._converter import ConverterProxy
+from ._compressor import Compressor
```
WilliamJamieson, line 9: The `Compressor` changes should be moved to a separate PR if possible.

asdf/extension/_extension.py
```diff
@@ -92,7 +94,7 @@ def compressors(self):
 
         Returns
         -------
-        iterable of asdf.extension.Compressor instances
+        iterable of asdf.extension.Compressor
```
WilliamJamieson, line 97: More `Compressor` changes.

asdf/extension/_extension.py
```diff
@@ -178,7 +192,20 @@ def __init__(self, delegate, package_name=None, package_version=None):
...
         # properties to already be available.
         self._converters = [ConverterProxy(c, self) for c in getattr(self._delegate, "converters", [])]
-        self._compressors = self._delegate.compressors if hasattr(self._delegate, "compressors") else []
+
+        self._compressors = []
+        if hasattr(self._delegate, "compressors"):
+            for compressor in self._delegate.compressors:
+                if not isinstance(compressor, Compressor):
+                    raise TypeError("Extension property 'compressors' must contain instances of asdf.extension.Compressor")
+                self._compressors.append(compressor)
```
WilliamJamieson, line 201: More `Compressor` changes.

asdf/extension/_manifest.py
```diff
@@ -16,6 +16,12 @@ class ManifestExtension(Extension):
     converters : iterable of asdf.extension.Converter, optional
         Converter instances for the tags and Python types
         supported by this extension.
+    compressors : iterable of asdf.extension.Compressor, optional
+        Compressor instances to support additional binary
+        block compression options.
```
WilliamJamieson, line 21: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -178,6 +197,8 @@ def test_extension_proxy():
     assert subclassed_proxy.legacy_class_names == proxy.legacy_class_names
     assert subclassed_proxy.asdf_standard_requirement == proxy.asdf_standard_requirement
     assert subclassed_proxy.converters == proxy.converters
+    assert subclassed_proxy.compressors == proxy.compressors
```
WilliamJamieson, line 200: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -195,9 +216,16 @@ def test_extension_proxy():
             types=[]
         )
     ]
+    validators = [
+        MinimalValidator()
+    ]
+    compressors = [
+        MinimalCompressor()
+    ]
```
WilliamJamieson, line 224: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -195,9 +216,16 @@ def test_extension_proxy():
...
     extension = FullExtension(
         converters=converters,
-        compressors=[MinimalCompressor],
+        compressors=compressors,
```
WilliamJamieson, line 227: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -620,18 +655,22 @@ def from_yaml_tree(self, *args):
                 pass
 
         converter = FooConverter()
+        validator = MinimalValidator()
+        compressor = MinimalCompressor()
```
WilliamJamieson, line 659: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -620,18 +655,22 @@ def from_yaml_tree(self, *args):
...
 
         extension = ManifestExtension.from_uri(
             "asdf://somewhere.org/extensions/foo",
             legacy_class_names=["foo.extension.LegacyExtension"],
             converters=[converter],
-            compressors=[MinimalCompressor],
+            compressors=[compressor],
```
WilliamJamieson, line 665: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -620,18 +655,22 @@ def from_yaml_tree(self, *args):
...
+            validators=[validator],
         )
         assert extension.extension_uri == "asdf://somewhere.org/extensions/foo"
         assert extension.legacy_class_names == ["foo.extension.LegacyExtension"]
         assert extension.asdf_standard_requirement == SpecifierSet(">=1.6.0,<2.0.0")
         assert extension.converters == [converter]
-        assert extension.compressors == [MinimalCompressor]
+        assert extension.compressors == [compressor]
```
WilliamJamieson, line 672: More `Compressor` changes.

asdf/tests/test_extension.py
```diff
@@ -644,7 +683,8 @@ def from_yaml_tree(self, *args):
         assert proxy.legacy_class_names == {"foo.extension.LegacyExtension"}
         assert proxy.asdf_standard_requirement == SpecifierSet(">=1.6.0,<2.0.0")
         assert proxy.converters == [ConverterProxy(converter, proxy)]
-        assert proxy.compressors == [MinimalCompressor]
+        assert proxy.compressors == [compressor]
```
WilliamJamieson, line 686: More `Compressor` changes.

WilliamJamieson (changes requested):

WilliamJamieson (approved):
Looks good to me.

CagtayFabry:
a few notes from me @eslavich 

I managed to implement the new API [in a quick draft](https://github.com/BAMWelDX/weldx/compare/master...CagtayFabry:asdf-3.0-validators), so far it seems to work 👍 

One issue that came up where I am not sure if the behavior was actually a a bug in the legacy API:
with the old API the validator would trigger on **any node** in the ASDF tree that had the keyword/validator-property.

Tha also meant that the validator would be in effect for non tagged nodes, here is a simple example from my test cases:

```yaml
%YAML 1.1
---
id: "asdf://weldx.bam.de/weldx/schemas/debug/test_unit_validator-0.1.0"

type: object
properties:
  length_prop:
    description: |
      a simple length quantity with unit validator
    tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.*"
    wx_unit: "m"

  simple_prop:
    description: simple property without any tag and unit validator
    type: object
    properties:
      value:
        type: number
      units:
        type: string
    wx_units: "kg"
```
Previously the `wx_unit` validator would trigger on both the `length_prop` and the `simple_prop`
Now it only gets triggered on the `length_prop`, when debugging it never enters the `validate` function.

Is this intended? The ability to enable the validator on everything was quite useful.

Do you see any way to also implement this old behavior (I tried not setting anything for `tags` in `Validator` but no luck, I didn't look into your code yet)

On a related note: For some reason the validator gets triggered when the keyword is put on the top level of the schema (instead of next to the individual properties)
```yaml
%YAML 1.1
---
id: "asdf://weldx.bam.de/weldx/schemas/debug/test_unit_validator-0.1.0"

type: object
properties:
  length_prop:
    description: |
      a simple length quantity with unit validator
    tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.*"
    wx_unit: "m"

  simple_prop:
    description: simple property without any tag and unit validator
    type: object
    properties:
      value:
        type: number
      units:
        type: string
    wx_units: "kg"

wx_unit:
  length_prop: "m"
  simple_prop: "kg"
```
The final `wx_unit` in this gets validated.

eslavich:
Thanks for the notes @CagtayFabry!

> Is this intended?

Yes, we wanted to support validators scoped to certain tags, so that different developers could share the same validator name without worrying as much about collisions.

> Do you see any way to also implement this old behavior

The wildcard tag pattern `"**"` should match anything, see https://github.com/asdf-format/asdf/blob/asdf-3.0-epic/asdf/extension/_manager.py#L291.  Thinking about this again I was a bit lazy there and should have used built-in tags like `tag:yaml.org,2002:map` for dict, list, etc, so that the tag pattern could select one of those specifically.  I'll make a note to fix that...

> For some reason the validator gets triggered when the keyword is put on the top level of the schema 

That's really strange!  I'll have to look into that one.

PR #1052, opened 2022-01-24
added support for __asdf_traverse__ method to allow custom introspect…

perrygreenfield:
This PR allows the asdf info functionality to check tag classes for a method __asdf_traverse__. If that method exists, then it is called whereupon it should return information that the info method expects. This allows any tag class to present information about its contents to the `info` method. 

The original purpose was to allow Roman data files to be displayed just like JWST files, which previously wasn't possible since the Roman asdf tree uses tag classes unlike the JWST tree. Roman_datamodels has been correspondingly updated to take advantage of this new capability.

### DIFF
--- a/asdf/_display.py
+++ b/asdf/_display.py
@@ -99,12 +99,11 @@ class _NodeInfo:
                         parent.children.append(info)
                     seen.add(id(node))
                     if cls.supports_info(node):
-                        info.tag = node._tag
-                        for child_identifier, child_node in node.__asdf_traverse__():
-                            next_nodes.append((info, child_identifier, child_node))
+                        tnode = node.__asdf_traverse__()
                     else:
-                        for child_identifier, child_node in get_children(node):
-                            next_nodes.append((info, child_identifier, child_node))
+                        tnode = node
+                    for child_identifier, child_node in get_children(tnode):
+                        next_nodes.append((info, child_identifier, child_node))
 
             if len(next_nodes) == 0:
                 break
--- a/asdf/_display.py
+++ b/asdf/_display.py
@@ -88,7 +88,10 @@ class _NodeInfo:
             next_nodes = []
 
             for parent, identifier, node in current_nodes:
-                if (isinstance(node, dict) or isinstance(node, list) or isinstance(node, tuple)) and id(node) in seen:
+                if (isinstance(node, dict) or 
+                    isinstance(node, list) or 
+                    isinstance(node, tuple) or
+                    cls.supports_info(node)) and id(node) in seen:
                     info = _NodeInfo(parent, identifier, node, current_depth, recursive=True)
                     parent.children.append(info)
                 else:
--- a/asdf/_display.py
+++ b/asdf/_display.py
@@ -129,7 +129,7 @@ class _NodeInfo:
 
     def supports_info(node):
         """
-        This method determines if the node is an instance of a class that 
+        This method determines if the node is an instance of a class that
         supports introspection by the info machinery. This determined by
         the presence of a __asdf_traverse__ method.
         """
--- a/asdf/_display.py
+++ b/asdf/_display.py
@@ -88,8 +88,7 @@ class _NodeInfo:
             next_nodes = []
 
             for parent, identifier, node in current_nodes:
-                if (isinstance(node, dict) or 
-                    isinstance(node, list) or 
+                if (isinstance(node, dict) or
                     isinstance(node, tuple) or
                     cls.supports_info(node)) and id(node) in seen:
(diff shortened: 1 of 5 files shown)


asdf/_display.py
```diff
@@ -1,5 +1,11 @@
 """
 Utilities for displaying the content of an ASDF tree.
+
+Normally these tools only will introspect dicts, lists, and primative values
```
eslavich, line 4: `primative` --> `primitive`

asdf/_display.py
```diff
@@ -1,5 +1,11 @@
...
+(with an exception for arrays). However, if the objecct that is generated
```
eslavich, line 5: `objecct` --> `object`

asdf/_display.py
```diff
@@ -114,6 +125,15 @@ def __init__(
         self.recursive = recursive
         self.visible = visible
         self.children = []
+        self.tag = tag
+
+    def supports_info(node):
```
eslavich, line 130: This is being called as a class method up above, but it's not decorated as such and doesn't have a `cls` argument.  On the other hand, it seems to be working...

asdf/_display.py
```diff
@@ -105,7 +115,8 @@ def from_root_node(cls, root_identifier, root_node):
         return root_info
 
     def __init__(
-        self, parent, identifier, node, depth, recursive=False, visible=True
+        self, parent, identifier, node, depth, recursive=False, visible=True,
+        tag=None
```
eslavich, line 119: Do we need this new `tag` argument?

eslavich (commented):
Wow, that turned out nice and tidy!  Can you add a brief section to readthedocs too?

perrygreenfield:
I believe I addressed the comments

CagtayFabry:
this functionality is great, looking forward to giving it a try, thanks @perrygreenfield !

WilliamJamieson (approved):
This seems good to me.

asdf/_display.py
```diff
@@ -92,9 +98,13 @@ def from_root_node(cls, root_identifier, root_node):
                     if parent is not None:
                         parent.children.append(info)
                     seen.add(id(node))
-
-                    for child_identifier, child_node in get_children(node):
-                        next_nodes.append((info, child_identifier, child_node))
+                    if cls.supports_info(node):
+                        info.tag = node._tag
+                        for child_identifier, child_node in node.__asdf_traverse__():
```
eslavich, line 103: Something I missed before, if we call `get_children` on the result of `__asdf_traverse__`, that will allow `__asdf_traverse__` to return a dict and have it processed correctly.

asdf/tests/test_api.py
```diff
@@ -557,6 +557,27 @@ def test_info_asdf_file(capsys, tmpdir):
     assert "baz" in captured.out
 
 
+class ObjectWithInfoSupport:
+
+    def __init__(self):
+        self._tag = "foo"
+
+    def __asdf_traverse__(self):
+        return list({'the_meaning_of_life_the_universe_and_everything': 42,
```
eslavich, line 566: If we add `get_children` above, we can return a dict here

docs/asdf/extending/extensions.rst
```diff
@@ -184,6 +184,19 @@ an additional list of class names that previously identified the extension:
...
         ]
 
+.. _exposing_extension_object_internals:
+
+Making converted object's contents visible to `info` and `search`
+-----------------------------------------------------------------
+
+If the object produced by the extension supports a class method
+`.__asdf_traverse__` then it can be used by those tools to expose the contents
+of the object. That method should accept no arguments and return a list.
```
eslavich, line 194: Mention dict here if you decide to support that

asdf/_display.py
perrygreenfield: I'm not sure what you mean here, particularly with _if we call_ `get_children` on the result. Doesn't `__asdf_traverse__` return the same kind of result that `get_children` does? I may be slow this morning. It is Wednesday.

eslavich: The way this is set up currently, the `__asdf_traverse__` method must return dict.items() if it wants to provide dict items, and enumerate(list) if it wants to provide list items.  If we instead call `get_children(__asdf_traverse__())` here, then `__asdf_traverse__` can return simple dict or list and have everything work correctly (get_children will take care of calling .items() on the dict and wrapping the list in enumerate).

perrygreenfield: OK, now I understand. I didn't realize you wanted to change what `__asdf_traverse__` should return, I'll make the corresponding changes. It would be a simpler requirement on the method.

eslavich (approved):
LGTM!

PR #1054, opened 2022-01-27
Drop support for Python 3.6

eslavich:
Our Python 3.6 CI job has started failing due to dependencies that require > Python 3.6, so it seems like a reasonable time to drop support for Python 3.6 in this package as well.


WilliamJamieson (approved):
LGTM

PR #1057, opened 2022-01-28
Add error message to setup.py.

WilliamJamieson:
If one runs `pip install .` without running `git submodule update --init`, then pip spits out an incomprehensible error due to `asdf-standard` being empty. This PR adds a check that for `asdf-standard` being empty, with an error raised which describes what to do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,11 @@
 import os
 from setuptools import setup, find_packages
 
+if not os.listdir('./asdf-standard'):
+    from setuptools.errors import SetupError
+
+    raise SetupError("asdf-standard is empty. Need to run `git submodule update --init` and try again!")
+
 
 packages = find_packages()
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,6 +2,11 @@
 import os
 from setuptools import setup, find_packages
 
+if not os.listdir('./asdf-standard'):
```
eslavich, line 5: I'm not sure if this script is always executed with asdf as the working directory, perhaps we should be defensive here and list the directory relative to this file's location:

```python
if not any((Path(_file_).parent / "asdf-standard").iterdir()):
  # ...
```

WilliamJamieson: See 7bf4fd3.

eslavich (approved):
LGTM!

PR #1060, opened 2022-01-31
Remove various ignore flags/arguments from ASDF

WilliamJamieson:
Removes the following flags/arguments:

- `ignore_version_mismatch`
- `ignore_unrecognized_tag`
- `ignore_inplicit_conversion`
for `AsdfFile`.

This is part of the process to make asdf-3.0 more strict.

### DIFF
--- a/asdf/treeutil.py
+++ b/asdf/treeutil.py
@@ -3,9 +3,11 @@ Utility functions for managing tree-like data structures.
 """
 
 import types
+import warnings
 from contextlib import contextmanager
 
 from . import tagged
+from .exceptions import AsdfWarning
 
 __all__ = ["walk", "iter_tree", "walk_and_modify", "get_children", "is_container", "PendingValue"]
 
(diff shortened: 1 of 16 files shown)


asdf/treeutil.py
```diff
@@ -357,9 +348,6 @@ def _handle_immutable_sequence(node, json_id):
             # The derived class signature is different, so simply store the
             # list representing the contents. Currently this is primarly
             # intended to handle namedtuple and NamedTuple instances.
-            if not ignore_implicit_conversion:
```
eslavich, line 360: Hmm... perhaps we should raise an error and halt serialization in this case?  It seems hazardous to perform any kind of implicit conversion when writing a file.

asdf/yamlutil.py
```diff
@@ -310,9 +309,6 @@ def _walker(node):
         tag_type = ctx.type_index.from_yaml_tag(ctx, tag, _serialization_context=_serialization_context)
         # This means the tag did not correspond to any type in our type index.
         if tag_type is None:
-            if not ctx._ignore_unrecognized_tag:
-                warnings.warn("{} is not recognized, converting to raw Python "
```
eslavich, line 314: I think we ought to keep this warning, just remove the flag that disables it.  Users can hide the warning with the built-in warnings module.

asdf/treeutil.py
WilliamJamieson: See 1bf58ce.

asdf/yamlutil.py
WilliamJamieson: See 1bf58ce.

eslavich (approved):
LGTM, feel free to merge after resolving the conflicts

PR #1062, opened 2022-02-01
Remove `extensions` argument from `AsdfFile`

WilliamJamieson:
This PR removes the out-dated `extensions` argument from the `AsdfFile` class, for passing in user specified extensions. This interface is being replaced with the context manager `asdf.config_context` combined with the `add_extension` interface.


eslavich:
Oops, just noticed the flake8 failure, can you fix that before we merge?

PR #1063, opened 2022-02-01
DOC: Improve AsdfInFits example

pllim:
`astropy.io.fits` complains when you do not include a Primary extension. Also, DQ array is usually 16-bit int (also `arange` does not make sense for DQ).


eslavich (approved):
LGTM, thanks!

pllim:
p.s. The comment in the same docstring about `fits:SCI,0` makes no sense to me. Would be nice if you can use that in the example too but I don't know enough about ASDF to make that change here. FYI.

PR #1065, opened 2022-02-02
Add uri to SerializationContext

Cadair:
Seems easy enough?

If this is merged, it would really be lovely if it could be backported to 2.8 :wink:

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -1576,7 +1576,7 @@ class AsdfFile:
     # This function is called from within yamlutil methods to create
     # a context when one isn't explicitly passed in.
     def _create_serialization_context(self):
-        return SerializationContext(self.version_string, self.extension_manager)
+        return SerializationContext(self.version_string, self.extension_manager, self.uri)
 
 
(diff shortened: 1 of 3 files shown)


asdf/asdf.py
```diff
@@ -1734,12 +1734,24 @@ class SerializationContext:
...
         self._version = validate_version(version)
         self._extension_manager = extension_manager
+        self._uri = uri
 
         self.__extensions_used = set()
 
+    @property
+    def uri(self):
+        """
+        The URI (if any) of the file being read.
```
eslavich, line 1747: Suggested docstring:

```
"""
The URL (if any) of the file being read or written.  Used to compute relative locations
of external files referenced by this ASDF file.  The URL will not exist in some cases
(e.g. when the file is written to an io.BytesIO).

Returns
--------
str or None
"""
```

eslavich (approved):
Thanks @Cadair!  I think I'm going to bump a minor version since we're adding new features in this release, so this will go out in 2.9.0 (which will still have support for new and old extensions).

PR #1069, opened 2022-02-02
Restore TagDefinition.schema_uri property but add deprecation warning

eslavich:
@WilliamJamieson I was wrong about this property, it is in fact in use by roman_datamodels.  I'm adding it back here (with a deprecation warning) so that roman_datamodels can have a chance to switch to the new property.

### DIFF
--- a/asdf/tests/test_extension.py
+++ b/asdf/tests/test_extension.py
@@ -17,6 +17,7 @@ from asdf.extension import (
 )
 
 from asdf import config_context
+from asdf.exceptions import AsdfDeprecationWarning
 from asdf.types import CustomType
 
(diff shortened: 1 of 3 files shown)


WilliamJamieson (approved):
LGTM

asdf/tests/test_extension.py
```diff
@@ -437,6 +438,21 @@ def test_tag_definition():
 
     assert "URI: asdf://somewhere.org/extensions/foo/tags/foo-1.0" in repr(tag_def)
 
+    with pytest.warns(DeprecationWarning):
```
WilliamJamieson, line 441:
```suggestion
    with pytest.warns(AsdfDeprecationWarning):
```

PR #1076, opened 2022-02-07
Remove schema version parsing from the pytest plugin

eslavich:
This fixes a bug that causes the schema tests to fail when the schemas are not located in a directory named "schemas".

I'm also removing checks related to astropy 3.0 that are no longer relevant and changing the plugin to use the latest ASDF Standard version always, so that we don't need to worry about inferring the tag URI from the schema and parsing the version.


WilliamJamieson (approved):
LGTM

PR #1079, opened 2022-02-11
Switch to using `asdf-standard` as a PyPi package.

WilliamJamieson:
PR asdf-format/asdf-standard#292 packages asdf-standard so that it can be released on PyPi. Once `asdf-standard` is available as a pypi package, the `git submodule` used to include asdf-standard into asdf will no longer be required. This PR aims to accomplish this.


eslavich (approved):
LGTM!

PR #1082, opened 2022-02-15
Use pre-commit

WilliamJamieson:
Adds [pre-commit](https://pre-commit.com/) support to asdf standard. Using pre-commit will allow for automatic running of codestyle tools on repository when committing files. Moreover, this allows for the bundling of all the codestyle checks into a single step of the CI workflow.


CagtayFabry:
just out of curiosity, are there any plans to add a code formatter such as black  or similar to `asdf` in the near future or maybe with the 3.0 release?

WilliamJamieson:
> just out of curiosity, are there any plans to add a code formatter such as black  or similar to `asdf` in the near future or maybe with the 3.0 release?

Yes, my goal is to introduce black and isort. They will be part of 3.0 or possibly an earlier release.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  hooks:
+  - id: check-added-large-files
+  - id: check-case-conflict
+  - id: check-yaml
+    args: ["--unsafe"]
+  - id: debug-statements
+  - id: end-of-file-fixer
+    exclude: ".*fits"
+  - id: trailing-whitespace
+    exclude: ".*fits"
```
eslavich, line 14: Do we need to exclude .asdf files here as well?

eslavich (commented):
Just one little thing

.pre-commit-config.yaml
WilliamJamieson: No because those are counted as "binary" files somewhere.  Which means those actions do not get run on them.

eslavich: Ah!  Okay, great, thanks.

eslavich (approved):
🚀

PR #1087, opened 2022-02-16
Update codecov so that it is functional again

WilliamJamieson:
Codecov seems to be broken, this PR attempts to fix this.


.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             python-version: "3.10"
             toxenv: py310
 
-          - name: Code Coverage with Python 3.9
+          - name: Python 3.9 with Code Coverage
```
eslavich, line 26: Fair :)

eslavich (approved):
LGTM, thanks for fixing that!

PR #1092, opened 2022-02-18
Remove AsdfFile.extension_manager, AsdfFile.extension_list, and AsdfFile.version setter

eslavich:
Now that user extensions are gone, we can move ExtensionManager to AsdfConfig instead of keeping a reference on each AsdfFile.  This simplifies the code vs passing around those instances, and saves users from having to instantiate an AsdfFile to get ahold of ExtensionManager.

I'm also dropping the AsdfFile.version setter in this PR, which mainly existed to reset the AsdfFile's extension variables.  I don't think it's safe to change the version of an open AsdfFile anyway (instead users should call write_to or maybe update).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,11 @@
 
 - Remove extensions argument from AsdfFile class and open functions. [#1062]
 
+- Remove ExtensionManager and ExtensionList from AsdfFile instance and store
+  in AsdfConfig instead. [#1092]
+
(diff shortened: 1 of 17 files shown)


WilliamJamieson:
@eslavich  It looks like there are some new warnings inside the tests: https://github.com/asdf-format/asdf/runs/5278863528?check_suite_focus=true. Are you planning on fixing them?

WilliamJamieson (approved):
LGTM!

PR #1094, opened 2022-02-21
Fix for circular dependencies between asdf packages

### LINKED ISSUE #1093: Cyclic build dependency asdf-transform-schemas and asdf-standard
While trying to package asdf 2.10 for Debian, I found that it has a new dependency on [asdf-transform-schemas](https://github.com/asdf-format/asdf-transform-schemas) and [asdf-standard](https://github.com/asdf-format/asdf-standard) https://github.com/asdf-format/asdf/blob/3483bfe273205eac39ca20097b2af7a27a3283be/setup.cfg#L28-L37
asdf-transform-schemas however already [depends on asdf](https://github.com/asdf-format/asdf-transform-schemas/blob/f2eeb2dc767c0a3df964a99397baf57060a3f629/setup.cfg#L18-L20), the same for [asdf-standards](https://github.com/asdf-format/asdf-standard/blob/c7360d1a5ef66af68f13f84a2b16c1de0b6c55b9/setup.cfg#L22-L24).

This is unfortunate, since is doesn't allow bootstrapping from sources only, and therefore complicates builds. Would it be possible to break these dependency loops up?

WilliamJamieson:
Fixes #1093. It does this by moving `DirectoryResourceMapping` from asdf to asdf-standard.

Depends on PR: asdf-format/asdf-standard#300 (along with releasing asdf-standard 1.0.1) and asdf-format/asdf-transform-schemas#38 (along with releasing asdf-transform-schemas 0.2.1).


eslavich (approved):
LGTM

PR #1095, opened 2022-02-22
Update to allow for better handling of multiple schema uris per tag uri

WilliamJamieson:
Fixes a small bug which prevents validation of a tag which has more than one schema associated with it, but which does not have a loaded converter extension.


CagtayFabry:
Hi @WilliamJamieson,
do you have an example at hand where one tag is associated with more than one schema?
I would like to look more into how and where to apply that concept.

WilliamJamieson:
@CagtayFabry there is a contrived example [here](https://github.com/asdf-format/asdf/blob/3483bfe273205eac39ca20097b2af7a27a3283be/asdf/tests/test_integration.py#L70-L135) in the tests.

The asdf-format/asdf-transform-schemas#40 PR was originally intended to also demonstrate this feature; however, we found some bugs in asdf (one of which is fixed by this PR) which prevented us from implementing examples in the schema files. These bugs were entirely within the testing infrastructure, and not with the actual implementation.

The multiple schema uris per tag is intended to help us tame the large number of new schemas in the asdf-transform-schemas package that results from changing the `transform-1.x.0` schema and cut down on the duplication of otherwise identical schemas. In asdf-format/asdf-transform-schemas#40, we are introducing schemas for `cos`, `tan`, `arcsine`, `arccos`, and `arctan` based models. All of these models use the same parameter sets as the `sin` transform just swapping the trigonometric function involved (note that the inverse-trigonometric functions implement the parameters so that they are exactly the inverse of their trigonometric counterpart). This means that essentially the `sine1d`, `cosine1d`, `tangent1d`, `arcsine1d`, `arccosine1d`, and `arctangent1d` schemas are identical save for their tags. The multiple schemas per uri would let one instead define one `trigonometric1d` schema as follows:
```
%YAML 1.1
---
id: "http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0"
title: >
  One dimensional trigonometric model.

description: >
  One dimensional trigonometric.

type: object
properties:
  amplitude:
    anyOf:
      - $ref: "../unit/quantity-1.1.0"
      - type: number
    description: Oscillation amplitude.
  frequency:
    anyOf:
      - $ref: "../unit/quantity-1.1.0"
      - type: number
    description: Oscillation frequency.
  phase:
    anyOf:
      - $ref: "../unit/quantity-1.1.0"
      - type: number
    description: Oscillation phase.

required: ['amplitude', 'frequency', 'phase']
...
```
Then one can define the tags for each of the trigonometric transforms in the manifest using:
```
- tag_uri: tag:stsci.edu:asdf/transform/sine1d-1.0.0
  schema_uri:
    - http://stsci.edu/schemas/asdf/transform/transform-1.2.0
    - http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0
  title: One dimensional sine model.
  description: |-
    One dimensional sine.
- tag_uri: tag:stsci.edu:asdf/transform/cosine1d-1.0.0
  schema_uri:
    - http://stsci.edu/schemas/asdf/transform/transform-1.2.0
    - http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0
  title: One dimensional cosine model.
  description: |-
    One dimensional cosine.
- tag_uri: tag:stsci.edu:asdf/transform/tangent1d-1.0.0
  schema_uri:
    - http://stsci.edu/schemas/asdf/transform/transform-1.2.0
    - http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0
  title: One dimensional tangent model.
  description: |-
    One dimensional tangent.
- tag_uri: tag:stsci.edu:asdf/transform/arcsine1d-1.0.0
  schema_uri:
    - http://stsci.edu/schemas/asdf/transform/transform-1.2.0
    - http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0
  title: One dimensional arcsine model.
  description: |-
    One dimensional arcsine.
- tag_uri: tag:stsci.edu:asdf/transform/arccosine1d-1.0.0
  schema_uri:
    - http://stsci.edu/schemas/asdf/transform/transform-1.2.0
    - http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0
  title: One dimensional arccosine model.
  description: |-
    One dimensional arccosine.
- tag_uri: tag:stsci.edu:asdf/transform/arctangent1d-1.0.0
   schema_uri:
    - http://stsci.edu/schemas/asdf/transform/transform-1.2.0
    - http://stsci.edu/schemas/asdf/transform/trigonometric1d-1.0.0
  title: One dimensional arctangent model.
  description: |-
    One dimensional arctangent.
```
Note that by doing this we do not need to add schemas for any of the `sine1d`, `cosine1d`, etc.

Moreover, note that `trigonometric1d-1.0.0` does not have an `allOf` between `transform-1.2.0` and its additional properties, this is instead handled by the manifest because it implies an `allOf` between all of the schema uris listed in the entry. This means that if a new `transform` schema were to be needed, we could simply update the manifest instead of essentially duplicating `trigonometric1d` with only the `transform` version updated. We plan to extend this usage to the rest of the transform schemas.

CagtayFabry:
Thank you for that thorough write-up @WilliamJamieson, makes it much clearer!

The implied `allOf` is probably not what I have in most use cases but certainly good to know this pattern for the future.

WilliamJamieson:
> The implied `allOf` is probably not what I have in most use cases but certainly good to know this pattern for the future.

Please feel free to suggest additional methods of combining the schemas into the tag. If there is a good use-case for other methods of combining them please let me know. It should be relatively simple to add a control property to each manifest entry to signal other modes of combining them (getting it to work on the backend is a different matter).

CagtayFabry:
> > The implied `allOf` is probably not what I have in most use cases but certainly good to know this pattern for the future.
> 
> Please feel free to suggest additional methods of combining the schemas into the tag. If there is a good use-case for other methods of combining them please let me know. It should be relatively simple to add a control property to each manifest entry to signal other modes of combining them (getting it to work on the backend is a different matter).

I'll take some time and look into it and let you know, thank you 👍

PR #1097, opened 2022-02-25
Bugfix for roman_datamodels

WilliamJamieson:
Quick bugfix for `roman_datamodels`, it expects `fits_embed` to be available from `import asdf`. This used to be available via `commands` which was removed recently. This PR explicitly imports `fits_embed` in `__init__.py`.


WilliamJamieson:
I think spacetelescope/roman_datamodels#69 is a better fix than this #1098 has been updated to include the downstream testing included in this PR. I believe it should be merged, closing this PR.

PR #1101, opened 2022-03-03
Update setuptools_scm configuration

WilliamJamieson:
Update usage of `setuptools_scm` to use the `pyproject.toml` file rather than the now depreciated `setup.py` method instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,4 +56,4 @@ pip-wheel-metadata
 .DS_Store
 
 # setuptools-scm generated version file
-asdf/version.py
+asdf/_version.py
--- a/asdf/__init__.py
+++ b/asdf/__init__.py
@@ -22,6 +22,7 @@ __all__ = [
 from jsonschema import ValidationError
 
 from ._convenience import info
+from ._version import version as __version__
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -56,4 +56,4 @@ pip-wheel-metadata
 .DS_Store
 
 # setuptools-scm generated version file
-asdf/version.py
+asdf/_version.py
```
eslavich, line 59: This is _technically_ a breaking change but I can't find any code that refers to this module outside of the asdf library, so I don't expect any problems.

eslavich (approved):

.gitignore
WilliamJamieson: Its essentially the same file, it just brings it in line with `setuptools_scm` convention. All it provides is `__version__`. Which we might want to update everything in `asdf` which was importing this to use instead?

eslavich: I just mean that previously it was possible for users to do this:

```
from asdf.version import version
```

and now it's not, but I don't think that's something we need to worry about since the API we intended to expose was `asdf.__version__`

PR #1107, opened 2022-03-15
Pin pytest to < 7.1 for Windows

eslavich:
Until https://github.com/pytest-dev/pytest/issues/9765 is resolved, our Windows CI job will fail with pytest 7.1.


WilliamJamieson (approved):
LGTM

Issue #1114, opened 2022-03-17
Suggested new text to be added to the RTD documentation


KenMighell:
The file  
```https://asdf.readthedocs.io/en/stable/_sources/asdf/extensions.rst.txt ```
has the following hidden code after the presentation of the first YAML code block of the "An Example" subsubsection:
```
.. runcode:: hidden

    import os
    import asdf
    # This is a hack in order to get the example below to work properly
    __file__ = os.path.join(asdf.__path__[0], 'tests', 'data', 'fraction-1.0.0.yaml')
```
This hidden code has important information for the reader. Specifically, the YAML code block needs to be saved in the local working directory with the filename of "fraction-1.0.0.yaml" in order for the example to perform properly in a local working directory of their own choosing.

I suggest that this new note be added to to the file  
```https://asdf.readthedocs.io/en/stable/_sources/asdf/extensions.rst.txt ```  
just after the first YAML code block is presented (i.e., before the text "Then, the Python implementation of the tag class and extension class."):
```
.. note::

    The above YAML code needs to be saved in the local working directory with the
    filename of "fraction-1.0.0.yaml" in order for this example to perform properly.

```

The addition of this note should make life easier for the eager reader trying to replicate the results presented in the RTD documentation in a local working directory of their own choosing.

WilliamJamieson:
Please feel free to contribute these changes by opening a pull request against asdf.

KenMighell:
Should I make the changes to version 2.7.5 or to master?  

I refer above to the file ```extensions.rst``` which is version 2.7.5.  The closest to that file that I can find in version 2.10.1 is ```legacy.rst`` which is for the section "Deprecated extension API".

The online RTD documention at the GitHub repository at ```https://github.com/asdf-format/asdf``` is linked to ```http://asdf.readthedocs.io/``` which points to ```https://asdf.readthedocs.io/en/stable/``` .

The latest RTD documentation (```https://asdf.readthedocs.io/en/latest/```) is version 2.10.1 -- unfortunately that documentation is not as easily visible as the 2.7.5 (stable) version.

WilliamJamieson:
@KenMighell I am closing this issue, please feel free to re-open it if you wish to discuss this further.

PR #1117, opened 2022-03-19
Added a note to legacy.rst (#1114) and updated versioning.py

KenMighell:
Added a note to legacy.rst (#1114) and updated versioning.py

### DIFF
--- a/asdf/versioning.py
+++ b/asdf/versioning.py
@@ -182,3 +182,4 @@ RESTRICTED_KEYS_MIN_VERSION = AsdfVersion("1.6.0")
 # This library never removed defaults for ASDF Standard versions
 # later than 1.5.0, so filling them isn't necessary.
 FILL_DEFAULTS_MAX_VERSION = AsdfVersion("1.5.0")
+
--- a/asdf/versioning.py
+++ b/asdf/versioning.py
@@ -156,7 +156,17 @@ supported_versions = [
     AsdfVersion("1.6.0"),
 ]
 
-default_version = AsdfVersion("1.5.0")
+""" quote from file:///config.rst:
+default_version
+---------------
+
+The default ASDF Standard version used for new files.  This can be overridden
+on an individual file basis (using the version argument to ``AsdfFile.__init__``)
+or set here to change the default for all new files created in the current session.
+
+Defaults to the latest supported ASDF Standard version.
+"""
+default_version = AsdfVersion("1.6.0")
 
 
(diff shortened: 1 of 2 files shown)


asdf/versioning.py
```diff
@@ -156,7 +156,17 @@ def __hash__(self):
...
+""" quote from file:///config.rst:
+default_version
+---------------
+
+The default ASDF Standard version used for new files.  This can be overridden
+on an individual file basis (using the version argument to ``AsdfFile.__init__``)
+or set here to change the default for all new files created in the current session.
+
+Defaults to the latest supported ASDF Standard version.
+"""
```
WilliamJamieson, line 168: First don't copy/paste quotes from the documentation to the source code. We want to avoid the issues that occur with duplication whenever possible.

Second, in `docs/asdf/config.rst` the following update needs to occur:
```
Defaults to the latest supported ASDF Standard version.
```
should be changed to
```
Defaults to the latest stable ASDF Standard version.
```
This is because ASDF Standard 1.6 is still in flux and has not been fully adopted yet.

asdf/versioning.py
```diff
@@ -156,7 +156,17 @@ def __hash__(self):
...
+default_version = AsdfVersion("1.6.0")
```
WilliamJamieson, line 169: This change should be reverted in light of my previous comment.

docs/asdf/extending/legacy.rst
```diff
@@ -66,6 +66,12 @@ First, the YAML Schema, defining the type as a pair of integers:
     maxItems: 2
     ...
 
+.. note::
+
+   If you are writing code in a local working directory and you wish to use
+   this YAML code, then save it in the local working directory with the 
+   filename of "fraction-1.0.0.yaml".
+
```
WilliamJamieson, line 74: Can you confirm that this actually makes the example work? Given the mechanism that the hidden code uses I am not sure that it will.

Also, `asdf.CustomType` has been deprecated in favor of `asdf.extension.Converter` hence why this file is `legacy.rst`. A note like this might be more appropriate in `docs/asdf/extending/converters.rst`

WilliamJamieson (changes requested):
Please also make sure that the unit tests pass and that your changes fit our code style requirements.

eslavich:
@KenMighell this PR hasn't seen activity in nearly 6 months, so I'm going to close it.  Please feel free to reopen it if you'd still like to get this change in.

PR #1123, opened 2022-03-24
Potential fix for windows CI

WilliamJamieson:


WilliamJamieson:
Looks like changing the windows VM did not solve the pytest issue.

pllim:
Sorry for the false lead. I got confused with a different problem involving C-extension. 😬

PR #1129, opened 2022-04-05
Refactor seekable in `GenericFile`

WilliamJamieson:
The `GenericFile` class (and its subclasses) in asdf is intended to be a wrapper/handler for `io.IOBase` derived python objects, which are what python opens files into. These objects are either "seekable" or not, where in the base python object the `.seekable()` method returns `True` if it is seekable and `False` otherwise. Currently, `GenericFile` and its subclasses default to having `.seekable()` always be `False` even when wrapping python `io` objects which are seekable. This PR changes this behavior so that `GenericFile.seekable()` defaults to the value of `.seekable()` value provided by the underlying python object.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,9 @@
 
 - Remove setter for AsdfFile.version. [#1092]
 
+- Refactor "seekable" default for GenericFile from False, to the value provided by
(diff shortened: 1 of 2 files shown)


eslavich (approved):
Looks like a good change!

Issue #1137, opened 2022-05-18
Display schema titles as comments in info() functionality


perrygreenfield:
Extract the title (if it exists) for any attributes from the corresponding schema for that attribute and display that as part of the info() output.

WilliamJamieson:
Completed by #1138.

Issue #1122, opened 2022-03-24
Unpin pytest 7.1 for Windows


WilliamJamieson:
Currently Windows CI is broken for pytest 7.1, see pytest-dev/pytest#9765. Currently we have pinned pytest below 7.1 for Windows.

PR #1138, opened 2022-05-19
added extraction of title from schema to show as a comment in info()

perrygreenfield:
Note that due to how the extension manager works, for the test file this required adding an option to info to force reloading of the extension manager so that the test extensions would be seen. There may be a better way to handle this. It should handle attributes that use patternProperty, oneOf, allOf, or anyOf constructs. This addresses issue #1137.


asdf/tests/test_search.py
```diff
@@ -189,3 +189,17 @@ def test_recursive_tree():
     result.replace("zap")
     assert af["foo"]["bar"] == "zap"
     assert af["foo"]["nested"]["bar"] == "zap"
+
+
+def test_search():
+    tree = dict(foo=42, bar="hello", baz=np.arange(20))
+    af = asdf.AsdfFile(tree)
```
WilliamJamieson, line 196:
```suggestion
    af = AsdfFile(tree)
```

I think this will fix the broken tests

asdf/tests/test_info.py
```diff
@@ -0,0 +1,359 @@
+import getpass
+import io
```
WilliamJamieson, line 2:
```suggestion
```

Partial fix to flake8

asdf/tests/test_info.py
```diff
@@ -0,0 +1,359 @@
...
+
+import numpy as np
+import pytest
+from jsonschema.exceptions import ValidationError
+
+import asdf
+from asdf import config_context, extension, get_config, resolver, schema, treeutil, versioning
+from asdf.exceptions import AsdfDeprecationWarning, AsdfWarning
+from asdf.extension import ExtensionManager, ExtensionProxy, ManifestExtension
+from asdf.resource import DirectoryResourceMapping
```
WilliamJamieson, line 18:
```suggestion
import os
import pathlib
import re
import shutil
import tempfile

import numpy as np

import asdf
from asdf.extension import ExtensionManager, ExtensionProxy, ManifestExtension
from asdf.resource import DirectoryResourceMapping
```

This should fix the other flake8 issues

perrygreenfield: done.

asdf/tests/test_search.py
perrygreenfield: Already done.

WilliamJamieson:
I think the windows failure can be solved by adding `!=7.1.2` to this line https://github.com/asdf-format/asdf/blob/20741c8043449a1274848cb6d229cf8f14515798/setup.cfg#L56
This is related to #1122.

setup.cfg
```diff
@@ -53,7 +53,7 @@ docs =
 tests =
     # Necessary for Windows until https://github.com/pytest-dev/pytest/issues/9765
     # is resolved:
-    pytest!=7.1.0, !=7.1.1;sys_platform=="win32"
+    pytest!=7.1.0, !=7.1.1 !=7.1.2;sys_platform=="win32"
```
WilliamJamieson, line 56:
```suggestion
    pytest!=7.1.0, !=7.1.1, !=7.1.2;sys_platform=="win32"
```

WilliamJamieson (approved):
LGTM

setup.cfg
```diff
@@ -53,7 +53,7 @@ docs =
 tests =
     # Necessary for Windows until https://github.com/pytest-dev/pytest/issues/9765
     # is resolved:
-    pytest!=7.1.0, !=7.1.1;sys_platform=="win32"
+    pytest!=7.1.0, !=7.1.1, !=7.1.2;sys_platform=="win32"
```
pllim, line 56: I am trying to track a failure in Jdaviz and I am suspecting it has something to do with `asdf`. Why did you pin a `!=7.1.2` here?

pllim: Oh hmm, this only is for Windows... my failure is on Linux.

pllim: Does this error ring any bell to you?

```
2022-05-25T20:01:04.0755535Z INTERNALERROR> Traceback (most recent call last):
2022-05-25T20:01:04.0756248Z INTERNALERROR>   File "<frozen importlib._bootstrap>", line 939, in _find_spec
2022-05-25T20:01:04.0756783Z INTERNALERROR> AttributeError: '_SixMetaPathImporter' object has no attribute 'find_spec'
```

WilliamJamieson: Is this for ASDF?

pllim: Sorry for the noise. It's not ASDF. Something happened in the Python 3.10 that shipped with Actions runner.

PR #1139, opened 2022-05-20
Add label to main asdf docs

WilliamJamieson:
Add label to main index of asdf docs so that intersphinx referencing can be done.


WilliamJamieson:
The windows CI failure is unrelated

PR #1140, opened 2022-05-24
Change macOS to run on arm64

WilliamJamieson:


WilliamJamieson:
@pllim is correct, this is not possible at this time.

PR #1143, opened 2022-05-27
Support for examples speciflying asdf-standard version

WilliamJamieson:
Adds ability to set asdf-standard for an example item in a schema example test, this is part of the fix for astropy/asdf-astropy#82.

The bug is actually that the pytest plugin always assumes the example is written under the default asdf-standard. This is incorrect in the reported bug because `time-1.0.0` only appears in asdf-standard 1.0.0, so the asdf-astropy extension never gets loaded, causing a fall back to the astropy extension.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,9 @@
-2.11.2
-------
+2.11.2 (unreleased)
+-------------------
 
 - Added ability to display title as a comment in using the
   ``info()`` functionality. [#1138]
+- Add ability to set asdf-standard version for schema example items. [#1143]
 
 2.11.1 (2022-04-15)
 -------------------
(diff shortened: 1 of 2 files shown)


WilliamJamieson:
The asdf-standard tests are expected to fail until this PR is merged and asdf-format/asdf-standard#314 is merged.

nden (approved):
I *think* I understand what is being done here.

nden:
The tests are failing?

perrygreenfield (approved):
LIke Nadia (or worse), I think it make sense, but I can't say I understand it completely.

WilliamJamieson:
> The tests are failing?

The CI job that is failing is for ASDF-standard, that will be fixed when I merge asdf-format/asdf-standard#314.

PR #1149, opened 2022-06-14
move build configuration into `pyproject.toml`

zacharyburnett:
`setuptools` support for `pyproject.toml` is currently "experimental" (as in, the `[tool.setuptools]` table is subject to change); however, the main `[project]` table is now set in [PEP621](https://peps.python.org/pep-0621) and is common metadata storage for many build systems, so for future compatibility the build configuration should move away from `setup.cfg`


pyproject.toml
```diff
@@ -1,10 +1,186 @@
...
+    'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
+    'Development Status :: 5 - Production/Stable',
+]
+dependencies = [
+    'asdf >=2.7.1',
+    'crds >=7.4.1.3',
+    'astropy >=5.0.4',
+    'stdatamodels >=0.2.4',
+]
```
WilliamJamieson, line 22: What is this for? I don't think we need these

pyproject.toml
```diff
@@ -1,10 +1,186 @@
...
+name = 'asdf'
+description = 'Python implementation of the ASDF Standard'
+readme = 'README.rst'
+requires-python = '>=3.7'
+license = { file = 'LICENSE' }
+authors = [{ name = 'The ASDF Developers', email = 'help@stsci.edu' }]
+classifiers = [
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.7',
```
WilliamJamieson, line 11: We also don't technically support python 3.7

zacharyburnett: should the following lines be changed in `master` then?
https://github.com/asdf-format/asdf/blob/80c3a84a00fd96708ce485f2d4c05088bf7f6116/setup.cfg#L18
https://github.com/asdf-format/asdf/blob/80c3a84a00fd96708ce485f2d4c05088bf7f6116/setup.cfg#L26

pyproject.toml
```diff
@@ -1,10 +1,182 @@
...
+    'jmespath >=0.6.2',
+    'jsonschema >=4.0.1',
+    'numpy >=1.10',
+    'packaging >=16.0',
+    'pyyaml >=3.10',
+    'semantic_version >=2.8',
+    # Multiple downstream packages accidentally depend on asdf to
+    # install six.  As a friendly gesture we list six as a dependency
+    # for older Python versions.
+    'six; python_version <"3.8"',
```
WilliamJamieson, line 30:
```suggestion
```
3.7 support has been dropped already.

WilliamJamieson (changes requested):
There will likely need to be a page added to the docs discussing configuring asdf in both the `setup.cfg` and `pyproject.toml`.

zacharyburnett:
built documentation for this PR is here: https://asdf--1149.org.readthedocs.build/en/1149/asdf/extending/extensions.html#installing-an-extension

WilliamJamieson:
Why is there a CI failure on the legacy?

tox.ini
```diff
@@ -15,13 +16,13 @@ deps=
...
-    legacy: semantic_version==2.8
-    legacy: pyyaml==3.13
     legacy: jsonschema==4.0.1
-    legacy: numpy~=1.14.6
-    legacy: pytest~=4.6.11
-    legacy: astropy~=3.0.0
+    legacy: numpy~=1.18
+    legacy: pytest~=6.0.0
+    legacy: pyyaml==3.13
+    legacy: semantic_version==2.8
```
WilliamJamieson, line 25: Why have these been reordered? It makes following the diffs really hard

zacharyburnett:
> Why is there a CI failure on the legacy?

it looks like the requested version of `gwcs` requires `six`:
https://github.com/spacetelescope/gwcs/blob/dc46fda0b007949945d8ac40f208caffd42ac7b0/gwcs/tags/wcs.py#L5

WilliamJamieson:
> > Why is there a CI failure on the legacy?
> 
> it looks like the requested version of `gwcs` requires `six`: https://github.com/spacetelescope/gwcs/blob/dc46fda0b007949945d8ac40f208caffd42ac7b0/gwcs/tags/wcs.py#L5

Not that again. GWCS never added `six` as an explicit dependency and somehow got that from `asdf`.

zacharyburnett:
> Not that again. GWCS never added `six` as an explicit dependency and somehow got that from `asdf`.

ah ok, so should the minimum supported version of GWCS be updated?

tox.ini
zacharyburnett: I ordered these alphabetically, as that's the standard I've seen in other projects. I can revert this change though, you're right it makes it difficult to track

tox.ini
```diff
@@ -16,12 +17,14 @@ deps=
...
+    legacy: six~=1.4.0
     legacy: semantic_version==2.8
     legacy: pyyaml==3.13
     legacy: jsonschema==4.0.1
-    legacy: numpy~=1.14.6
-    legacy: pytest~=4.6.11
-    legacy: astropy~=3.0.0
+    legacy: numpy~=1.18
+    legacy: pytest~=6.0.0
+    legacy: astropy~=5.0.4
```
zacharyburnett, line 27: `pytest` was updated so that it can read `pyproject.toml`, `astropy` was updated to reflect the current minimum supported version, and `numpy` was updated to match `astropy`'s new version

zacharyburnett:
it looks like specifying the correct `six` version fixed the issue https://github.com/asdf-format/asdf/runs/6925126993?check_suite_focus=true

WilliamJamieson (approved):

PR #1150, opened 2022-06-14
cache tox environments and make test workflows dependent upon linting 

zacharyburnett:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
+          fetch-depth: 0
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
@@ -127,6 +128,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
+          fetch-depth: 0
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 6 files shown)


WilliamJamieson:
Some of the issues in this maybe resolved by rebasing this PR on master.

zacharyburnett:
the tests appear to need `six`
```
/home/runner/work/asdf/asdf/asdf/entry_points.py:46: AsdfWarning: asdf_extensions plugin from package gwcs==0.17.1 failed to load:

  ModuleNotFoundError: No module named 'six'
```
https://github.com/asdf-format/asdf/runs/7027693077?check_suite_focus=true#step:7:1145

WilliamJamieson:
> the tests appear to need `six`
> 
> ```
> /home/runner/work/asdf/asdf/asdf/entry_points.py:46: AsdfWarning: asdf_extensions plugin from package gwcs==0.17.1 failed to load:
> 
>   ModuleNotFoundError: No module named 'six'
> ```
> 
> https://github.com/asdf-format/asdf/runs/7027693077?check_suite_focus=true#step:7:1145

Maybe a similar change to the `tox` environment to add `six` should be used as was done in #1149.

Though why aren't we pulling the latest version of gwcs in this case? Another concerning thing is `asdf-astropy` is missing from the `pip freeze` which is why there are so many warnings (I see it in the current CI jobs' `pip freeze`). Is it possible that this is somehow interfering with the correct resolution of the dependencies?

zacharyburnett:
it looks like the dependency resolution is having issues when installing `asdf-standard` from source
```
INFO: pip is looking at multiple versions of pytest-sugar to determine which version is compatible with other requirements. This could take a while.
```
https://github.com/asdf-format/asdf/runs/7133915672?check_suite_focus=true#step:7:171

PR #1151, opened 2022-06-15
fix usages of backticks for inline code in docs

zacharyburnett:
changed usage of ` `` ` (italic in RST) to ` ```` ` (inline code in RST)
```sed
s/ `([\w.-~]+)`([^_])/ ``\1``\2/g
```


zacharyburnett:
never mind, this breaks the linkages in the docs; closing

PR #1154, opened 2022-06-20
unpin dependency versions in `s390x` build

zacharyburnett:
fixes issue with `s390x` test failure: https://github.com/asdf-format/asdf/actions/runs/2526970260


WilliamJamieson (approved):
LGTM, everything is passing.

PR #1156, opened 2022-06-20
Fix for failing latex build

WilliamJamieson:
RTD is failing due to an issue building the latex version of the docs. This PR fixes that bug.

Note the docs builds done as part of the CI do not check if LaTeX is building. This can be checked in by running `make latex` or `make latexpdf` in the `docs` directory of the project.


zacharyburnett (approved):
looks good to me, good catch

once [the RTD build](https://readthedocs.org/projects/asdf/builds/17230725/) passes it should be good to go

PR #1157, opened 2022-06-24
Apply pyupgrade to ASDF

WilliamJamieson:
Applies [`pyupgrade`](https://github.com/asottile/pyupgrade) to ASDF and adds it to the ASDF pre-commit hooks. `pyupgrade` is a tool to automatically update the older python syntax to the newer more preferred syntax.


perrygreenfield (approved):
I'm going to defer to the wisdom of the automation on this.

PR #1162, opened 2022-07-15
Attempt to use the openastronomy tox workflow

Cadair:


.github/workflows/ci.yml
```diff
@@ -15,13 +15,22 @@ on:
...
+    steps:
+      - uses: actions/checkout@v3
+      - id: requirements-hash
+        run: echo "::set-output name=hash::${{ hashFiles('**/pyproject.toml', '**/setup.*', 'tox.ini') }}"
   test:
+    needs: [setup]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
     with:
       cache-path: ~/.cache/pip
-      cache-key: ${{ runner.os }}-pip-${{ hashFiles('**/pyproject.toml', '**/setup.*', 'tox.ini') }}
```
zacharyburnett: Would this work on non-wheel packages that are installed with platform-specific scripts and / or compiled?

Cadair: I am not sure. pip might put wheels it built back in this cache or it might not.

This approach is going to be a lot more reliable than caching the tox directories though. In my opinion you *really * don't want to be doing that, as it kind of destroys half the point of the idempotent nature of the CI.

zacharyburnett:
ignore the labeling, I thought there were also changes to the `downstream` CI

zacharyburnett (approved):
LGTM

.github/workflows/ci.yml
Cadair: I should also note I haven't tried this anywhere else so I am interested in how it goes.

PR #1167, opened 2022-08-12
Add ability to pull information from tagged schemas

WilliamJamieson:
Adds the ability to search and pull information about fields in an asdf file from the underlying schema files. For example, this enables one to store keywords pertaining to a given asdf object's data entries within the schema(s) defining that object. This however requires that the underlying schemas to all be tagged.

### DIFF
--- a/asdf/_node_info.py
+++ b/asdf/_node_info.py
@@ -17,6 +17,18 @@ def collect_schema_info(key, node, identifier="root", preserve_list=True, refres
     return schema_info.collect_info(preserve_list=preserve_list)
 
 
+def _get_extension_manager(refresh_extension_manager):
+    from .asdf import AsdfFile, get_config
+    from .extension import ExtensionManager
+
+    af = AsdfFile()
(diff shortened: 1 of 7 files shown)


perrygreenfield (commented):
Mainly a comment on the docstring and name of the node class.

asdf/_node_info.py
```diff
@@ -0,0 +1,218 @@
...
+                break
+
+            current_nodes = next_nodes
+            current_depth += 1
+
+        return root_info
+
+    def collect_info(self, preserve_list=True):
+        """
+        Collect the information from the NodeSchemaData tree, and return it as nested dict.
```
perrygreenfield, line 198: This reference needs updating.

perrygreenfield (commented):
Only one lingering name left.

asdf/_node_info.py
WilliamJamieson: Done

perrygreenfield (approved):
LGTM

Issue #1168, opened 2022-08-12
New warning from pytest


WilliamJamieson:
A new `pytest` warning has appeared in the ASDF CI. I am uncertain what is causing it:

- https://github.com/asdf-format/asdf/runs/7808618629?check_suite_focus=true shows that we are now getting a warning from `pytest`.
- https://github.com/asdf-format/asdf/runs/7774908375?check_suite_focus=true from two days ago is not failing (this one was triggered by the latest merge to master).

The warning in question is:
```python
pytest.PytestAssertRewriteWarning: Module already imported so cannot be rewritten: __editable___asdf_2_12_1_dev88_g5dda110_finder
```

Examining the logs for these two runs of the CI indicate:

- The runner version,
- The python version,
- The  `tox` version,
- All the dependency versions,
- and the exact version of ASDF (down two the commit) master branch

are all identical. Moreover, I can reproduce this locally when delete my `tox` cache and run `tox`, but cannot reproduce it if I try to manually build an environment to replicate the `tox` one and run the tests directly using `pytest`.

WilliamJamieson:
This appears to have resolved itself, and the warning suppression was removed by PR #1218.

PR #1169, opened 2022-08-12
Temporally Ignore pytest warning

WilliamJamieson:
This ignores the warning discussed in #1168, so that the CI can proceed as normal. The cause of the warning should be found and fixed later.


zacharyburnett (approved):
LGTM, test passes: https://github.com/asdf-format/asdf/runs/7812547937?check_suite_focus=true

PR #1171, opened 2022-08-16
Pin jsonschema below 4.10.0

WilliamJamieson:
`jsonschema 4.10.0` was just released, asdf is now broken: https://github.com/asdf-format/asdf/actions/runs/2868804615.

This PR pins `jsonschema` below `4.10.0`.


zacharyburnett (approved):
LGTM

WilliamJamieson:
Right now this is a test. If I figure out a quick fix it I'll merge that fix instead.

pllim:
Huh... `AssertionError: asdf library version 2.12.0 failed to read an ASDF Standard 1.4.0 file produced by this code`

WilliamJamieson:
This makes no sense: Ran yesterday with no issues.

WilliamJamieson:
> Huh... `AssertionError: asdf library version 2.12.0 failed to read an ASDF Standard 1.4.0 file produced by this code`

This makes no sense: Ran yesterday with no issues: https://github.com/asdf-format/asdf/runs/7845366750?check_suite_focus=true

WilliamJamieson:
Okay, I figured out the issue. The compatibility tests are spawning their own virtual environment and installing a legacy version of asdf into that virtual environment, tests are then run in a subprocess using that environment. Since asdf 2.12 does not pin jsonschema to strictly less than 4.10.0, the virtual environment running the tests has been broken by jsonschema meaning the test will fail.

I think this means that these tests are failing for the same reason(s) asdf is failing for jsonschema 4.10.0.

pllim:
Does this mean you need a hotfix release with this pin? 🤔

WilliamJamieson:
> Does this mean you need a hotfix release with this pin? 🤔

I think it means I need to make new releases for 2.11 and 2.12 with this pin (2.11 is affected by this too).

WilliamJamieson:
The compatibility tests are expected to fail

pllim:
Is it possible to merge this pin and do a release ASAP? This is hitting astropy hard -- https://github.com/astropy/astropy/runs/7871226197?check_suite_focus=true

zacharyburnett:
I imagine this should go into a patch release `2.12.1` when merged, but should it also be added to a `2.11.2` release?

pllim:
astropy has oldest dep of asdf set to 2.10.0 -- is 2.10.x still supported?

https://github.com/astropy/astropy/blob/f0717033e244b69c7591e51f4925fbeba8a45e56/tox.ini#L77

WilliamJamieson:
> astropy has oldest dep of asdf set to 2.10.0 -- is 2.10.x still supported?
> 
> [astropy/astropy@`f071703`/tox.ini#L77](https://github.com/astropy/astropy/blob/f0717033e244b69c7591e51f4925fbeba8a45e56/tox.ini#L77)

asdf 2.10 and below pins jsonschema to less than 4.0. So those versions are safe.

pllim:
OK then 2.11.x onwards is fine by us. Thx!

WilliamJamieson:
Just waiting on the Roman regression tests to pass.

PR #1180, opened 2022-08-24
Clean-up flake8 config

WilliamJamieson:


zacharyburnett (approved):
looks good to me, do we want to rename `setup.cfg` -> `.flake8` here? I don't think the `upload_docs` and `build_sphinx` directives are used anywhere that I can see

Issue #1172, opened 2022-08-16
jsonschema 4.10.0 is breaking asdf


WilliamJamieson:
The release of jsonschema 4.10.0 has broken asdf

- https://github.com/asdf-format/asdf/runs/7860422378?check_suite_focus=true shows that asdf is now broken.
- #1171 demonstrates that this is due to jsonschema 4.10.0.

WilliamJamieson:
There seams to be at least two errors:

1. This Depreciation warning: https://github.com/asdf-format/asdf/runs/7860409541?check_suite_focus=true#step:8:29720
2. This `_id_of` get error: https://github.com/asdf-format/asdf/runs/7860422378?check_suite_focus=true#step:9:28059

WilliamJamieson:
Per this comment: https://github.com/python-jsonschema/jsonschema/issues/981#issuecomment-1217946104, it appears that our subclass of the `jsonschema` validator appears to not actually be supported. They appear to be working on a solution to this issue, which is being tracked python-jsonschema/jsonschema#982.

Thus we need to start looking into alternative ways to handle the extension of the validator which `ASDFValidator` provides.

WilliamJamieson:
This was fixed by PR #1203

PR #1185, opened 2022-08-31
Stop subclassing JSONSchema validator

WilliamJamieson:
Potentential fix for issue #1172


WilliamJamieson:
@Julian this is my initial attempt to stop subclassing `Validator` for use in ASDF. I found your suggestion in https://github.com/python-jsonschema/jsonschema/issues/981#issuecomment-1220391907 a bit hard to implement because it fails to have several methods we use from the `Validator` class, rather than re-implement (copy from jsonschema) all the missing parts, I am simply patching the `Validator` class.

Please feel free to leave comments, alternate suggestions, or a review. I would welcome your thoughts.

WilliamJamieson:
I believe the failing tests are due to a bug in `jsonschema`, see python-jsonschema/jsonschema#944 for details.

Julian:
Hi! Patching an object you don't "own" (i.e. a validator class) is definitely even more fragile than subclassing from it, so this definitely could be improved somehow. I'll have to spend some time with it to see what to recommend though, unless someone else sees it first.

WilliamJamieson:
@Julian I completely agree that Monkeypatching is less than ideal. Essentially, I am just "replicating" the changes the previous subclassing would have accomplished. This "avoids" the warning that has been added as that breaks the required testing in downstream packages that rely on asdf (they turn all warnings into raised errors); moreover, the warning does mention that subclassing will be removed at a later date.

The alternate method to this is to start with what you suggested in https://github.com/python-jsonschema/jsonschema/issues/981#issuecomment-1220391907; however, as far as I can tell this then requires me to then replicate much of the jsonschema `Validator` class's interface through copying the necessary parts from the jsonschema source. This also seems fragile and less than ideal because it requires me to watch for all changes to the jsonschema `Validator` and adjust asdf accordingly.  

As far as I can tell there isn't an interface provided by jsonschema which is rich enough to enable asdf to add its additional logic on top of the jsonschema `Validator`'s `iter_errors` method. As far as I can tell, asdf truly needs to modify the `iter_errors` method in order to handle validation with respect to `yaml` `tags` (which are not part of the `json` specification and so are outside the jsonschema specification), which are used as a core part of asdf's functionality.

WilliamJamieson:
Tagging @eslavich  to weigh in, as they understand how asdf leverages jsonschema better than I do.

eslavich:
I agree that we won't be able to support descending into tag schemas by composing jsonschema's validator or modifying the evolve-able attributes.

Should asdf simply vendorize jsonschema at a version that works for us?  Our use case is bizarre and I don't think it's reasonable to expect jsonschema to officially support the level of customization that we need.  Since we're stuck on draft 4 anyway, we won't miss the new features, and that draft is nearly a decade old so I don't expect that there will be much to maintain.

Julian:
> As far as I can tell there isn't an interface provided by jsonschema which is rich enough to enable asdf to add its additional logic on top of the jsonschema Validator's iter_errors method. As far as I can tell, asdf truly needs to modify the iter_errors method in order to handle validation with respect to yaml tags (which are not part of the json specification and so are outside the jsonschema specification), which are used as a core part of asdf's functionality.

If there's any detail you can share about what's missing that'd definitely be useful to hear, though I still haven't dug in in detail yet, apologies.

WilliamJamieson:
> If there's any detail you can share about what's missing that'd definitely be useful to hear, though I still haven't dug in in detail yet, apologies.

Currently the `iter_errors` that asdf uses is: https://github.com/asdf-format/asdf/blob/ebac36e05e90e724be36789c9b25e2d625fc4563/asdf/schema.py#L266-L311

It provides two functions:
1.  It prevents the validator from getting caught in "reference cycles" as yaml provides [anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) which allow one to reference already stated information.  See https://github.com/asdf-format/asdf/blob/ebac36e05e90e724be36789c9b25e2d625fc4563/asdf/schema.py#L272-L278
2. More importantly for asdf it handles the [yaml tags](https://yaml.org/spec/1.2.2/#24-tags), which asdf uses in a similar manner to `$ref` but instead asdf needs to pull a specific schema and use it to validate only the subtree in question against this new schema. This includes validating that the yaml tag from the data matches the one listed in the schema. See https://github.com/asdf-format/asdf/blob/ebac36e05e90e724be36789c9b25e2d625fc4563/asdf/schema.py#L284-L301

The remaining components of the `iter_errors` we use just set it up so that it operates like the jsonschema `iter_errors` or directly call the jsonschema `iter_errors` itself.

The main point is that asdf has to do some work to handle yaml/asdf specific things outside the scope of the original `iter_errors` method (checking/handling them prior to calling down into the original `iter_errors`). The natural way (to me) we can support this is via subclassing, but I am open to suggestions.

braingram (commented):
Looks good to me.

braingram (approved):

PR #1187, opened 2022-09-08
Add ability to search for specific information from schemas

WilliamJamieson:


nden:
I tested it with Roman data and it does what I think it should do.
```
In [33]: cal.schema_info('title', 'roman.meta.ephemeris.velocity_x')
Out[33]: {'title': SchemaInfo(info='[km/s] X component of Roman velocity', value=10)}

In [34]: cal.schema_info('description', 'roman.meta.guidestar.gs_pattern_error')
Out[34]: {'description': SchemaInfo(info='RMS of guide star position in guide window from pattern matching (error on centroid not explicitly calculated, the FACE information takes all the centroids and calculates the error across the guiding pattern)', value=0.0)}


```

WilliamJamieson:
Should we accept an `AsdfSearchResult` object in `schema_info` for the `path` argument in addition to the dot separated string?

perrygreenfield:
That would be nice to accept `AsdfSearchResult`, but I suspect we should leave that for a subsequent PR.

asdf/_node_info.py
```diff
@@ -5,16 +5,41 @@
...
+def collect_schema_info(key, path, node, identifier="root", preserve_list=True, refresh_extension_manager=False):
     """
-    Collect from the underlying schemas any of the info stored under key.
+    Collect from the underlying schemas any of the info stored under key, relative to the path
+
+    Parameters
+    ----------
+    key : str
+        The key to look up.
+    path : str or None
```
perrygreenfield, line 16: I wonder about how lists are handled in all this with regard to path. Something I've entertained is using square brackets for lists in the path (dots not really needed). So turning a list into a dict means turning an integer key into a string I presume and implicit conversion back to an integer for indexing the list. But I don't see schema_info in master so do I need to review a different PR first?

perrygreenfield: After looking at the corresponding code, it doesn't preclude such an enhancement in the future.

perrygreenfield (approved):
LGTM

PR #1189, opened 2022-09-12
Refactor testing to include Python 3.11 and Update Docs testing

WilliamJamieson:
Adds python 3.11 testing, moves the docs testing fully to RTD, and makes docs testing more strict.


WilliamJamieson:
@Cadair Is there a good way to add testing with `3.11-rc.1` to the open astronomy actions?

Cadair:
It's already supported through the `python-version:` flag.

tox.ini
```diff
@@ -12,6 +14,9 @@ deps=
     asdftransformschemasdev: git+https://github.com/asdf-format/asdf-transform-schemas
     asdfwcsschemasdev: git+https://github.com/asdf-format/asdf-wcs-schemas
     asdfcoordinatesschemasdev: git+https://github.com/asdf-format/asdf-coordinates-schemas
+    devdeps: scipy>=0.0.dev0
+    devdeps: numpy>=0.0.dev0
+    devdeps,mpldev: matplotlib>=0.0.dev0
```
zacharyburnett, line 19: is this for the nightly build?

zacharyburnett (approved):
Looks good to me, we will have to revisit when `3.11` is released

tox.ini
WilliamJamieson: Yes

asdf/asdf.py
```diff
@@ -111,7 +111,7 @@ def __init__(
             open during the lifetime of the tree. Setting to False causes
             all data arrays to be loaded up front, which means that they
             can be accessed even after the underlying file is closed.
-            Note: even if `lazy_load` is `False`, `copy_arrays` is still taken
+            Note: even if ``lazy_load`` is `False`, ``copy_arrays`` is still taken
```
zacharyburnett, line 114:
```suggestion
            Note: even if ``lazy_load`` is ``False``, ``copy_arrays`` is still taken
```

asdf/asdf.py
```diff
@@ -74,7 +74,7 @@ def __init__(
             The URI for this ASDF file.  Used to resolve relative
             references against.  If not provided, will be
             automatically determined from the associated file object,
-            if possible and if created from `AsdfFile.open`.
+            if possible and if created from `asdf.AsdfFile.open`.
```
zacharyburnett, line 77: it's really neat how this does linking in the built docs!

zacharyburnett (approved):
looks good to me

asdf/asdf.py
Cadair: Putting False in single ticks actually links it into the CPython docs.

asdf/asdf.py
WilliamJamieson: This is using [`automodapi`](https://github.com/astropy/sphinx-automodapi).

PR #1191, opened 2022-09-16
Only build HTML for RTD

WilliamJamieson:


zacharyburnett (approved):
Are there any users who might want a PDF version? I guess they could manually run `sphinx` in that case.

WilliamJamieson:
> Are there any users who might want a PDF version? I guess they could manually run `sphinx` in that case.

I think this is a reasonable case. What happens is RTD only builds HTML for PRs so you won't get a failure until it goes to deploy after the PR is merged. Its rather annoying. In this case the epub format was broken for some reason.

PR #1195, opened 2022-09-20
Add some more pre-commit checks

WilliamJamieson:


zacharyburnett (approved):
LGTM, assuming the pre-commit checks don't take long

WilliamJamieson:
> LGTM, assuming the pre-commit checks don't take long

These are all basically grep/regex checks. So they are basically instant. flake8 is so slow that you don't even notice them

PR #1199, opened 2022-09-27
Replace use of `pkg_resources` with `importlib`

### LINKED ISSUE #941: Replace use of pkg_resources with importlib.metadata
See discussions:

https://github.com/astropy/astropy/pull/11091
https://github.com/pypa/pip/issues/7413

WilliamJamieson:
Replaces all uses of `pkg_resources` with `importlib` throughout asdf.

Fixes #941


zacharyburnett (approved):
This looks good; should we also add `importlib_metadata` to the build configuration as a dependency contingent on Python version?

WilliamJamieson:
> This looks good; should we also add `importlib_metadata` to the build configuration as a dependency contingent on Python version?

We already are for other uses: https://github.com/asdf-format/asdf/blob/d54afa283214e74569d509a49e888e6bcffa9371/pyproject.toml#L19

WilliamJamieson:
Note "Required" needs to be updated after this is merged due to the CI job name changes.

asdf/tests/test_util.py
```diff
@@ -99,3 +99,20 @@ def read(self, size=-1):
...
+
+
+def test_minversion():
+    import numpy
+    import yaml
+
+    good_versions = ["1.16", "1.16.1", "1.16.0.dev", "1.16dev"]
+    bad_versions = ["100000", "100000.2rc1"]
+    for version in good_versions:
+        assert util.minversion(numpy, version)
```
braingram, line 111: Will this fail for numpy versions <1.16? The pyproject.toml dependencies have numpy >= 1.10. I'm failing to build any old numpy (<1.16) locally so haven't so far been able to test this.

WilliamJamieson: Oh numpy 1.18 and below isn't even supported by numpy anymore. So we may just want to pin above that anyways.

braingram (approved):

PR #1203, opened 2022-10-03
Unpin jsonschema, fix default validation

braingram:
Add custom default validator to 'applicable_validators' argument to jsonschema.validators.create

Tested locally with jsonschema 4.16

This is work towards fixing #1172

One test is still failing locally (a doctest in docs/asdf/config.rst) but I wanted to trigger the CI and discuss the changes so far.

closes #1177

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@
 - Use forc ndarray flag to correctly test for fortran array contiguity [#1206]
 - Unpin ``jsonschema`` version and fix ``jsonschema`` deprecation warnings. [#1185]
 - Replace ``pkg_resources`` with ``importlib.metadata``. [#1199]
+- Unpin jsonschema, fix default validation for jsonschema 4.10+ [#1203]
 
 2.13.0 (2022-08-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@
 - Use forc ndarray flag to correctly test for fortran array contiguity [#1206]
 - Unpin ``jsonschema`` version and fix ``jsonschema`` deprecation warnings. [#1185]
 - Replace ``pkg_resources`` with ``importlib.metadata``. [#1199]
-- Unpin jsonschema, fix default validation for jsonschema 4.10+ [#1203]
+- Fix default validation for jsonschema 4.10+ [#1203]
 
 2.13.0 (2022-08-19)
 -------------------
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,7 +5,7 @@ git+https://github.com/asdf-format/asdf-transform-schemas
 git+https://github.com/asdf-format/asdf-wcs-schemas
 git+https://github.com/astropy/astropy
 git+https://github.com/spacetelescope/gwcs
-git+https://github.com/yaml/pyyaml.git
+git+https://github.com/python-jsonschema/jsonschema
 
 scipy>=0.0.dev0
 numpy>=0.0.dev0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,7 +6,6 @@ git+https://github.com/asdf-format/asdf-wcs-schemas
 git+https://github.com/astropy/astropy
 git+https://github.com/spacetelescope/gwcs
 git+https://github.com/yaml/pyyaml.git
-git+https://github.com/python-jsonschema/jsonschema
 
 scipy>=0.0.dev0
 numpy>=0.0.dev0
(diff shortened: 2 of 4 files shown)


WilliamJamieson:
Unfortunately, we cannot unpin `jsonschema` with these changes as the `DeprecationWarning` raised by the way we use the `Validator` in other places does not allow us to validate schemas using our pytest plugin (we do not accept any warnings while testing schemas). The changes here need to be combined with something that solves the `DeprecationWarning such as PR #1185.

WilliamJamieson:
I opened PR: braingram/asdf#1 on your branch in your fix for the `config.rst` failure you are seeing. Feel free to merge it to fix your bug.

WilliamJamieson:
Also your history is messed up. As a general rule, we do not accept PRs with merge commits inside them like 48527ec784064ca4c3c0ac7ee98bee0091cc4231 (note that merge commits of a PRs onto your branch are fine). This is so we can have a clean history.

This is typically caused by trying to use the "Update branch" button on github. Instead, you should do a `git rebase` locally onto `master` and then force push the result. (Note that similar issues maybe caused by using the "Sync fork" option on github, though I have not checked).

braingram:
Thanks! I rebased and accepted your PR braingram#1.

Thanks also for pointing out #1185 Am I understanding the CI logs right in thinking the errors are in tests for asdf-standard? I have not run those locally and will do so.

pyproject.toml
```diff
@@ -18,7 +18,8 @@ dependencies = [
     'asdf-transform-schemas >=0.2.2',
     'importlib_resources >=3; python_version <"3.9"',
     'jmespath >=0.6.2',
-    'jsonschema >=4.0.1, <4.10.0',
+    #'jsonschema >=4.0.1, <4.10.0',
+    'jsonschema >=4.0.1',
```
WilliamJamieson, line 22:
```suggestion
    'jsonschema >=4.0.1, <4.10.0',
```

I would limit this PR to fixing the failing tests for `jsonschema 4.10.0+` not unpinning it. As part of that I want it tested on earlier versions in addition to the latest `jsonschema`. As such I request we not change the pin and instead add `jsonschema` to the [`requirements-dev.txt`](https://github.com/asdf-format/asdf/blob/master/requirements-dev.txt) instead. This will test your changes in the later devdeps CI jobs.

WilliamJamieson (changes requested):
I think we should limit the scope away from releasing the `jsonschema` pin and focus on fixing `jsonschema` as a dev dependency instead.

CHANGES.rst
```diff
@@ -3,6 +3,7 @@
 
 - Update citation. [#1184]
 - Add search support to `~asdf.AsdfFile.schema_info`. [#1187]
+- Unpin jsonschema, fix default validation for jsonschema 4.10+ [#1203]
```
WilliamJamieson, line 6:
```suggestion
- Fix default validation for jsonschema 4.10+ [#1203]
```

braingram:
Ran pytest locally with jsonschema==4.0.1. All asdf tests passed.

WilliamJamieson:
It is unfortunate that the devdeps jobs are not allowing an install with the development dependency. It will take more effort to resolve this...

WilliamJamieson:
A simple approach is, we merge #1185 with the tests this fixes marked as `xfail`. Then rebase this PR on the result and then remove the `xfail` marks.

WilliamJamieson:
@braingram can you rebase this now that #1185 is merged and remove the two `xfail` in the schema tests?

requirements-dev.txt
```diff
@@ -6,6 +6,7 @@ git+https://github.com/asdf-format/asdf-wcs-schemas
 git+https://github.com/astropy/astropy
 git+https://github.com/spacetelescope/gwcs
 git+https://github.com/yaml/pyyaml.git
+git+https://github.com/python-jsonschema/jsonschema
```
WilliamJamieson, line 9: We will have to revert this change for now.

Basically, `tox` is installing the packages from `requirements-dev.txt` individually, prior to installing asdf itself. Some of these requirements require ASDF to operate (we are just using them for testing purposes in this case). This means that a published version of asdf is installed first during this install phase, all of which require `jsonschema<4.10`.  However, during this process we are also trying to install the development version of `jsonschema`, which is version `4.16`. Thus pip cannot resolve the versions, which causes the crash. 

Thus we need to publish a version of asdf without the `jsonschema` upper pin, prior to enabling dev testing with it

WilliamJamieson (changes requested):

requirements-dev.txt
braingram: Thanks! I removed jsonschema in [af59c7a](https://github.com/asdf-format/asdf/pull/1203/commits/af59c7a8dd7affaee8852f3fdac4eab2dd8189fe)

WilliamJamieson (approved):

Issue #1205, opened 2022-10-06
Metadata don't distinguish between Fortran- and C-order arrays?


bilderbuchi:
I am confused how the metadata seems to not indicate if an array is stored in column- or row-major order.
Based on #538, and triggered by [this HackerNews comment](https://news.ycombinator.com/item?id=33092662), I came up with this example:
```python
import numpy as np
import asdf

farray = np.array([[11,12,13], [21,22,23]], order='F')
carray = np.array([[11,12,13], [21,22,23]], order='C')
file = asdf.AsdfFile({"farray": farray, "carray": carray})
file.write_to("strides.asdf")

file2 = asdf.open("strides.asdf")

farray2 = file2.tree["farray"]
print("Original farray:\n", farray[:,:])
print("Recovered farray:\n", farray2[:,:])
carray2 = file2.tree["carray"]
print("Original carray:\n", carray[:,:])
print("Recovered carray:\n", carray2[:,:])
```
which, when running with current asdf from conda, successfully recovers the arrays correctly:
```
Original farray:
 [[11 12 13]
 [21 22 23]]
Recovered farray:
 [[11 12 13]
 [21 22 23]]
Original carray:
 [[11 12 13]
 [21 22 23]]
Recovered carray:
 [[11 12 13]
 [21 22 23]]
```

However, the header of the created file shows no indication that those two arrays have different order:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 2.13.0}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension.BuiltinExtension
    software: !core/software-1.0.0 {name: asdf, version: 2.13.0}
carray: !core/ndarray-1.0.0
  source: 0
  datatype: int32
  byteorder: little
  shape: [2, 3]
farray: !core/ndarray-1.0.0
  source: 2
  datatype: int32
  byteorder: little
  shape: [2, 3]
...
```
How can this successfully round-trip arrays if the row/column-major information is not contained in the metadata? Is this using some numpy-specific thing (cf #827) when saving the contents to disk?

[...]

PR #1206, opened 2022-10-06
fix saving fortran ndarray issue #1205

braingram:
per eslavich the contiguous array check should check for both f and c order array so modified flags.contiguous to flags.forc

modified ndarray test_fortran_order test to verify fortran ordered on 'roundtrip' write then read

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@
 - Add search support to `~asdf.AsdfFile.schema_info`. [#1187]
 - Add `asdf.search.AsdfSearchResult` support for `~asdf.AsdfFile.schema_info` and
   `~asdf.search.AsdfSearchResult.schema_info` method. [#1197]
+- Use forc ndarray flag to correctly test for fortran array contiguity [#1206]
 
 2.13.0 (2022-08-19)
 -------------------
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -424,7 +424,7 @@ class NDArrayType(AsdfType):
         # is contiguous.  If not, we need to make a copy to avoid
         # writing a nonsense view.
         base = util.get_array_base(data)
-        if not base.flags.contiguous:
+        if not base.flags.forc:
             data = np.ascontiguousarray(data)
             base = util.get_array_base(data)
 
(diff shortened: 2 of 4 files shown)


WilliamJamieson:
Does this fix the extra source being written to the file and add the `stride` to the file?

braingram:
Both those issues appear to be fixed.

Running the test script from issue #1205 produces the following output:
```bash
Original farray:
 [[11 12 13]
 [21 22 23]]
Recovered farray:
 [[11 12 13]
 [21 22 23]]
Original carray:
 [[11 12 13]
 [21 22 23]]
Recovered carray:
 [[11 12 13]
 [21 22 23]]
```
and the following strides.asdf (blocks removed for readability but including the index to show the 2 blocks)
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
  name: asdf, version: 2.13.1.dev54+g271fe197.d20221006}
history:
  extensions:
    software: !core/software-1.0.0 {name: asdf, version: 2.13.1.dev54+g271fe197.d20221006}
carray: !core/ndarray-1.0.0
  source: 1
  datatype: int64
  byteorder: little
  shape: [2, 3]
farray: !core/ndarray-1.0.0
  source: 0
  datatype: int64
  byteorder: little
  shape: [2, 3]
  strides: [8, 16]
...
#ASDF BLOCK INDEX
%YAML 1.1
---
- 674
- 776
...
```
This also appears to fix the related issue where validating a tree with a f order array results in allocation of a new unused block (each time validate is called).

WilliamJamieson:
We should add regression tests for this.

braingram:
Agreed! Is there a specific place for these and something you would recommend?

I modified [test_fortran_order](https://github.com/braingram/asdf/blob/bug/fortran_ndarray/asdf/tags/core/tests/test_ndarray.py#L803) in commit bf2c14479 to check for fortran ordering after 'round tripping' the data to disk. This wasn't being tested before and after adding the check the test failed until flags.forc was used.

WilliamJamieson:
One should be able to test that `strides` are being written as part of the `ndarray` tests: https://github.com/asdf-format/asdf/blob/master/asdf/tags/core/tests/test_ndarray.py. I'm not sure where testing that only one block is created should go.

WilliamJamieson (approved):

PR #1209, opened 2022-10-12
Update pre-commit with yesqa

WilliamJamieson:
Enable [`yesqa` ](https://github.com/asottile/yesqa) as part of pre-commit.

`yesqa` checks that `# noqa` codes apply and orders the `# noqa` codes.


braingram (approved):
Looks good to me.

PR #1210, opened 2022-10-13
Add asdf-unit-schemas as dependency

WilliamJamieson:
Shortly, we will drop `units` from `asdf-standard` in favor of the `asdf-unit-schemas`. For backwards compatibility reasons, we then need to add `asdf-unit-schemas` as a dependency to `asdf`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
       coverage: codecov
 
   test:
-    needs: [core, asdf-standard]
+    needs: [core, asdf-schemas]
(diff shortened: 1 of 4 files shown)


braingram:
Any clue why the  `Run asdf-standard tests` seems to be hung? Otherwise this looks good to me.

braingram:
Should this be merged then the requirements changed or the opposite?

braingram (approved):

WilliamJamieson:
We cannot add the new CI as a requirement until this PR is merged. So we merge and then update the requirements.

Issue #1204, opened 2022-10-05
Use of `pytest` `tmpdir` fixture should be replaced with `tmp_path`


WilliamJamieson:
See astropy/astropy#13787 for details of why this should be done.

WilliamJamieson:
Fixed by #1217. @braingram please link issues to PRs with [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

PR #1216, opened 2022-10-26
Treat tmpdir as str in _assert_roundtrip_tree

pllim:
With the current implementation, it is impossible to replace tmpdir with tmp_path in astropy test that uses this helper function and then try to pass back tmp_path to this thing as a string.

I don't want to start adding Path support here so this is the simplest thing I can think of.

cc @WilliamJamieson 

ref: https://github.com/astropy/astropy/issues/13787#issuecomment-1292020635

### DIFF
--- a/asdf/tests/helpers.py
+++ b/asdf/tests/helpers.py
@@ -184,7 +184,7 @@ def _assert_roundtrip_tree(
     tree_match_func="assert_equal",
 ):
 
-    fname = str(tmpdir.join("test.asdf"))
+    fname = os.path.join(tmpdir, "test.asdf")
 
     # First, test writing/reading a BytesIO buffer
     buff = io.BytesIO()


WilliamJamieson:
Does this actually work if you pass a `tmp_path` to it?

WilliamJamieson:
Also, since this is technically to fix the "public API" (not really, but clearly something downstream is using it), can you update the changelog?

pllim:
I didn't really use this with anything yet. Just opening a PR for discussion. If you move that astropy test from cosmology to astropy.io.misc.asdf , then all this will be part of your asdf maintenance and maybe then we can close this PR without merge.

WilliamJamieson:
> I didn't really use this with anything yet. Just opening a PR for discussion. If you move that astropy test from cosmology to astropy.io.misc.asdf , then all this will be part of your asdf maintenance and maybe then we can close this PR without merge.

If this makes this helper function work with `tmp_path` then I can at least switch `asdf` internally to `tmp_path` to fix #1204.

pllim:
No, I think to natively support tmp_path, you need more than this change. Alas, I don't have time to dig into that for this package.

I am happy with https://github.com/astropy/astropy/pull/13904 and do not wish to pursue this now. Thanks!

PR #1217, opened 2022-10-26
change use of tmpdir to tmp_path in tests

braingram:


asdf/tests/helpers.py
```diff
@@ -184,7 +184,7 @@ def _assert_roundtrip_tree(
     tree_match_func="assert_equal",
 ):
 
-    fname = str(tmpdir.join("test.asdf"))
+    fname = str(tmp_path / "test.asdf")
```
WilliamJamieson, line 187: Since this will be used for testing other packages (e.g. `astropy`), we need to confirm this does not change the interface. Does this work still if one passes a `tmpdir` fixture?

CHANGES.rst
```diff
@@ -11,6 +11,7 @@
 - Fix default validation for jsonschema 4.10+ [#1203]
 - Add ``asdf-unit-schemas`` as a dependency, for backwards compatibility. [#1210]
 - Remove stray toplevel packages ``docker`` ``docs`` and ``compatibility_tests`` from wheel [#1214]
+- Update tests to use tmp_path instead of tmpdir [#1217]
```
WilliamJamieson, line 14:
```suggestion
```

Please add the `no-changelog-entry-needed` label and squash out this change.

asdf/tests/helpers.py
braingram: Thanks for pointing that out. I made a small change in [b7edb3a](https://github.com/asdf-format/asdf/pull/1217/commits/b7edb3a71a965ed738ddfb356331853db19a9826) (convert tmp_path to string and us os.path.join) to support both.

braingram:
I ran the weldx tests with asdf master and got similar failures as the downstream CI for this PR.
When I applied the fix in the weldx PR [818](https://github.com/BAMWelDX/weldx/pull/818) the weldx tests passed (with the asdf master branch).
The weldx tests also passed with the changes in this PR and the weldx 818 PR.

WilliamJamieson (changes requested):
Please squash out the changelog commit and rebase.

braingram:
This is rebased to master and changelog commit squashed. I see 1 change requested so if I'm missing something please let me know as I don't see any code comments or suggestions.

PR #1218, opened 2022-10-26
Turn warnings into errors

WilliamJamieson:
Turns all warnings into errors and cleans up pytest config

### DIFF
--- a/asdf/tests/test_fits_embed.py
+++ b/asdf/tests/test_fits_embed.py
@@ -291,7 +291,7 @@ def test_open_gzipped():
         assert af.tree["stuff"].shape == (20, 20)
 
 
-@pytest.mark.filterwarnings("ignore::astropy.io.fits.verify.VerifyWarning")
+@pytest.mark.filterwarnings("ignore:unclosed file .*")
 def test_bad_input(tmp_path):
     """Make sure these functions behave properly with bad input"""
     text_file = str(tmp_path / "test.txt")
--- a/asdf/tests/test_generic_io.py
+++ b/asdf/tests/test_generic_io.py
@@ -361,6 +361,12 @@ def test_unicode_open(tmp_path, small_tree):
 
 
 @pytest.mark.filterwarnings("ignore:unclosed file .*")
+def test_open_stdout():
+    """Test ability to open/write stdout as an output stream"""
+    with generic_io.get_file(sys.__stdout__, "w"):
+        pass
+
+
 def test_invalid_obj(tmp_path):
     with pytest.raises(ValueError):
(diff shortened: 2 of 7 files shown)


asdf/tests/test_generic_io.py
```diff
@@ -360,6 +360,7 @@ def test_unicode_open(tmpdir, small_tree):
                 pass
 
 
+@pytest.mark.filterwarnings("ignore:unclosed file .*")
```
braingram, line 363: Is it a bug that the following (from line 387 in this test) leaves an unclosed file?
```
with generic_io.get_file(sys.__stdout__, "w"):
    pass
```

asdf/tests/test_fits_embed.py
```diff
@@ -292,7 +292,7 @@ def test_open_gzipped():
         assert af.tree["stuff"].shape == (20, 20)
 
 
-@pytest.mark.filterwarnings("ignore::astropy.io.fits.verify.VerifyWarning")
+@pytest.mark.filterwarnings("ignore:unclosed file .*")
```
braingram, line 295: Should we open an issue for this as asdf open leaves the file open?

braingram (approved):
Looks good. Should we open issue(s) for the unclosed files?

asdf/tests/test_generic_io.py
WilliamJamieson: No, the issue is this part of the test: https://github.com/WilliamJamieson/asdf/blob/8d872b0703792871cdc440e09c051c35ef7846d6/asdf/tests/test_generic_io.py#L368-L371 leaves the file open

braingram: I'm still a bit puzzled by this one. Any clue why the open to __stdout__ is in the invalid_obj test in the first place? It does appear that get_file has a close argument that if added to line 387 removes the need for the ignore filter. If the intention is to test the unclosed file perhaps that should be pulled out into a separate test.

WilliamJamieson: First, the test is not intended to test the unclosed file. It clearly is intended to to test that an error is raised when opening invalid files. As I noted above, the `get_file` does not even attempt to close files and so therefore it would be expected that this test would leave behind unclosed files.

I do agree that the part of the test you reference should be somewhere else as it does succeed without error. However, that is a red hearing with respect to the changes in this PR. Feel free to open a separate PR moving that part of the test into another test or its own.

zacharyburnett (approved):
I'm not totally familiar with the pytest structure here but it LGTM

braingram (approved):
Look good to me.

PR #1220, opened 2022-10-26
Switch to using ruff

WilliamJamieson:
Switches from using `autoflake` and `Flake8` to using the must faster `ruff` linter. Which combined with `black` will perform all the same functions as `autoflake` and `Flake8`


zacharyburnett (approved):
LGTM! Glad to see that ruff has `--fix` as well

PR #1221, opened 2022-10-27
close files opened during a failed call to asdf.open

### LINKED ISSUE #1006: asdf.open leaves behind an open file if it can't sort out how to open it
The following test illustrates:

```python
# test_open_fail.py

from pathlib import Path

import asdf
import pytest


def test_asdf(tmp_path):
    path = tmp_path / "empty.asdf"
    Path(path).touch()

    with asdf.open(path) as af:
        pass
```

When run with `pytest test_open_fail.py --open-files`, the test fails as expected with

```
E           ValueError: Input object does not appear to be an ASDF file or a FITS with ASDF extension
```

But also produces a teardown error, since it leaves the file open:

```
E           AssertionError: File(s) not closed:
E             /private/var/folders/jg/by5st33j7ps356dgb4kn8w900001n5/T/pytest-of-jdavies/pytest-157/test_asdf0/empty.asdf
```

braingram:
Fixes #1006 and allows removal of most of the ignore:unclosed file warning filters in tests.

~~[Opening __stdout__ with get_file](https://github.com/braingram/asdf/blob/bug/unclosed_file/asdf/tests/test_generic_io.py#L366) leaves an open file but this is intended behavior documented in get_file.~~ (see discussion below)


asdf/asdf.py
```diff
@@ -908,51 +909,57 @@ def _open_impl(
...
         generic_file = generic_io.get_file(fd, mode=mode, uri=uri)
-        file_type = util.get_file_type(generic_file)
-
-        if file_type == util.FileType.FITS:
-            try:
-                # TODO: this feels a bit circular, try to clean up. Also
-                # this introduces another dependency on astropy which may
-                # not be desireable.
-                from . import fits_embed
+        try:
```
WilliamJamieson, line 914: I am not a big fan of putting yet another try/except in this method. I find it difficult to follow the logic of why we are catching errors raised within the function itself.

I suggest refactoring slightly, where we split part of your version of `_open_mpl` into a new method,say `_open_base` (think of a better name if you can).

I would keep `_open_mpl` with the same interface, but use it as a wrapper with your try/except around `_open_base` i.e:
```python
close_on_fail = isinstance(fd, (str, pathlib.Path))
generic_file = generic_io.get_file(fd, mode=mode, uri=uri)

try:
    return cls._open_base(..., generic_file,....)
except Exception as e:
    if close_on_fail:
        generic_file.close()
    raise e
```
Then `_open_base` would essentially be the "original" `_open_mpl`, these parts: https://github.com/asdf-format/asdf/blob/2b4afb65e932a3fcd39056efa221857ee798ae96/asdf/asdf.py#L914-L955  But with interface modified so that it accepts `generic_file` directly rather than calling `get_file` on something.

Doing this will make it much more clear that we are specifically trying to close the file after an error is raised during a file open. This can be documented in the document strings for the two functions.

asdf/tests/test_asdf.py
```diff
@@ -166,7 +166,6 @@ def test_asdf_file_version_requirement():
             af = AsdfFile(version="1.4.0", extensions=[extension_with_requirement])
 
 
-@pytest.mark.filterwarnings("ignore:unclosed file .*")
```
WilliamJamieson, line 169: Since this is fixing issue #1006 and it has a nice reproducer, I think it would also be a good idea to specifically create a regression test based on that reproducer, which documents that it is a regression test for fixing #1006.

Technically, this isn't necessary because some of the tests you are fixing the filters on do basically test for this, but those are side effects of the main purposes of those tests. Hence, we should specifically test for this behavior.

WilliamJamieson (changes requested):
Overall I like these changes. I have some suggestions to improve these changes below.

I am also curious if `AsdfInFits` suffers from the same issue? Namely, if we force an exception during steps of opening directly with it that the file is actually closed? I think this is the case but if it isn't maybe a similar fix will work here.

WilliamJamieson:
> [Opening **stdout** with get_file])([braingram/asdf@`bug`/unclosed_file/asdf/tests/test_generic_io.py#L366](https://github.com/braingram/asdf/blob/bug/unclosed_file/asdf/tests/test_generic_io.py?rgh-link-date=2022-10-27T20%3A55%3A53Z#L366)) leaves an open file but this is intended behavior documented in get_file.

Could we just call `.close()` on the file in the this test to eliminate the warning suppression? I should probably have tried that when writing the test in #1218; however, it just occurred to me now.

asdf/asdf.py
WilliamJamieson: Note that in general, I find tracing the logic of opening the asdf file through each of the open functions involved within `asdf.py` quite tedious. I'd like to see those functions refactored into more, but smaller logical units.

braingram: Thanks for the descriptions. Refactoring the open functions sounds like a great idea.

Currently open (for an asdf file) calls:
- open_asdf
- _open_impl
- _open_asdf
All 3 of these have overlapping arguments with defaults. Cleaning this up might be best handled by pulling some of those arguments into asdf_config (as mentioned in AL-652: Move remaining asdf.open and AsdfFile configuration options to AsdfConfig).

braingram:
> 

Adding close=True allows for removing the warning suppression. I'll commit that fix here.

asdf/asdf.py
WilliamJamieson: This clean up should probably go into a separate PR. It might be a good intermediate project. We should look at the 3.0 branch to see if there are any changes that we can pull back to 2.x

WilliamJamieson: To clarify the part of the refactor I initially proposed here should stay in this PR. I think the discussion here warrants further discussion. Lets move this to issue #1223.

WilliamJamieson (changes requested):
Excellent, I really like your name choice `_open_generic_file`. This is a solid improvement. I only request now that you add the regression test I requested here: https://github.com/asdf-format/asdf/pull/1221/files#r1007468048.

braingram:
I'm going to leave this as a draft until I can sort out the close file behavior for AsdfInFits. It appears to handle some cases of file closure which could result in this PR creating a double closure.

asdf/tests/test_asdf.py
```diff
@@ -363,3 +362,27 @@ def test_reading_extension_metadata():
...
+    text_file = str(tmp_path / "test.txt")
+
+    with open(text_file, "w") as fh:
+        fh.write("I <3 ASDF!!!!!")
+
+    with pytest.raises(ValueError):
+        open_asdf(text_file)
+
+
+def test_unclosed_file(tmp_path):
```
braingram, line 378: @WilliamJamieson Here's the regression test. I also moved test_bad_input here from test_fits_embed as it did not test fits.

WilliamJamieson: Excellent! Looks good.

WilliamJamieson (approved):
Looks good!

Issue #1146, opened 2022-06-08
`asdf` does not appear able to open cloud-hosted files using `fsspec` file handles


barentsen:
## Summary

Using `asdf` v2.12.0, it does not appear possible for `asdf` to work with file handles which have been opened using `fsspec`.  (See reproducible example and traceback appended below.)

## Background - what is fsspec?!

[Fsspec](https://filesystem-spec.readthedocs.io/) is a Python package which provides a `file`-like interface to remote files systems (e.g., web servers, Amazon S3, Google Cloud Storage, etc). For example, given the HTTP or S3 url of a file, `fsspec` enables that file to be accessed in an efficient way using standard random access file operations:

```python
import fsspec
with fsspec.open(url).open() as fh:
    fh.seek(80)
    fh.read(10)
```

Behind the scenes, `fh.read(10)` will tend to trigger a buffered HTTP Range Request (or a protocol equivalent).  As a result, fsspec enables small sections of large remote files to be accessed efficiently without downloading the entire file.

Because fsspec is rapidly becoming a standard way for Python libraries to interact with remote data (e.g., fsspec powers `dask` and `pandas`), I recently opened AstroPy PR https://github.com/astropy/astropy/pull/13238 to demonstrate how AstroPy could potentially leverage fsspec to work with cloud-hosted FITS files in an efficient way.  This led me to evaluate whether `asdf` could work with `fsspec` as well.

## Issue encountered

`asdf` does not appear to work with `fsspec` file handles out of the box at this time.  For example the following snippet raises an `AttributeError`:
```python
import asdf
import fsspec

fh = fsspec.open("example.asdf").open()  
af = asdf.open(fh)
data = af['mydata1'][100]
```

Traceback:
```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [1], in <cell line: 7>()
      5 fh = fsspec.open("example.asdf").open()
      6 af = asdf.open(fh)
----> 7 data = af['mydata1'][100]

File ~/.pyenv/versions/asdf/lib/python3.9/site-packages/asdf/tags/core/ndarray.py:559, in _make_operation.<locals>.__operation__(self, *args)
    558 def __operation__(self, *args):
--> 559     return getattr(self._make_array(), name)(*args)

File ~/.pyenv/versions/asdf/lib/python3.9/site-packages/asdf/tags/core/ndarray.py:263, in NDArrayType._make_array(self)
    260 else:
    261     dtype = self._dtype
[...]

PR #1226, opened 2022-10-31
Use ducktyping in generic_io get_file

braingram:
Instead of using isinstance, look for attributes 'raw' and 'getvalue' to determine what generic_io subclass to use. This adds support for fsspec objects.


WilliamJamieson:
Is this intended to solve #1146? If so can you include a regression test that actually tests opening using a file using `fsspec` (don't worry about ASDF-in-fits support here)?

It is fine to add `fsspec` to the test dependencies for this (make sure its one compatible with [`astropy`](https://github.com/astropy/astropy/blob/56571050cdc30bc5ec7a0d022d690538ae9cfa33/setup.cfg#L97).

braingram:
That issue was part of the motivation. I'll look into which fsspec version to add and write a test.

WilliamJamieson (approved):

PR #1228, opened 2022-11-02
Add fsspec http support

braingram:
Opening an http filesystem file using fsspec returns an object that is incompatible with generic_io. The fsspec object, when treated like a generic_io.RealFile, fails on both the _size and name/uri calculations.

[generic_io.GenericFile._size](https://github.com/asdf-format/asdf/blob/989cd955420ae717dff763c152be3cc21343d244/asdf/generic_io.py#L221) does not appear to be used so this PR removes it from all GenericFile classes.

[fd.name](https://github.com/asdf-format/asdf/blob/989cd955420ae717dff763c152be3cc21343d244/asdf/generic_io.py#L727) is used to calculate the uri (when not otherwise supplied). fsspec http files do not have a name so a different method of defining uri (such as providing one at the time of opening) will need to be used when a uri is necessary. This PR adds a `hasattr` check for name to avoid an exception during creation of the RealFile.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@
 - Remove stray toplevel packages ``docker`` ``docs`` and ``compatibility_tests`` from wheel [#1214]
 - Close files opened during a failed call to asdf.open [#1221]
 - Modify generic_file for fsspec compatibility [#1226]
+- Add fsspec http filesystem support [#1228]
 
 2.13.0 (2022-08-19)
 -------------------
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -218,7 +218,6 @@ class GenericFile(metaclass=util.InheritDocstrings):
         self._fd = fd
         self._mode = mode
         self._close = close
-        self._size = None
         self._uri = uri
 
         self.block_size = get_config().io_block_size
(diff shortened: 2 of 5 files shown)


braingram:
@WilliamJamieson do you know of any uses of generic_io.GenericFile._size? This appears to be calculated for several of the GenericFile subclasses but I'm not seeing a use anywhere in the code.

WilliamJamieson:
> @WilliamJamieson do you know of any uses of generic_io.GenericFile._size? This appears to be calculated for several of the GenericFile subclasses but I'm not seeing a use anywhere in the code.

No

WilliamJamieson:
The jwst failure is unrelated and should be fixed by spacetelescope/jwst#7331.

WilliamJamieson (changes requested):
These changes seem reasonable, do they enable the example in #1146 to work properly now?

If so can you create a test based on that example and add it to `test_asdf.py`?

braingram:
> These changes seem reasonable, do they enable the example in #1146 to work properly now?
> 
> If so can you create a test based on that example and add it to `test_asdf.py`?

Both examples, the local and http filesystems, should now work with tests for both (this PR adds the http filesystem support and tests). Let me know if there's anything else you'd like me to add to this PR otherwise it looks good to me.

WilliamJamieson (approved):

PR #1230, opened 2022-11-14
Feature/whole file memmap

braingram:
Instead of memmaping each array (which consumes one open file descriptor per array) this PR will memmap the entire file (when enabled) and provide the mmap.mmap to each array.

This has the added benefit of greatly increasing open speed for files with a large number of arrays. For example the current jwst ifupost nirspec reference file has 240 arrays. With this PR (on my machine) open takes ~1.5 seconds compared to ~7 without this PR.

~I'm opening this as draft to run the downstream tests.~

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -1210,6 +1210,10 @@ class AsdfFile:
 
             self.blocks.finish_reading_internal_blocks()
 
+            # flush all pending memmap writes
+            if hasattr(fd, "_mmap"):
+                fd._mmap.flush()
+
             self._pre_write(fd, all_array_storage, all_array_compression, compression_kwargs=compression_kwargs)
 
             try:
@@ -1220,6 +1224,11 @@ class AsdfFile:
                     # write out in a serial fashion.
                     self._serial_write(fd, pad_blocks, include_block_index)
                     fd.truncate()
+                    # if we have memmapped data close the memmap so it will
+                    # regenerate
+                    if hasattr(fd, "_mmap"):
+                        fd._mmap.close()
+                        del fd._mmap
                     return
 
                 # Estimate how big the tree will be on disk by writing the
@@ -1243,6 +1252,11 @@ class AsdfFile:
                     # write out in a serial fashion.
                     self._serial_write(fd, pad_blocks, include_block_index)
                     fd.truncate()
+                    # if we have memmapped data close the memmap so it will
+                    # regenerate
+                    if hasattr(fd, "_mmap"):
+                        fd._mmap.close()
+                        del fd._mmap
                     return
 
                 fd.seek(0)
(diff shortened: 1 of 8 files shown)


asdf/asdf.py
```diff
@@ -1243,13 +1250,19 @@ def update(
...
                     return
 
                 fd.seek(0)
                 self._random_write(fd, pad_blocks, include_block_index)
                 fd.flush()
             finally:
                 self._post_write(fd)
+            # close memmaps so they will regenerate
+            if fd.can_memmap():
+                fd.close_memmap()
```
WilliamJamieson, line 1265: Should this be under the `finally`?

braingram: That is a much better way to do it. I'll make that change.

WilliamJamieson: Hopefully that will work. It is nice because it puts it cleanly into the context manager framework being used in that case.

asdf/block.py
```diff
@@ -1224,16 +1223,6 @@ def data(self):
...
-        if self._memmapped and self._data is not None:
-            if NUMPY_LT_1_7:  # pragma: no cover
-                try:
-                    self._data.flush()
-                except ValueError:
-                    pass
-            else:
-                self._data.flush()
-            if self._data._mmap is not None:
-                self._data._mmap.close()
```
WilliamJamieson, line 1236: Nice, we should try to find more of these bits left over from supporting old versions of dependencies and remove them.

asdf/generic_io.py
```diff
@@ -615,6 +616,19 @@ def memmap_array(self, offset, size):
...
+        """
+        raise NotImplementedError()
+
+    def flush_memmap(self):
+        """
+        Flush any pending writes to the memmapped file (if one was mapped with
+        memmap_array)
+        """
+        raise NotImplementedError()
+
```
WilliamJamieson, line 631: Could you add an error message here describing that this should be implemented if needed.

Also, do we need these implemented by all the child classes?

WilliamJamieson (approved):
Overall these changes look great.

asdf/generic_io.py
braingram: I added a rather generic error message to these (and the NotImplementedError in memmap_array). Improvements to the wording are appreciated.

I believe all of the calls to memmap_array, flush_memmap and close_memmap are within conditional blocks that check can_memmap. RealFile is the only subclass that returns True for can_memmap (I'm not sure if others would be possible) and RealFile has all of these implemented. I want to say a mixin might make sense here but I am open to suggestions.

WilliamJamieson: I dislike mixins in general. I was more asking in case all the subclasses where implementing their own version in which case we would want this to be an abc method which is required to be implemented by the subclasses rather than having the error message.

perrygreenfield (approved):
LGTM

PR #1233, opened 2022-11-15
Fix problem of duplicate HDU and ASDF arrays with embedded asdf in FITS files

### LINKED ISSUE #1232: Unnecessary duplicate arrays when using asdf embedded in FITS files
JP-2900 reports a failure of `calwebb_spec3 `when it tries to create source-based products containing slit cutouts from multiple exposures. In one case the asdf extension grew to larger than 4GB, exceeding the size that astropy.io.fits can handle for a binary table entry. This revealed two problems. First, the pipeline was generating cutouts that made no sense (with vertical sizes in pixel in the hundreds) illustrating a problem with the segmentation algorithm and the lack of a sensible filter to reject cases of considerable confusion of sources. This led to trying to stuff many large cutouts into one file. Second, and the one related to a problem with ASDF, is that the same cutout array appeared both in a FITS HDU, and in the ASDF content contained in the ASDF HDU, meaning that data was unnecessarily duplicated in the file.

perrygreenfield:
This fixes #1232 

This is currently an incomplete PR (mainly awaiting results from JWST pipeline tests), and possible discussion about the ASDF behavior with regard to handling FITS blocks.

To summarize the original problem, when collecting slit cutouts (e.g., from slitless or MSA observations) of the same source from multiple exposures into one file, it was found that the cutout data appeared in both FITS extensions and the embedded ASDF, essentially doubling the amount of space the file was using, unnecessarily. This was manifested as an error when one such file required an ADSF extension size more than 4GB, exceeding the permitted astropy.io.fits size for the contents of a FIT binary table cell.

The cause of this problem appears to happen in more than one place in the code. Most explicitly, the current ASDF code presumes that an attribute in the ASDF that refers to a NDArray object, should not be mapped to a FITS block. When extracting the data from exposure-based collections of slit cutouts, these will be represented as NDArrays in the resulting ASDF tree, thus preventing them from being converted to FITS blocks (The FITS extensions are created with such data, but ASDF will not update the tree to refer to the FITS HDUS). But beyond that issue, somewhere in the code (as yet not clear) the two entities no longer refer to the same buffer pointer, and other ASDF code prevent the creation of a FITS block if they do not share the same buffer pointer, even if the two arrays are identical in shape, type, and contents.

To correct the existing problem, the requirement that the ASDF array not be an NDArray was removed, and in two other areas of the code, if identical buffer pointers were not found, an additional test to see if the two arrays had identical content, thus allowing the creation of a FITS block anyway. These changes only broke one existing test, the last test “test_array_view_different_layout”. This test has been disabled since there are significant questions as to how the ASDF library should behave in such a case.

There is a larger question of what the actual semantics of the handling of FITS blocks should be. I do not see such explained clearly anywhere, and it is not clear to me that the current assumptions are what should be desired. For example, in the above referred to failing test, a new tree attribute is created as a view of an existing FITS HDU data array, and the test assumes that attribute should become another FITS block. But should it? Why not allow it to determine if should instead be an internal ASDF array (we do allow such for GWCS arrays)? Or a distinct FITS HDU instead. I believe we should be more careful about what we describe how FITS blocks behave. Currently much is implicit and not clearly spelled out. I tend to be of a mind that we should allow FITS blocks when the array is identical to an existing HDU, if not the same memory contents. For array views we should be explicit that FITS will not support what ASDF  allows in that regard (the views will be different contents one way or the other (either a different HDU or an ASDF internal array)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@
 - Close files opened during a failed call to asdf.open [#1221]
 - Modify generic_file for fsspec compatibility [#1226]
 - Add fsspec http filesystem support [#1228]
+- Fix problem with duplicate arrays with asdf embedded in FITS files [].
 
 2.13.0 (2022-08-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@
 - Close files opened during a failed call to asdf.open [#1221]
 - Modify generic_file for fsspec compatibility [#1226]
 - Add fsspec http filesystem support [#1228]
-- Fix problem with duplicate arrays with asdf embedded in FITS files [].
+- Fix problem with duplicate arrays with asdf embedded in FITS files [#1233]
 
 2.13.0 (2022-08-19)
 -------------------
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -91,15 +91,17 @@ class _EmbeddedBlockManager(block.BlockManager):
 
     def find_or_create_block_for_array(self, arr, ctx):
         from .tags.core import ndarray
-
-        if not isinstance(arr, ndarray.NDArrayType):
-            base = util.get_array_base(arr)
-            for hdu in self._hdulist:
-                if hdu.data is None:
-                    continue
-                if base is util.get_array_base(hdu.data):
-                    return _FitsBlock(hdu)
-
+        base = util.get_array_base(arr)
+        for hdu in self._hdulist:
+            if hdu.data is None:
+                continue
+            if base is util.get_array_base(hdu.data):
+                return _FitsBlock(hdu)
+         
+            if ((hdu.data.dtype == arr.dtype) and 
+                (hdu.data.shape == arr.shape) and
+               np.all(hdu.data == arr)):
+                return _FitsBlock(hdu)
         return super().find_or_create_block_for_array(arr, ctx)
 
 
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -91,16 +91,15 @@ class _EmbeddedBlockManager(block.BlockManager):
 
     def find_or_create_block_for_array(self, arr, ctx):
         from .tags.core import ndarray
+
         base = util.get_array_base(arr)
         for hdu in self._hdulist:
             if hdu.data is None:
                 continue
             if base is util.get_array_base(hdu.data):
                 return _FitsBlock(hdu)
-         
-            if ((hdu.data.dtype == arr.dtype) and 
-                (hdu.data.shape == arr.shape) and
-               np.all(hdu.data == arr)):
+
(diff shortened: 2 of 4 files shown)


WilliamJamieson:
I'm going to convert this to a draft until it is decided that this is "complete".

braingram:
I worry that checking based on value might lead to strange behavior when multiple hdus are equal. This could result in inclusion of all hdus (as those are added before the tree serialization) and the tree would have multiple arrays that point to the same hdu (and some hdus that point to no arrays).

Perhaps we should check for unrefereced hdus and warn or throw an exception. Can you think of any instances where this might cause an issue?

Also FWIW I pushed some of the changes I've been testing to a branch on my fork: https://github.com/braingram/asdf/tree/bug/hdu_array_link_breaking
Here's the diff:
https://github.com/asdf-format/asdf/compare/master...braingram:asdf:bug/hdu_array_link_breaking?expand=1

braingram:
I opened a PR on your fork with a new test that illustrates an issue with using equality to match hdus and arrays on write https://github.com/perrygreenfield/asdf/pull/1
This creates files where multiple arrays in the tree reference the same HDU. On read, this means that modifying any of those arrays changes the same memory.

There is an additional issue of nan != nan which I think would mean that any array with nan would still be duplicated in the hdu and internal blocks.

perrygreenfield:
Good point about NaNs. Yes, the comparison would have to deal with that. The main question is there any way of avoiding doing the equality check.

As Howard mentions, it is currently unlikely the pipeline would have a problem with that since they generally copy arrays on reading. But I agree if this can be avoided it would be preferable..

braingram:
I believe the changes in my branch fixes this issue: https://github.com/braingram/asdf/tree/bug/hdu_array_link_breaking I'm rerunning the jwst tests now but it passes the minimal cases for asdf, stdatamodels and jwst and passes all tests for asdf and stdatamodels (locally).

The main changes are adding NDArrayType in util.get_array_base here:
https://github.com/asdf-format/asdf/blob/master/asdf/util.py#L89

and removing the test for NDArrayType in fits_embed during find_or_create_block_for_array here:
https://github.com/asdf-format/asdf/blob/master/asdf/fits_embed.py#L95

If we decide to go in this direction, should I open a separate PR or should we merge the changes into your branch?

I cherry picked the test for hdu independence and fixed a bug in it so it checks for that possible issue as well.

Full diff with the master branch is here: https://github.com/asdf-format/asdf/compare/master...braingram:asdf:bug/hdu_array_link_breaking?expand=1

perrygreenfield:
I've just looked your version over and it looks good to me as far as I can tell. So the main issue is that one could have a NDArrayType as a value as well, thus needing to dig down further to get the base. Eliminating the equality check is a big plus. I wonder if that first test is even necessary given the elimination of the equality check.

Just issue your own PR and we will go with that. I'll delete mine.

braingram:
Thanks for looking it over. I opened a PR. Hopefully we squished the bug!

PR #1234, opened 2022-11-21
Fix handling of NDArrayType during block to fits hdu matching

### LINKED ISSUE #1232: Unnecessary duplicate arrays when using asdf embedded in FITS files

braingram:
Fixes #1232 

When an asdf tree containing arrays is read from a fits file, the arrays are created as NDArrayType objects (instead of ndarray). If this tree was resaved to a fits file the NDArrayType was mishandled during matching arrays to hdu items which resulted in duplicate storage of array data in hdu items and internal blocks.

This PR adds a few tests for this issue and fixes the handling of NDArrayType to allow for accurate matching between hdu items and arrays to avoid data duplication.

### DIFF
--- a/asdf/tests/test_fits_embed.py
+++ b/asdf/tests/test_fits_embed.py
@@ -464,6 +464,16 @@ def test_array_view_compatible_dtype(tmp_path):
 
 
 def test_hdu_link_independence(tmp_path):
+    """
+    As links between arrays and hdu items are made during
+    saving, it's possible that if this goes wrong links
+    might be made between multiple arrays and a single hdu.
+    In this case, modifying one array will change memory
+    shared with another array. This test creates a file
+    with multiple arrays, writes it to a fits file,
+    reads it back in and then modifies the contents of
+    each array to check for this possible errort.
+    """
     asdf_in_fits = create_asdf_in_fits("f4")
     # set all arrays to same values
     asdf_in_fits["model"]["sci"]["data"][:] = 0
--- a/asdf/tests/test_fits_embed.py
+++ b/asdf/tests/test_fits_embed.py
@@ -482,8 +482,8 @@ def test_hdu_link_independence(tmp_path):
         aif["model"]["err"]["data"][:] = 3
 
         assert np.all(aif["model"]["sci"]["data"] == 1)
-        assert np.all(aif["model"]["dq"]["data"] == 1)
-        assert np.all(aif["model"]["err"]["data"] == 1)
+        assert np.all(aif["model"]["dq"]["data"] == 2)
+        assert np.all(aif["model"]["err"]["data"] == 3)
 
 
 def test_array_view_different_layout(tmp_path):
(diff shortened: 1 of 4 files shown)


perrygreenfield (approved):
LGTM

asdf/tests/test_fits_embed.py
```diff
@@ -463,6 +463,29 @@ def test_array_view_compatible_dtype(tmp_path):
             af.write_to(file_path)
 
 
+def test_hdu_link_independence(tmp_path):
```
WilliamJamieson, line 466: Could you add a nice description/documentation string like you did for the `test_resave_breaks_hdulist_tree_array_link` test? This isn't necessary, but the one in the other test was very helpful for me.

braingram: Great suggestion! I added a docstring to the test. Let me know if it looks like what you had in mind.

https://github.com/braingram/asdf/blob/bug/hdu_array_link_breaking/asdf/tests/test_fits_embed.py#L467-L476

WilliamJamieson (approved):

hbushouse:
Is this fix available yet in an ASDF release or only on github master?

perrygreenfield:
@hbushouse I'd note that this fixes only one of two problems that arose. The particular case that triggered the failure has many useless cutouts due to confused source leading to large cutouts. This fix prevents those cases from crashing the pipeline, but the results are still mostly useless, and for those cases where the segmentation is reasonable, the duplication of the data, while wasting space, doesn't cause any pipeline failures. For the pipeline to generate more meaningful results means fixing the segmentation issue (even if only rejecting any that require y sizes more than some threshold).

braingram:
@hbushouse The fix for the data duplication was released in 2.14 ([here are the release notes](https://github.com/asdf-format/asdf/releases/tag/2.14.0)), we did a patch release 2.14.1 (the current newest released version) to deal with astropy CI issues.

For jwst this PR bumped the asdf version to 2.14.1 and included a test for the data duplication: https://github.com/spacetelescope/jwst/pull/7358

hbushouse:
> For jwst this PR bumped the asdf version to 2.14.1 and included a test for the data duplication: [spacetelescope/jwst#7358](https://github.com/spacetelescope/jwst/pull/7358)

Excellent! It was the update to required jwst dependencies that I was most concerned about.

braingram:
Let me know if I can help :)
There was also a similar PR to stdatamodels (adding a test and bumping the asdf version): https://github.com/spacetelescope/stdatamodels/pull/105 However, I do not believe a new version of that was released.

PR #1237, opened 2022-11-22
Remove dead docker file

WilliamJamieson:
I noticed when making #1236 that the docker file is very out of date and not being used. This PR removes this.


braingram (approved):
LGTM

Issue #1239, opened 2022-11-23
memmapped arrays only closed at garbage collection when asdf.open given an open file


braingram:
This resulted in failure of the astropy CI as astropy.Table.open opens the file before handing it to asdf to read the table.

This regression was introduced in PR https://github.com/asdf-format/asdf/pull/1230

PR #1241, opened 2022-11-23
always close memmaps with the asdf file

braingram:
even when reading an already opened file

fixes issue #1239

update changelog

add regression test docstring

### DIFF
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -779,11 +779,9 @@ class RealFile(RandomAccessFile):
         return np.fromfile(self._fd, dtype=np.uint8, count=size)
 
     def close(self):
+        self.flush_memmap()
         super().close()
-        if self._close:
-            if hasattr(self, "_mmap") and not self._mmap.closed:
-                self._mmap.flush()
-                self._mmap.close()
+        self.close_memmap()
 
 
 class MemoryIO(RandomAccessFile):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+2.14.1 (unreleased)
+-------------------
+
+The ASDF Standard is at v1.6.0
+
+- Fix issue #1239, close memmap with asdf file context [#1241]
+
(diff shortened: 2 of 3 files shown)


asdf/generic_io.py
```diff
@@ -779,11 +779,9 @@ def read_into_array(self, size):
         return np.fromfile(self._fd, dtype=np.uint8, count=size)
 
     def close(self):
+        self.flush_memmap()
         super().close()
-        if self._close:
-            if hasattr(self, "_mmap") and not self._mmap.closed:
-                self._mmap.flush()
-                self._mmap.close()
```
WilliamJamieson, line 786: Are there any additional implications of this change?

braingram: None that I've been able to uncover. It does not appear that any other tests aside from the one added in this PR and the ones in astropy (that load a table from asdf) suffer from this issue. With 2.13 memmaps were handled by numpy.memmap which has no api for closing the memmap (as per the docstring here: https://github.com/numpy/numpy/blob/v1.23.0/numpy/core/memmap.py#L40-L43) so it's difficult if not impossible to say we've mimicked the old behavior.
Is there something specific that you had in mind?

WilliamJamieson: No, it seems like a reasonable change then.

braingram:
The added regression test fails on with asdf 2.14 (and the current master branch) with both pytest-openfiles 0.4.0 and 0.5.0 (as it checks for the number of open files directly instead of relying on the plugin). The test passes with this PR.

I'm inclined to ignore the minor coverage change (-0.01%) but am happy to attempt to update the PR with more coverage if that's preferred.

WilliamJamieson (approved):

WilliamJamieson:
> I'm inclined to ignore the minor coverage change (-0.01%) but am happy to attempt to update the PR with more coverage if that's preferred.

Coverage seems to fluctuate for strange reasons at times. I didn't see any lines drop coverage so its no big deal.

PR #1246, opened 2022-11-28
Add support for ndarray 1.1 and fix issue #1245

### LINKED ISSUE #1245: NDArrayType does not allow for supported_versions
Adding supported_versions to NDArrayType results in the following error on asdf import:
```python
  File "<string>", line 1, in <module>
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/__init__.py", line 24, in <module>
    from ._convenience import info
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/_convenience.py", line 8, in <module>
    from ._display import DEFAULT_MAX_COLS, DEFAULT_MAX_ROWS, DEFAULT_SHOW_VALUES, render_tree
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/_display.py", line 12, in <module>
    from ._node_info import create_tree
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/_node_info.py", line 4, in <module>
    from .schema import load_schema
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/schema.py", line 16, in <module>
    from . import constants, extension, generic_io, reference, tagged, treeutil, util, versioning, yamlutil
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/yamlutil.py", line 11, in <module>
    from .tags.core import AsdfObject
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/tags/core/__init__.py", line 6, in <module>
  File "/Users/bgraham/projects/221124_ndarray_supported_versions_bug/asdf/asdf/tags/core/ndarray.py", line 214, in <module>
    class NDArrayType(AsdfType):
TypeError: __class__ set to <class 'asdf.tags.core.ndarray.NDArrayType'> defining 'NDArrayType' as <class 'asdf.tags.core.ndarray.NDArrayType'>
```

The use of`super` in NDArrayType here: https://github.com/asdf-format/asdf/blob/05d5001b27766b01688f8bd4b6e01d92445e1e3c/asdf/tags/core/ndarray.py#L378 creates a `__classcell__` closure that must be propagated inside the metaclass __new__ here: https://github.com/asdf-format/asdf/blob/05d5001b27766b01688f8bd4b6e01d92445e1e3c/asdf/types.py#L86
However, ExtensionTypeMeta.__new__ creates multiple (sibling) classes when supported_versions is found in the class definition: https://github.com/asdf-format/asdf/blob/05d5001b27766b01688f8bd4b6e01d92445e1e3c/asdf/types.py#L136 These sibling classes do not have valid __classcell__ closures resulting in the above error.

One fix for the above issue is to remove the call to super in NDArrayType. We should also add an error when a class of type ExtensionTypeMeta attempts to use super.

We are planning to move away from this old type model for extensions and `__classcell__` [can't be created in python](https://stackoverflow.com/questions/55614454/classcell-generates-error-in-python-3-6-when-the-metaclass-calls-multiple-su). For these reasons it might not be worth making a fix that allow for the use of super in a class that defined supported_versions.

braingram:
This PR fixes #1245 by removing the use of super in NDArrayType and adding an exception to check for incompatible use of super in classes of type ExtensionTypeMeta.

One test is failing locally:
```python
FAILED asdf/tests/test_extension.py::test_builtin_extension - AssertionError: NDArrayType supports tag, tag:stsci.edu:asdf/core/ndarray-1...
```
due to the missing schema added in this PR: https://github.com/asdf-format/asdf-standard/pull/350 As that PR is failing due to missing support added in this PR there appears to be no way to have either pass prior to merging without skipping tests.

Additional changes to asdf will be required to fully support https://github.com/asdf-format/asdf-standard/pull/350 including an update to add supported_versions to IntegerType. This should be left as a draft until a full set of PRs is ready.

### DIFF
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -567,6 +567,9 @@ def _make_operation(name):
     return __operation__
 
 
+classes_to_modify = NDArrayType.__versioned_siblings + [
+    NDArrayType,
+]
 for op in [
     "__neg__",
     "__pos__",
@@ -631,7 +634,8 @@ for op in [
     "__delitem__",
     "__contains__",
 ]:
-    setattr(NDArrayType, op, _make_operation(op))
+    [setattr(cls, op, _make_operation(op)) for cls in classes_to_modify]
+del classes_to_modify
 
 
 def _get_ndim(instance):
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -215,6 +215,7 @@ class NDArrayType(AsdfType):
     name = "core/ndarray"
     version = "1.0.0"
     types = [np.ndarray, ma.MaskedArray]
+    supported_versions = {"1.1.0", "1.0.0"}
 
     def __init__(self, source, shape, dtype, offset, strides, order, mask, asdffile):
         self._asdffile = asdffile
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -373,10 +373,12 @@ class NDArrayType(AsdfType):
         # can cause problems when the array is passed to other
         # libraries.
         # See https://github.com/asdf-format/asdf/issues/1015
-        if name in ("name", "version"):
+        if name in ("name", "version", "supported_versions"):
             raise AttributeError(f"'{self.__class__.name}' object has no attribute '{name}'")
         else:
-            return super().__getattribute__(name)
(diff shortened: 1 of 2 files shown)


braingram:
Closing this in favor of PR #1250

PR #1247, opened 2022-11-28
use pytest to parametrize test_version_map_support

braingram:
Allow for more fine-grained control over explicit tag/extension version match failures.

test_version_map_support checks for explicit tag/extension matches. All tags currently not supported in legacy extensions fail these matches which required xfailing all checks with standard tags: https://github.com/asdf-format/asdf/blob/05d5001b27766b01688f8bd4b6e01d92445e1e3c/asdf/tests/test_versioning.py#L269-L271

As astropy.io.misc.asdf supports some of the core tags (time, unit, etc) but is a deprecated extension any updates to the schema for these tags will result in core tag test failures in test_version_map_support. This PR will allow xfailing just those tags supported outside of asdf and allow for continued testing of other core tags.

### DIFF
--- a/asdf/tests/test_versioning.py
+++ b/asdf/tests/test_versioning.py
@@ -260,47 +260,96 @@ def test_spec_equal():
     assert (1, 3, 0) == spec
 
 
-@pytest.mark.parametrize("version", supported_versions)
-def test_version_map_core_support(version):
-    _test_version_map_support(version, "core")
+def _standard_versioned_tags():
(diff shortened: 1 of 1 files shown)


WilliamJamieson (approved):

WilliamJamieson:
Please rebase so I can merge.

braingram:
Thanks for reviewing this PR. It is rebased and the CI finished.

PR #1249, opened 2022-11-29
remove uses of asdf metaschema

braingram:
The asdf metaschema: https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/asdf/asdf-schema-1.0.0.yaml
appears to only be used in a few tests and is no longer used in the ndarray schema. Modifying the tests to use the other included metaschema: https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/yaml-schema/draft-01.yaml
does not result in any test failures (locally). Removal of the asdf metaschema (which contains a reference to the ndarray schema version 1.0 which contains a bug) will simplify the fix for ndarray schema bug: https://github.com/asdf-format/asdf-standard/issues/345

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+2.14.2 (unreleased)
+-------------------
+
+The ASDF Standard is at v1.6.0
+
+- Remove uses of asdf-schema-1.0.0 metaschema [#1249]
+
 2.14.1 (2022-11-23)
 -------------------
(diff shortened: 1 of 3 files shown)


braingram:
It looks like this is used extensively in jwst, so removing the usage here won't allow removal of the metaschema from asdf-standard.

PR #1250, opened 2022-11-29
support integer-1.1 and ndarray-1.1

### LINKED ISSUE #1245: NDArrayType does not allow for supported_versions

braingram:
fixes an issue with integer where IntegerType instead of cls was used during class instantiation which resulted in objects being created with the incorrect version. Prior code (that only had one integer version, 1.0) did not suffer from issues because of this bug.

fixes #1245 by removing the use of super in NDArrayType and adding an exception to check for incompatible use of super in classes of type ExtensionTypeMeta.

This should be merged with asdf-standard PR https://github.com/asdf-format/asdf-standard/pull/350

### DIFF
--- a/asdf/tests/test_versioning.py
+++ b/asdf/tests/test_versioning.py
@@ -298,6 +298,7 @@ def xfail_version_map_support_cases(request):
         ("1.6.0", "tag:stsci.edu:asdf/fits/fits-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/time/time-1.2.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/defunit-1.1.0"),
+        ("1.6.0", "tag:stsci.edu:asdf/unit/defunit-1.0.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/quantity-1.2.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/unit-1.1.0"),
         ("1.5.0", "tag:stsci.edu:asdf/unit/defunit-1.0.0"),
--- a/asdf/tests/test_versioning.py
+++ b/asdf/tests/test_versioning.py
@@ -294,12 +294,9 @@ def xfail_version_map_support_cases(request):
     version = request.getfixturevalue("version")
     if (version, tag) in [
         ("1.6.0", "tag:stsci.edu:asdf/core/column-1.1.0"),
-        ("1.6.0", "tag:stsci.edu:asdf/core/integer-1.1.0"),
-        ("1.6.0", "tag:stsci.edu:asdf/core/ndarray-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/core/table-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/fits/fits-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/time/time-1.2.0"),
-        ("1.6.0", "tag:stsci.edu:asdf/unit/defunit-1.0.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/defunit-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/quantity-1.2.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/unit-1.1.0"),
--- a/asdf/types.py
+++ b/asdf/types.py
@@ -133,6 +133,15 @@ class ExtensionTypeMeta(type):
                     new_attrs["version"] = version
                     new_attrs["supported_versions"] = set()
                     new_attrs["_latest_version"] = cls.version
+                    if "__classcell__" in new_attrs:
+                        raise RuntimeError(
+                            "Subclasses of ExtensionTypeMeta that define "
+                            "supported_versions cannot used super() to call "
+                            "parent class functions. super() creates a "
+                            "__classcell__ closure that cannot be duplicated "
+                            "during creation of versioned siblings. "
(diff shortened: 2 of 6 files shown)


WilliamJamieson:
Now that asdf-format/asdf-standard#350 is merged can you Remove the skips added in #1251 and @1252?

asdf/tests/test_versioning.py
```diff
@@ -278,12 +278,9 @@ def xfail_version_map_support_cases(request):
...
         ("1.6.0", "tag:stsci.edu:asdf/core/column-1.1.0"),
-        ("1.6.0", "tag:stsci.edu:asdf/core/integer-1.1.0"),
-        ("1.6.0", "tag:stsci.edu:asdf/core/ndarray-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/core/table-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/fits/fits-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/time/time-1.2.0"),
-        ("1.6.0", "tag:stsci.edu:asdf/unit/defunit-1.0.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/defunit-1.1.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/quantity-1.2.0"),
         ("1.6.0", "tag:stsci.edu:asdf/unit/unit-1.1.0"),
```
braingram, line 286: I removed some of the skips added in #1251 and #1252
Removing the remaining skips would require updates to astropy.io.misc.asdf or merging the PRs in asdf-standard that remove these tags (table, column, time, unit, etc) from the standard version maps.

braingram:
It looks like this will require either a release of asdf-standard or refactoring test_builtin_extension (and readding one of the xfails for defunit).

asdf/types.py
```diff
@@ -133,6 +133,15 @@ def __new__(mcls, name, bases, attrs):
...
                     new_attrs["_latest_version"] = cls.version
+                    if "__classcell__" in new_attrs:
+                        raise RuntimeError(
+                            "Subclasses of ExtensionTypeMeta that define "
+                            "supported_versions cannot used super() to call "
+                            "parent class functions. super() creates a "
+                            "__classcell__ closure that cannot be duplicated "
+                            "during creation of versioned siblings. "
+                            "See https://github.com/asdf-format/asdf/issues/1245"
+                        )
```
WilliamJamieson, line 144: Can you write a test that raises this error?

WilliamJamieson (changes requested):

asdf/types.py
braingram: Sure thing. I added one to 80d5904

WilliamJamieson (approved):

PR #1253, opened 2022-11-30
skip testing type versions not in any deployed standard

braingram:
Modifies _assert_extension_type_correctness to skip checking type versions that are not part of a non-development supported version of the asdf-standard.

Two other tests are added `test_default_in_supported_versions` and `test_development_is_not_default` to catch possible errors where the default is set to a non-supported (or development) version.

As this PR will skip testing completeness (that a type has a schema of the same version) of type versions that are only in development version, it's possible that updating the development version of the asdf-standard will cause test failures if the schema and type versions do not match. Put another way, a PR that advances the asdf-standard development version might reveal and then have to fix these type/schema version issues.

This should allow PR #1250 to pass CI without an asdf-standard release.

### DIFF
--- a/asdf/tests/helpers.py
+++ b/asdf/tests/helpers.py
@@ -31,7 +31,13 @@ from ..exceptions import AsdfConversionWarning
 from ..extension import default_extensions
 from ..resolver import Resolver, ResolverChain
 from ..tags.core import AsdfObject
-from ..versioning import AsdfVersion, get_version_map
+from ..versioning import (
+    AsdfVersion,
+    asdf_standard_development_version,
+    get_version_map,
+    split_tag_version,
+    supported_versions,
+)
 from .httpserver import RangeHTTPServer
 
 try:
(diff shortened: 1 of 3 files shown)


braingram:
It might make more sense to add this to #1250 so we can verify it passes CI.

If the CI here passes we'll at least have confirmation that it didn't add any new failures to the master branch.

asdf/tests/helpers.py
```diff
@@ -436,7 +442,29 @@ def _assert_extension_type_correctness(extension, extension_type, resolver):
         extension_type.__name__
     )
 
-    for check_type in extension_type.versioned_siblings + [extension_type]:
+    # check the default version
+    types_to_check = [
+        extension_type,
```
WilliamJamieson, line 447:
```suggestion
        extension_type
```
`black` should make this less ugly now.

WilliamJamieson (changes requested):

asdf/tests/helpers.py
braingram: Thanks! I made this change and reran black in 472b0b7

braingram:
PR #1255 (the combination of this PR and #1250) passed CI which suggests that the changes included in this PR should be sufficient to allow #1250 CI to pass.

WilliamJamieson (approved):

WilliamJamieson:
close/reopen to attempt to reset doc build

PR #1257, opened 2022-12-05
Bugfix for `enum` bug

### LINKED ISSUE #1256: `enum` fails to validate for `asdf.tagged.TaggedString`
This issue was initially described in spacetelescope/rad#155 and now is needed to implement the work described by https://github.com/spacetelescope/rad/pull/192#discussion_r1038209723.

The work involved in implementing the unit specifiers in spacetelescope/rad#192 gives a much clearer picture of the issue. If one adds:

```yaml
      unit:
        tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
        enum: ["DN"]
```

to the `amp33` quantity in say `guidewindow-1.0.0.yaml` in that spacetelescope/rad#192 and then runs the tests for `roman_datamodels` using the `roman_datamodels` branch from spacetelescope/roman_datamodels#111, one will get several failures from this `enum` statement. In particular if one runs

```pytest "tests/test_factories.py::test_instance_valid[Guidewindow]"```

One gets the error message:
```
jsonschema.exceptions.ValidationError: 'DN' is not one of ['DN']
```

Which is a bit confusing as this message seems like nonsense.

Looking into this deeper, one can find that this message ultimately originates from this function https://github.com/python-jsonschema/jsonschema/blob/dbc398245a583cb2366795dc529ae042d10c1577/jsonschema/_validators.py#L279-L285 namely:
```python
    elif instance not in enums:
        yield ValidationError(f"{instance!r} is not one of {enums!r}")
```

Which leads to the conclusion that something in the `instance not in enums` is the root of the problem. Diving deeper into this, one discovers that `instance` is actually of type`asdf.tagged.TaggedString` while all of the values in `enum` are of type `str`. If one then examines`asdf.tagged.TaggedString`'s equality method: https://github.com/asdf-format/asdf/blob/ad6697147366bc77b5d9cf3c62871324bc1db20a/asdf/tagged.py#L102-L103
One finds that for a `asdf.tagged.TaggedString` can only be equal to something of type `asdf.tagged.TaggedString` which results in the truth of the `instance not in enums` statement, which leads to the failure.

Indeed, if one replaces the equality function with:
```python
    def __eq__(self, other):
        if isinstance(other, TaggedString):
            return str.__eq__(self, other) and self._tag == other._tag
        elif isinstance(other, str):
            return str.__eq__(self, other)
        else:
            return False
```
(which allows for equality between a string and tagged string). Then the `enum` error disappears.

Inspection of the other objects in `asdf.tagged` reveals that this issue may technically extend to those as well. Clearly, it makes sense to be able to tag these types of objects and use `enum` to limit the values in some cases (e.g. like what we want to do in `rad`), thus this is a bug in how we handle tags.

I see two possibliites:

1. Patch the `__eq__` for each of the `asdf.tagged` objects to allow comparison with their un-tagged counter parts like what I did above to demonstrate the error.
2. See if we can patch validation so that we pass the pure "untagged objects" down two `jsonschema`.

WilliamJamieson:
In spacetelescope/rad#155, a bug with using `enum` was reported. It of course had an unhelpful error message. The details of the bug's cause and possible resolution are detailed in issue #1256. This PR fixes #1256 by implemented one of possible paths (path 3) for resolution and adds regression tests for this behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+2.14.2 (unreleased)
+-------------------
+
+The ASDF Standard is at v1.6.0
+
+- Fix issue #1256, where ``enum`` could not be used on tagged objects. [#1257]
+
 2.14.1 (2022-11-23)
 -------------------
 
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -150,6 +150,17 @@ def validate_type(validator, types, instance, schema):
     return mvalidators.Draft4Validator.VALIDATORS["type"](validator, types, instance, schema)
 
 
+def validate_enum(validator, enums, instance, schema):
+    """
+    `asdf.tagged.Tagged` objects will fail in the default enum validator
+    """
+
+    if isinstance(instance, tagged.Tagged):
+        instance = instance.base
+
+    yield from mvalidators.Draft4Validator.VALIDATORS["enum"](validator, enums, instance, schema)
+
+
 YAML_VALIDATORS = util.HashableDict(mvalidators.Draft4Validator.VALIDATORS.copy())
 YAML_VALIDATORS.update(
     {
@@ -158,6 +169,7 @@ YAML_VALIDATORS.update(
         "flowStyle": validate_flowStyle,
         "style": validate_style,
         "type": validate_type,
+        "enum": validate_enum,
     }
 )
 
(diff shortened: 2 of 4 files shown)


braingram (approved):
Looks good to me.

Issue #1254, opened 2022-12-01
duplicate extensions in asdf 2.14


braingram:
In certain circumstances, weldx.asdf.utils.get_weldx_extension fails due to duplicate extensions: https://github.com/BAMWelDX/weldx/issues/825

Installing a local clone of the weldx repository and running the following while in the repository parent directory produces the error:
```python
import asdf
import weldx

af = asdf.AsdfFile()
ctx = af._create_serialization_context()
weldx.asdf.util.get_weldx_extension(ctx)
```

```python
  File "/Users/bgraham/projects/221201_weldx_downstream_test/weldx/get_extension.py", line 6, in <module>
    weldx.asdf.util.get_weldx_extension(ctx)
  File "/Users/bgraham/projects/221201_weldx_downstream_test/weldx/weldx/asdf/util.py", line 459, in get_weldx_extension
    raise ValueError("Could not determine correct weldx extension.")
ValueError: Could not determine correct weldx extension.
```

Running the same code outside the weldx repository does not result in an error. There appears to be an issue with importlib:
https://github.com/pypa/setuptools/issues/3649

Forcing the use of importlib-metadata==4.11.4 (installed via pip, tested with python 3.10.6) by modifying these lines fixes the issue: https://github.com/asdf-format/asdf/blob/ad6697147366bc77b5d9cf3c62871324bc1db20a/asdf/entry_points.py#L4-L7

It is unclear to me if this is intended behavior of importlib or a bug that per this comment appears to be in the python stdlib https://github.com/pypa/setuptools/issues/3649#issuecomment-1294229882

I'm not sure if other extensions are being duplicated or if this duplication is causing any issues (aside from the one in weldx).

One possible fix would be to put a lower pin on importlib-metadata and to not use the stdlib importlib for python 3.10/3.11.

marscher:
Thanks! This precisely describes the problem and the workaround of only relying on one dependency (importlib_metadata) seems like a good approach.

I think the duplication of the extensions is not a problem per se. It would be hard to tell which extension is used then (there is some sorting going on somewhere, but you cannot be sure, which version you'll get). In weldx there is an explicit check, that we only receive one (matching) extension, so the underlying "problem" actually appears.

braingram:
Fixed in #1260

PR #1260, opened 2022-12-06
update lower pin to importlib_metadata for all python versions

braingram:
importlib in the standard library and importlib_metadata prior to 4.11.4 library return duplicate entry points in some cases. This leads to duplication of extensions in entry_points.get_extensions.

remove use of importlib for python >= 3.10 and require use of importlib_metadata >= 4.11.4

fixes issue #1254


asdf/tests/test_entry_points.py
```diff
@@ -1,12 +1,6 @@
-import sys
-
+import importlib_metadata as metadata
 import pytest
 
-if sys.version_info < (3, 10):
-    import importlib_metadata as metadata
-else:
-    import importlib.metadata as metadata
-
```
WilliamJamieson, line 9: I do not like this, we should be using the builtin version of `importlib.metadata` whenever possible. I seems like you need v4.11 of the `importlib-metadata` package due to the bug. Looking at this table: https://github.com/python/importlib_metadata#compatibility, python 3.11 has a new enough version of `importlib-metadata` builtin in order to avoid the bug.

Please change the `version_info` check to `(3, 11)` instead of removing this.

pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = [
     'asdf-transform-schemas >=0.2.2',
     'asdf-unit-schemas >= 0.1.0',
     'importlib_resources >=3; python_version <"3.9"',
-    'importlib-metadata >=3; python_version <"3.10"',
+    'importlib-metadata >=4.11.4',
```
WilliamJamieson, line 21:
```suggestion
    'importlib-metadata >=4.11.4; python_version<"3.11"',
```

See above comment

WilliamJamieson (changes requested):
This will also need a regression test for the issue #1254

braingram:
Do you have something in mind for a regression test?

Should we test for duplicate entry points?

If instead we test for duplicate extensions, what should be considered a duplicate vs an extension that supports the same types (and versions of those types) to override the default behavior (as explained in the [docs](https://asdf.readthedocs.io/en/stable/asdf/extending/legacy.html#overriding-built-in-extensions))?

I'd appreciate any additional details you can provide on this.

WilliamJamieson:
> Do you have something in mind for a regression test?
> 
> Should we test for duplicate entry points?
> 
> If instead we test for duplicate extensions, what should be considered a duplicate vs an extension that supports the same types (and versions of those types) to override the default behavior (as explained in the [docs](https://asdf.readthedocs.io/en/stable/asdf/extending/legacy.html#overriding-built-in-extensions))?
> 
> I'd appreciate any additional details you can provide on this.

In https://github.com/asdf-format/asdf/issues/1254#issue-1471537642 you claim:

```python
import asdf
import weldx

af = asdf.AsdfFile()
ctx = af._create_serialization_context()
weldx.asdf.util.get_weldx_extension(ctx)
```
Produces an error which is then fixed by this PR. These steps should be reducible to a (possibly contrived) minimal reproducer, independent of `weldx` which reproduces this error. This minimal reproducer can be turned into a regression test.

Currently, we don't know what exactly is causing the problem only that the change in this PR fixes it.

braingram:
Thanks for the explanation.

The issue is a bug in importlib_metadata that occurs (at least in one instance) when the example in #1254 is run with weldx installed and in the weldx source directory. It is similar to https://github.com/python/importlib_metadata/issues/410 which was fixed in https://github.com/python/importlib_metadata/pull/379

This comment mentions that 3.11 may also have issues https://github.com/python/importlib_metadata/issues/410#issuecomment-1304258228 so relying on the standard library might not work and will require more testing.

One regression test I could imagine might work to reproduce the issue would involve creating a package, installing it, and producing a matching .egg-info to provide the conditions for the duplicate entry points. This would involve significant test infrastructure for what appears to be a bug in a dependency. Can you think of an alternative? If not, I don't think the complexity required to add this test is worth what it provides.

WilliamJamieson:
The OpenAstronomy [workflow](https://github.com/OpenAstronomy/github-actions-workflows#build-and-publish-a-pure-python-package) we use can be used to test a package using the installed version, which we should be doing anyways.

If it's too much work I understand, but I prefer having a minimal reproducer for regressions so that we don't have repeat issues.

braingram:
Testing with 3.11 it also returns duplicate entry points
```python
(weldx_3.11) aerum2022:asdf bgraham$ ipython
iPython 3.11.0 (main, Dec  7 2022, 10:03:05) [Clang 14.0.0 (clang-1400.0.29.102)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.7.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import importlib.metadata

In [2]: importlib.metadata.entry_points(group='asdf_extensions')
Out[2]:
[EntryPoint(name='builtin', value='asdf.extension:BuiltinExtension', group='asdf_extensions'),
 EntryPoint(name='builtin', value='asdf.extension:BuiltinExtension', group='asdf_extensions'),
 EntryPoint(name='astropy', value='astropy.io.misc.asdf.extension:AstropyExtension', group='asdf_extensions'),
 EntryPoint(name='astropy-asdf', value='astropy.io.misc.asdf.extension:AstropyAsdfExtension', group='asdf_extensions')]
 ``` 

 We could do one or both of the following:
 1) change the code in asdf to not assume entry points will not be duplicated
 2) require importlib_metadata>=4.11.4 for all python versions

Writing a regression test for 1 would not add significant complexity to the tests (as we already mock entry points in test_entry_points).

@WilliamJamieson are there other options and which do you prefer?

WilliamJamieson:
> @WilliamJamieson are there other options and which do you prefer?

Hmm... I think we should remove the `if` wrapper choosing importing `importlib_metadata` vs `importlib.metadata` (maybe comment it out?). While doing so we should add a comment that links back to issue #1254 and this PR for details on reproducing the issue (maybe leave #1254 open after this PR is closed?). We can try to re-introduce using the builtin `importlib.metadata` when future versions of Python are released. This  would also require changing the `importlib-metadata` to be for all python versions.

I think step 1 is a good idea also, though I think that change is outside the scope of this PR. Let's open a separate issue/PR to track doing that separately from this. Once that is done, we should revisit #1254 and the changes from this PR.

WilliamJamieson (approved):

braingram:
Closing and reopening to trigger CI as previous failures for mac osx test appear unrelated to code changes (it appears to be a network timeout).

WilliamJamieson:
> The remaining CI failure is the codecov/project check failing due to a insignificant change in coverage.

Coverage changes are not that important (hence why they are not required CI to pass). In reality this change should not effect coverage at all but it does reduce the number of lines in a imperfectly covered file meaning there is a tiny shift in converage.

We really need to set the thresholds better.

PR #1262, opened 2022-12-08
update tox.ini for 4

braingram:
applied fix for missing distdir, likely will see other failures which this PR will test for

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
     with:
       submodules: false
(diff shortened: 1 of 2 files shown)


braingram:
Compatibility tests were failing in tox 4 because the call to `virtualenv.cli_run` was leaving an open file. Switched this call to subprocess to work around the issue.

zacharyburnett:
now that Tox is using PEP517 by default it might be good to add `python -m build` to the `check-build` testenv (adding `build` to `deps`)

zacharyburnett:
> Now that #1266 is in, our packaging tests will actually build the wheels and install them and then test against those wheels. So this is done implicitly I think.

oh, nice!

WilliamJamieson (approved):

PR #1265, opened 2022-12-09
pin tox to <4

braingram:


braingram:
I'm not seeing usage of tox in downstream or s390x. I will add labels to trigger those just in case.

WilliamJamieson (approved):

PR #1270, opened 2022-12-12
add codecov.yml and set threshold to 1%

braingram:


WilliamJamieson (approved):
LGTM.

I dislike having yet another configuration file in the top-level directory, but this is better than annoying codecov failures for no reason.

braingram:
Should we hide it in `.github` on one of the other options here?:
https://github.com/codecov/codecov-action/issues/554#issuecomment-1261250304

WilliamJamieson:
> Should we hide it in `.github` on one of the other options here?: [codecov/codecov-action#554 (comment)](https://github.com/codecov/codecov-action/issues/554#issuecomment-1261250304)

This is much better

WilliamJamieson (approved):

Issue #1013, opened 2021-08-14
controlling array storage in to_yaml_tree methods


CagtayFabry:
In the legacy API the complete `AsdfFile` context was accessible in the `to_tree(cls, node, ctx: AsdfFile)` classmethods
This allowed to control the array storage mode (e.g. `"inline"`) for individual numpy arrays directly during the `to_tree` call like so
```python
    @classmethod
    def to_tree(cls, node, ctx: AsdfFile):
        if node.array.ndim < 3: # let's assume node.array is np.ndarray
            ctx.set_array_storage(node.array, "inline")
        return {"data" : node.data}
```

With the new API `ctx` in `to_yaml_tree(self, obj, tag, ctx)` is now `asdf.asdf.SerializationContext` which doesn't allow setting the array storage mode.
Is there a way to pass the option down to the `AsdfFile` from `to_yaml_tree(self, obj, tag, ctx)` ?

It would be great to have this control again

eslavich:
I agree that we should support this, probably by adding a similar method to SerializationContext.  The reason why we'd like to avoid calling `AsdfFile.set_array_storage` is that it modifies the currently open AsdfFile instance, which isn't supposed to happen when the file is being written using `AsdfFile.write_to`.

PR #1273, opened 2022-12-13
Refactor/block api

braingram:
This PR updates several block/block_manager functions to ease asdf/zarr integration. These changes so far include:
- split block.py into block.py and block_manager.py 7947591
- add block option to disable caching e7f2eda
- add the block_manager to the SerializationContext passed to extensions 1eff6ff

The corresponding updates to the experimental asdf_zarr extension can be found here: https://github.com/braingram/asdf_zarr/tree/no_write_context


braingram:
This PR does not include a DeferredBlockSource and instead relies on [patching block.write](https://github.com/braingram/asdf_zarr/blob/3f56eb0620048373d1c025e67bc312f7dfab33a1/src/asdf_zarr/converter.py#L59-L70) to allow the converter to add blocks during `to_yaml_tree`.

One option for removing the need for this patching would be to allow blocks to have a data callback (similar to the DeferredBlockSource) which is resolved when needed and optionally cached.

braingram:
I'm adding some notes here on my thoughts for further developing this refactor.

Block objects currently have a combination of state and configuration. There are many possible combinations of these states and configurations that lead to many possible situations where a block can be used including:
- Read, Inline (this is all in the tree, so doesn’t use a block on disk)
- Write, Inline (this is all in the tree, so doesn’t use a block on disk)
- Read, Seekable, Lazy Loaded, Memmapped, Internal
- Read, Seekable, Lazy Loaded, Non-memmapped, Internal
- Read, Seekable, Non-Lazy Loaded, Memmapped, Internal
- Read, Seekable, Non-Lazy Loaded, Non-memmapped, Internal
- Read, Non-Seekable, Non-Lazy Loaded, Non-memmapped, Internal
- Read, Seekable, Lazy Loaded, Memmapped, Stream
- Read, Seekable, Lazy Loaded, Non-memmapped, Stream
- Read, Seekable, Non-Lazy Loaded, Memmapped, Stream
- Read, Seekable, Non-Lazy Loaded, Non-memmapped, Stream
- Read, Non-Seekable, Non-Lazy Loaded, Non-memmapped, Stream
- Write, Seekable, Internal
- Write, Non-Seekable, Internal
- Write, Seekable, Stream
- Write, Non-Seekable, Stream
- etc (including external blocks, compression)

It might be beneficial (hopefully with little cost) to drop certain unused features including:

- Non-seekable files
- External blocks
- Stream blocks
- Ability to disable lazy loading

Additionally, refactoring of the Block object into two objects might simplify interactions with the blocks. These would be:
- BlockConfig: used to store defined configuration (memmap, compression, lazy load, etc)
- BlockState: used to store block read state (has the data and/or header been read, data verified, cached, where is the data in the file)

Attempts have been made in this PR to simplify (without losing features) the block code to work towards better defining the above plan.

One goal of the refactor is to produce a consistent internal api that can be used by NDAarrayType from which we can select portions to make public to allow extensions to access block storage (like for zarr).

@eslavich @perrygreenfield and all else opinions and discussion here are appreciated.

CagtayFabry:
a quick question after reading over description and comments @braingram

>  add the block_manager to the SerializationContext passed to extensions 1eff6ff

If I understand correctly this would open up block metadata/config access of converters in `to_yaml_tree` again, basically solving  #1013?
(given that one can identify the right block?)

braingram:
@CagtayFabry Thanks for looking it over!

Yes, solving #1013 is one of the goals of this refactoring. I'm currently doubtful this PR will be merged and instead more manageable pieces will be pulled out into separate PRs. Some other goals of this refactoring are:
- prevent write_to from modifying the asdf object (the issue that prevented a straightforward fix to #1013)
- allow for extensions that add/read/manage blocks (to make a zarr extension that uses internal blocks possible)

PR #1274, opened 2022-12-14
Switch to using `tox` and openastronomy for downstream

WilliamJamieson:
Switches to using tox environments and openastronomy for downstream tests.

Note: the required CI passes will need to be updated after this PR is merged. So ignore required labels for now.


WilliamJamieson:
@braingram I think the failure in the `Downstream/stsci/stdatamodels` job is related to the fix you proposed in spacetelescope/stdatamodels#104. I think we need to revisit that PR and merge that fix.

braingram:
> @braingram I think the failure in the `Downstream/stsci/stdatamodels` job is related to the fix you proposed in [spacetelescope/stdatamodels#104](https://github.com/spacetelescope/stdatamodels/pull/104). I think we need to revisit that PR and merge that fix.

Is there a new dependency that is triggering the install of asdf-astropy for this test that causes the issue?

WilliamJamieson:
> > @braingram I think the failure in the `Downstream/stsci/stdatamodels` job is related to the fix you proposed in [spacetelescope/stdatamodels#104](https://github.com/spacetelescope/stdatamodels/pull/104). I think we need to revisit that PR and merge that fix.
> 
> Is there a new dependency that is triggering the install of asdf-astropy for this test that causes the issue?

Yes, `tox` is installing `asdf` using its `tests` dependency as it should.

braingram:
> > > @braingram I think the failure in the `Downstream/stsci/stdatamodels` job is related to the fix you proposed in [spacetelescope/stdatamodels#104](https://github.com/spacetelescope/stdatamodels/pull/104). I think we need to revisit that PR and merge that fix.
> > 
> > 
> > Is there a new dependency that is triggering the install of asdf-astropy for this test that causes the issue?
> 
> Yes, `tox` is installing `asdf` using its `tests` dependency as it should.

What's the reasoning behind changing the install to use the `test` dependency for the downstream CI? Wouldn't an install of asdf from pypi not include those dependencies?

WilliamJamieson:
> What's the reasoning behind changing the install to use the `test` dependency for the downstream CI? Wouldn't an install of asdf from pypi not include those dependencies?

Because thats how our toxenv is setup. It installs `asdf` from the current state using `tests` and `all`.

Anything taking issue with `asdf-astropy` is a bug and needs to be resolved anyways. Fall back on `astropy.io.misc.asdf` will cense to work next October.

tox.ini
```diff
@@ -47,3 +47,158 @@ extras= all,tests
...
+commands=
+    git clone https://github.com/spacetelescope/jwst.git
+    pip install -e jwst[test]
+    pip freeze
+    pytest jwst
+
+[testenv:stdatamodels]
+changedir={envtmpdir}
+allowlist_externals=
+    git
```
braingram, line 159:
```suggestion
    git
extras=
```
This would fix the issue with stdatamodels failing because of the included deps from the `extras= all,tests`.

It makes sense to me to have this for all downstream CI tests.

braingram (approved):
Looks good to me.

zacharyburnett (approved):
looks very comprehensive; one thing, should these blocks:
```
git clone https://github.com/spacetelescope/<package>.git
pip install -e <package>[test]
pip freeze
```
go in `commands_pre` instead of in `commands`?

WilliamJamieson:
> looks very comprehensive; one thing, should these blocks:
> 
> ```
> git clone https://github.com/spacetelescope/<package>.git
> pip install -e <package>[test]
> pip freeze
> ```
> 
> go in `commands_pre` instead of in `commands`?

Good idea

zacharyburnett (approved):
LGTM, looks like a much better structure for downstream tests

PR #1275, opened 2022-12-15
Update changes for 2.14.3 release

WilliamJamieson:
This release is to release the fix for BAMWelDX/weldx#825 so that it can be used by weldx.


WilliamJamieson:
Lets wait on #1278 to merge and then rebase on that.

PR #1279, opened 2022-12-15
TST fix s390x build

WilliamJamieson:
Attempt to fix s390x build.

Changes based on astropy/astropy#13940 and this part of the astropy CI: https://github.com/astropy/astropy/blob/5b2915e22de504280bdaf573464b2e99f49f7ab6/.github/workflows/ci_cron_weekly.yml#L91-L145

### DIFF
--- a/.github/workflows/s390x.yml
+++ b/.github/workflows/s390x.yml
@@ -18,24 +18,37 @@ on:
       - '*'
 
 jobs:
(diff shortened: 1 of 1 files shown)


braingram (approved):
Looks good to me. Nice find for the easy add of aarch64.

PR #1280, opened 2022-12-19
Discard cache of lazy-loaded block data when it is no longer referenced by the tree

eslavich:
This wraps the Block class's `_data` reference in a weakref when reading in lazy_load=True and copy_arrays=False mode.  There isn't any benefit when reading ndarray but this will allow us to process large chunked arrays without memory mapping.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+2.15.0 (unreleased)
+-------------------
+
+The ASDF Standard is at v1.6.0
+
+- Discard cache of lazy-loaded block data when it is no longer referenced
+  by the tree. [#1280]
+
 2.14.3 (2022-12-15)
 -------------------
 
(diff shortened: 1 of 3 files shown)


eslavich:
The astropy@main failures appear unrelated (the same tests are failing in other recent PRs on astropy).

braingram:
This a clever way to handle the caching. I spent some time testing it and it looks like it should be compatible with chunking access (I wrote a test script that provides the blocks to zarr by accessing block.data and memory was freed as expected).

I'll close and reopen this to re-trigger the CI. The astropy issues might be fixed (they may have been related to the new numpy version).

WilliamJamieson:
@eslavich I like these changes, they are very clever. Can you rebase them?

WilliamJamieson (approved):

WilliamJamieson (approved):

PR #1283, opened 2022-12-30
count blocks._internal_blocks during update

braingram:
Instead of searching the tree for NDArrayType instances to count the number of internal blocks to estimate the tree size, count the internal blocks directly.

This should be more efficient (by removing 1 tree traversal during update) and should make adding block-enabled extensions easier (by removing a case where NDArrayType is specially handled and assumed to be the only block-using object).

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -6,7 +6,6 @@ import pathlib
 import time
 import warnings
 
-import numpy as np
 from jsonschema import ValidationError
 from packaging.version import Version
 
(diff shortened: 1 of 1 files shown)


WilliamJamieson (approved):
Looks good to me, I'm running the downstream just to be sure (the jwst downstream job has some CRDS issue that needs to be fixed so it will fail due to that).

WilliamJamieson:
Now that #1274 is merged can this be rebased one more time?

PR #1284, opened 2022-12-30
remove redundant Block.readonly

braingram:
Block.readonly appears to only be used to unset the write flag during ndarray creation:
https://github.com/asdf-format/asdf/blob/master/asdf/tags/core/ndarray.py#L267-L268

This flag is already set for readonly memmaps (perhaps as a result of #1230) making the Block.readonly flag redundant.

This PR removes the Block.readonly flag and the associated code that sets it and adds to test_readonly to explicitly force memmapping (rather than relying on the defaul).

### DIFF
--- a/asdf/tags/core/tests/test_ndarray.py
+++ b/asdf/tags/core/tests/test_ndarray.py
@@ -874,9 +874,14 @@ def test_readonly(tmpdir):
     # Opening in read mode (the default) should mean array is readonly
     with asdf.open(tmpfile) as af:
         assert af["data"].flags.writeable is False
-        with pytest.raises(ValueError) as err:
+        with pytest.raises(ValueError, match=r"assignment destination is read-only"):
             af["data"][0] = 41
-            assert str(err) == "assignment destination is read-only"
+
+    # Forcing memmap, the array should still be readonly
+    with asdf.open(tmpfile, copy_arrays=False) as af:
+        assert af["data"].flags.writeable is False
+        with pytest.raises(ValueError, match=r"assignment destination is read-only"):
+            af["data"][0] = 41
 
     # This should be perfectly fine
     with asdf.open(tmpfile, mode="rw") as af:
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -61,7 +61,6 @@ class AsdfFile:
         copy_arrays=False,
         lazy_load=True,
         custom_schema=None,
-        _readonly=False,
     ):
         """
         Parameters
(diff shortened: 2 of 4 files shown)


WilliamJamieson:
Now that #1274 is merged this needs to be rebased.

asdf/tags/core/tests/test_ndarray.py
```diff
@@ -878,6 +878,13 @@ def test_readonly(tmpdir):
             af["data"][0] = 41
             assert str(err) == "assignment destination is read-only"
 
+    # Forcing memmap, the array should still be readonly
+    with asdf.open(tmpfile, copy_arrays=False) as af:
+        assert af["data"].flags.writeable is False
+        with pytest.raises(ValueError) as err:
+            af["data"][0] = 41
+            assert str(err) == "assignment destination is read-only"
```
WilliamJamieson, line 886:
```suggestion
        with pytest.raises(ValueError, match=r"assignment destination is read-only"):
            af["data"][0] = 41
```
The `match` API is a little nicer

WilliamJamieson (changes requested):

asdf/tags/core/tests/test_ndarray.py
braingram: Thanks! I updated the new test and the one above it that performs the same check with defaults arguments to asdf.open.

asdf/tags/core/tests/test_ndarray.py
```diff
@@ -874,16 +874,15 @@ def test_readonly(tmpdir):
     # Opening in read mode (the default) should mean array is readonly
     with asdf.open(tmpfile) as af:
         assert af["data"].flags.writeable is False
-        with pytest.raises(ValueError) as err:
+        with pytest.raises(ValueError, match=r"assignment destination is read-only") as err:
             af["data"][0] = 41
             assert str(err) == "assignment destination is read-only"
```
WilliamJamieson, line 879: This line is actually unreachable as the error will occur on 878 and the context will end there. If you remove this line you should also remove the `as err` on 877 too

asdf/tags/core/tests/test_ndarray.py
```diff
@@ -874,16 +874,15 @@ def test_readonly(tmpdir):
...
 
     # Forcing memmap, the array should still be readonly
     with asdf.open(tmpfile, copy_arrays=False) as af:
         assert af["data"].flags.writeable is False
-        with pytest.raises(ValueError) as err:
+        with pytest.raises(ValueError, match=r"assignment destination is read-only") as err:
```
WilliamJamieson, line 884: Remove the `as err` here too.

WilliamJamieson (changes requested):
Everything else looks good. Just a few more notes

asdf/tags/core/tests/test_ndarray.py
braingram: Thanks for finding the issues. I updated the test code removing the `as err` lines and the unreachable line.

WilliamJamieson (approved):

PR #1288, opened 2023-01-08
Drop support for ASDF-in-FITS

eslavich:
This PR removes support from ASDF-in-FITS from this library.  We'll be moving that feature up to https://github.com/spacetelescope/stdatamodels where the implementation is a good deal cleaner.


braingram:
I'm looking around for other uses of AsdfInFits so we can sort out how to either fix those uses to not require AsdfInFits or reintroduce an AsdfInFits object somewhere (possibly stdatamodels). So far I've found 2 uses outside stdatamodels:
- jdaviz: https://github.com/spacetelescope/jdaviz/blob/20b224e616f61f0fd4245807560d7025aefc6867/jdaviz/configs/imviz/plugins/parsers.py#L232
- hcipy: https://github.com/ehpor/hcipy/blob/058c473b31ae094f44ad1a3e445f6773c18f0d8a/hcipy/util/io.py#L140

WilliamJamieson:
@eslavich I made a bunch of style updates to ASDF (sorry), can you go through rebase/fix the conflicts. They shouldn't be too terrible.

braingram:
The roman downstream failure is expected and should be fixed with this PR (it is awaiting fixes to regression tests to allow those to be run against the PR before merging):
https://github.com/spacetelescope/roman_datamodels/pull/126

The specutils downstream failure is expected as it uses AsdfInFits.

The weldx failure is due to use of undocumented parts of asdf.
https://github.com/asdf-format/asdf/actions/runs/4347187875/jobs/7594191808#step:9:380
It is using `asdf.util.get_file_type` and `asdf.util.FileType` which this PR removed (presumably because the only remaining types were `ASDF` and `UNKNOWN`).

braingram:
> The weldx failure is due to use of undocumented parts of asdf.
> https://github.com/asdf-format/asdf/actions/runs/4347187875/jobs/7594191808#step:9:380
> It is using `asdf.util.get_file_type` and `asdf.util.FileType` which this PR removed (presumably because the only remaining types were `ASDF` and `UNKNOWN`).

I pushed a commit to re-add these parts of the API as they're not required for the other changes. Weldx appears to use these to detect if a file is ASDF which generally seems useful.. Removal could be discussed in a separate PR.

braingram:
A PR is open for specutils that should fix the CI failures by adding stdatamodels as a dependency and using it to support ASDF-in-FITS files: https://github.com/astropy/specutils/pull/1029

braingram (approved):
LGTM. A second review would be appreciated given the scope of the changes.

asdf/asdf.py
```diff
@@ -807,9 +807,9 @@ def _parse_header_line(cls, line):
 
         try:
             version = versioning.AsdfVersion(parts[1].decode("ascii"))
-        except ValueError as err:
+        except ValueError:
             msg = f"Unparsable version in ASDF file: {parts[1]}"
-            raise ValueError(msg) from err
+            raise ValueError(msg) from None
```
WilliamJamieson, line 812: Why has this been changed to `from None`?

WilliamJamieson (changes requested):

asdf/asdf.py
braingram: I reverted this change.

WilliamJamieson (approved):

PR #1289, opened 2023-01-08
Add _tagged_tree_transform argument to asdf.open

eslavich:
This adds a `_tagged_tree_transform` argument to `asdf.open` that permits arbitrary transformations of the tagged tree before it is converted.  We'll need this in order to move ASDF-in-FITS support to stdatamodels.

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -810,7 +810,7 @@ class AsdfFile:
         _force_raw_types=False,
         strict_extension_check=False,
         ignore_missing_extensions=False,
-        tree_mapper=None,
+        _tagged_tree_transform=None,
         **kwargs,
     ):
         """Attempt to populate AsdfFile data from file-like object"""
@@ -882,8 +882,8 @@ class AsdfFile:
                 self._blocks.read_internal_blocks(fd, past_magic=True, validate_checksums=validate_checksums)
                 self._blocks.read_block_index(fd, self)
 
-            if tree_mapper is not None:
-                tree = tree_mapper(tree)
+            if _tagged_tree_transform is not None:
+                tree = _tagged_tree_transform(tree)
 
             tree = reference.find_references(tree, self)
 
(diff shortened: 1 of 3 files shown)


asdf/asdf.py
```diff
@@ -867,6 +868,9 @@ def _open_asdf(
                 self._blocks.read_internal_blocks(fd, past_magic=True, validate_checksums=validate_checksums)
                 self._blocks.read_block_index(fd, self)
 
+            if tree_mapper is not None:
```
braingram, line 871: If I'm following this correctly, `tree_mapper` is only used to transform the tagged tree on read. Is that correct? If so, can we call it `tagged_tree_transform`?

asdf/asdf.py
```diff
@@ -867,6 +868,9 @@ def _open_asdf(
...
+                tree = tree_mapper(tree)
```
braingram, line 872: This seems like a reasonable place for a `provisional api` warning. I opened a PR with a suggested format: #1290
Another option would be to make the `tree_mapper` argument '_tree_mapper' to discourage it's use in case we need to make changes.

braingram:
Given the target ASDF 3.0 version, should this be against the 3.0 branch? If that branch is sufficiently dead, should we make a new one?

WilliamJamieson:
> Given the target ASDF 3.0 version, should this be against the 3.0 branch? If that branch is sufficiently dead, should we make a new one?

This is a good question, 3.0 has a lot of good stuff in it... This is something we ought to discuss soon as I think we can automate a lot of the issues causing the deadness of 3.0 away. I'll open an issue about this.

WilliamJamieson:
Here's a thought. What if we call dropping asdf-in-fits ASDF 3 and then pivot to the rest of currently ASDF 3 changes into "ASDF 4".

This would properly version the API change here and let us fix our development patterns for a new major version.

braingram (approved):
Looks good to me.

WilliamJamieson (approved):

WilliamJamieson:
@braingram and @eslavich, based on the discussion in https://github.com/spacetelescope/stdatamodels/pull/110#issuecomment-1386080800 and PR eslavich/stdatamodels#1, I am going to convert this to a draft so it doesn't get merged until that discussion is resolved.

Feel free to close or mark this PR for review depending on the outcome of that discussion.

PR #1287, opened 2023-01-03
add custom validator support to extensions

braingram:
old style (legacy) extensions allowed for custom validators, functions called when a particular schema keyword was encountered (see CustomType.validators).

This PR introduces a similar feature for new style Extensions where custom validators can be defined in Extension.validators.

The scope of the validators is global (same as legacy validators).

fixes #1012

PR #1290, opened 2023-01-09
provisionally add block manager to serialization context

braingram:
adds AsdfProvisionalAPIWarning and issues this warning during use of block manager from the serialization context

Adding the block manager to the serialization context seemed like a simple and small change to be used to test the warning (and is something we want to the warning for).

Other new features that are candidates for using this warning are:
- #1287
- #1289


WilliamJamieson:
Marking as draft until #1295 is merged.

braingram:
Closing this until we have a plan for 3.0.

PR #1297, opened 2023-01-12
Use `--open-files`, `--run-slow`, and `--remote-data` flags for astropy downstream

### LINKED ISSUE #1244: Update CI to run slow downstream tests prior to release
To hopefully avoid post-release issues like https://github.com/asdf-format/asdf/issues/1239 where downstream package CI fails after release, what is a good way to add slow downstream tests to our CI?

Specifically, adding astropy tests with both `--open-files` and `--run-slow` args to pytest to replicate this tests might be a place to start:
https://github.com/astropy/astropy/blob/main/.github/workflows/ci_workflows.yml#L68

I am not familiar enough with github CI to know a good way to do this but perhaps there is a way to notice a *.*.x branch change and run the slower tests as these branches are created/changed prior to a release.

WilliamJamieson:
Fixes #1244, by adding the appropriate flags to the `tox` environment for `astropy` downstream testing

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -136,7 +136,7 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest astropy/astropy/io/misc/asdf
+    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data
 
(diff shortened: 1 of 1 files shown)


braingram (approved):
Looks good to me.

PR #1293, opened 2023-01-10
Increase `ruff` usage by ASDF

WilliamJamieson:
This PR enables `ruff` to use all its builtin checks, not just its default checks. This will overlap with some of the other `pre-commit` hooks. However, 100% coverage of the current tools we are using is not guaranteed, so I am leaving those out for now.

Note that some of the builtin checks have been manually turned off, but by default any new checks added to `ruff` will be applied to ASDF when the `pre-commit` gets updated. This is so we can recognize and decide if those checks should be included/fixed as they occur in a semi-automated fashion.

Not all the fixes are entirely automated, so I suggest we review this for style closely.

PR #1299, opened 2023-01-12
Add `pyupgrade` check to `ruff`

WilliamJamieson:
This PR is part of breaking #1293 into multiple parts.

It enables `ruff` to explicitly check:

- `Pyflakes` (a `ruff` default)
- `pycodestyle` (also a `ruff default)
- `isort` (we were running `isort` as a standalone tool)
- `pyupgrade` (we were running parts of `pyupgrade` standalone, this runs all those checks and found some extra things to change)

### DIFF
--- a/asdf/tests/test_yaml.py
+++ b/asdf/tests/test_yaml.py
@@ -122,8 +122,12 @@ def test_named_tuple_typing(tmp_path):
     Ensure that we are able to serialize a typing.NamedTuple.
     """
 
-    nt = NamedTuple("TestNamedTuple2", (("one", int), ("two", int), ("three", int)))
-    tree = {"val": nt(1, 2, 3)}
+    class NT(NamedTuple):
+        one: int
+        two: int
+        three: int
+
+    tree = {"val": NT(1, 2, 3)}
 
     run_tuple_test(tree, tmp_path)
 
@@ -141,9 +145,12 @@ def test_named_tuple_collections_recursive(tmp_path):
 
 
 def test_named_tuple_typing_recursive(tmp_path):
-    nt = NamedTuple("TestNamedTuple4", (("one", int), ("two", int), ("three", np.ndarray)))
+    class NT(NamedTuple):
+        one: int
+        two: int
+        three: np.ndarray
 
-    tree = {"val": nt(1, 2, np.ones(3))}
+    tree = {"val": NT(1, 2, np.ones(3))}
 
     def check_asdf(asdf):
         assert (asdf.tree["val"][2] == np.ones(3)).all()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -170,6 +170,10 @@ extend_skip_glob = ["asdf/extern/*", ".eggs/*", ".tox/*"]
 
 [tool.ruff]
 line-length = 120
+select = [
+    "F", # Pyflakes
+    "E", "W", # pycodestyle
+]
 extend-exclude = ["asdf/extern/*", "docs/*"]
 
 [tool.codespell]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,6 +175,8 @@ select = [
     "E", "W", # pycodestyle
     # "C90", # mccabe complexity
     "I", # isort
+    # "D", # pydocstyle
+    "UP", # pyupgrade
 ]
 extend-exclude = ["asdf/extern/*", "docs/*"]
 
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,3 +1,6 @@
+ci:
+  autofix_prs: false
```
braingram, line 2: Will this stop the bot from automatically running pre-commit and fixing failures?

WilliamJamieson: It will stop it from doing so automatically. Instead we will have to make a comment on its own with the magic phrase `pre-commit.ci autofix`

WilliamJamieson: This basically prevents pre-commit.ci from creating a merge conflict if you haven't pushed all your changes to the PR yet. And lets you undo changes which might be breaking the "style" to test with our full CI.

braingram: That feature has caught and fixed a number of issues. Why does this PR disable it?

braingram: I'm against turning this off.

WilliamJamieson: I think you misunderstand. This won't disable anything having to do with the pre-commit.ci bot except automatically pushing "autofix" commits to open PRs. That is pre-commit locally will act exactly the same and the CI test will run exactly the same. The only difference is that if the CI test fails and an autofix is available, it will not automatically push a commit to the PR's branch.

Instead, it leaves it up to the reviewer or submitter to comment `pre-commit.ci autofix` in a single comment on the PR. This will then trigger the pre-commit.ci bot to re-run its CI and push any autofix changes to the PR's branch.

WilliamJamieson: This is how `astropy` does it, and I vastly prefer it this way now.

braingram: So turning this off would disable automatic commits like [2a819ff](https://github.com/asdf-format/asdf/pull/1143/commits/2a819ff09298a9dc5cb21a4d1aeb69f2c9034743) in this PR: https://github.com/asdf-format/asdf/pull/1143

I am against turning off that feature.

WilliamJamieson: > So turning this off would disable automatic commits like [2a819ff](https://github.com/asdf-
format/asdf/pull/1143/commits/2a819ff09298a9dc5cb21a4d1aeb69f2c9034743) in this PR: #1143

Correct. However, instead the CI (which is required for to merge the PR) will just fail. Instead humans involved can then manually trigger the bot to make the changes.

> I am against turning off that feature.

Why? We cannot merge any PRs which are failing, and its simple for anyone to manually trigger the bot to push an autofix.

There are two big reasons to turn of this behavior:

1. If the PR submitter creates a PR and continues to make local changes, but the bot makes a autofix commit then the submitter has to deal with a merge conflict. This is extremely hostile to new and/or outside contributors and is the main reason to avoid this behavior.
2. Not all auto-fixable changes are necessarily "safe" meaning they can lead to failures (often in configurations outside the submitter's normal development environment). In these cases, it is reasonable to revert autofixes and re-run the main CI in order to establish that a `# noqa: ...`  might be necessary. In these cases one then runs into conflict with the bot (yes one can work around this but its annoying).

braingram: I'm against turning it off because it allows PRs to be fixed without the submitter needing to know the magic comment, run pre-commit locally or manually edit the code to apply the fixes.

I don't understand why this change is included in this PR. Is it required for the other changes?

WilliamJamieson: > I'm against turning it off because it allows PRs to be fixed without the submitter needing to know the magic comment, run pre-commit locally or manually edit the code to apply the fixes.

Yes but fixing merge conflicts is something that is very hostile to most new users. We as reviewers know the magic to comment autofix if it is needed. Reviewing/fixing most of the things under the `pre-commit` checks is pretty trivial and they are intended to make the code-base as a whole more readable.

> I don't understand why this change is included in this PR. Is it required for the other changes?

In creating #1293 I ran into some of the issues I enumerated above when I was debugging a few small issues. I figured it was a trivial thing to add as part of this process. Clearly that is not the case and I will revert this change and make it in a separate PR later for more discussion.

braingram: Thanks for moving that change to a new PR.

I doubt I will have time to finish reviewing this PR (or start on the others related to #1293) until next week at the earliest so there is no rush from my end.

WilliamJamieson: @braingram I'll open the autofix debate after we finish all these `ruff` changes. I have already stripped this change from all the `ruff` PRs so feel free to start reviewing them

braingram (approved):

PR #1300, opened 2023-01-12
Add `pep8-naming` check to `ruff`

WilliamJamieson:
This PR is part of breaking #1293 into multiple parts and is built off #1299.

It enables `ruff` to explicitly check `pep8-naming` style checks.

### DIFF
--- a/asdf/extension/_compressor.py
+++ b/asdf/extension/_compressor.py
@@ -22,9 +22,9 @@ class Compressor(abc.ABC):
     """
 
     @classmethod
-    def __subclasshook__(cls, C):
+    def __subclasshook__(cls, class_):
         if cls is Compressor:
-            return hasattr(C, "label") and (hasattr(C, "compress") or hasattr(C, "decompress"))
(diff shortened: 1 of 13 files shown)


asdf/extension/_compressor.py
```diff
@@ -22,9 +22,9 @@ class Compressor(abc.ABC):
     """
 
     @classmethod
-    def __subclasshook__(cls, C):
+    def __subclasshook__(cls, c):
```
WilliamJamieson, line 25:
```suggestion
    def __subclasshook__(cls, class_):
```

Is this better?

asdf/schema.py
```diff
@@ -263,20 +263,20 @@ def _create_validator(validators=YAML_VALIDATORS, visit_repeat_nodes=False):
         }
     )
     id_of = mvalidators.Draft4Validator.ID_OF
-    ASDFValidator = mvalidators.create(
+    asdf_validator = mvalidators.create(
```
braingram, line 266: Keeping this as upper case is more readable.

braingram (approved):

PR #1301, opened 2023-01-12
Add `flake8-bandit` check to `ruff`

WilliamJamieson:
This PR is part of breaking #1293 into multiple parts and is built off #1300.

It enables `ruff` to explicitly check:

- `flake8-2020` (no changes where necessary)
- `flake8-bandit` (no changes except explicit `# noqa` where needed), this is similar to the `bandit` checks but is not feature complete with respect to `bandit` yet.


WilliamJamieson:
Remove bandit completely

braingram (approved):

PR #1311, opened 2023-01-13
Add `flake8-return` check to `ruff`

WilliamJamieson:
This PR is part of breaking #1293 into multiple parts and is built off #1310.

It enables `ruff` to explicitly check `flake8-return` style checks.


WilliamJamieson:
@braingram despite https://github.com/asdf-format/asdf/pull/1293#issuecomment-1385515878, I would like you to specifically review the changes in this PR as they are quite significant.

braingram (approved):

PR #1323, opened 2023-01-18
Document the ASDF release cycle and supported dependencies

WilliamJamieson:
This PR proposes a release cycle and dependency support policy for ASDF. It is based off of [APE 2](https://github.com/astropy/astropy-APEs/blob/main/APE2.rst), [APE 18](https://github.com/astropy/astropy-APEs/blob/main/APE18.rst), and [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

Please make any comments concerning the details of these policies here.

PR #1324, opened 2023-01-20
Revert #1280

### LINKED ISSUE #1298: Flakey test
The test: https://github.com/asdf-format/asdf/blob/a2d2c4723ab8a8cdc0eb3f87fb913a80d728e5b9/asdf/tests/test_array_blocks.py#L834-L851

introduced in #1280 seems to be intermittently flaky for reasons I cannot explain. I see failures in the CI

See:

- https://github.com/asdf-format/asdf/actions/runs/3903417026/jobs/6667758950
- https://github.com/asdf-format/asdf/actions/runs/3880159122/jobs/6618149855

But when simply re-running the CI:

- https://github.com/asdf-format/asdf/actions/runs/3903417026/jobs/6668675088
- https://github.com/asdf-format/asdf/actions/runs/3885577317/jobs/6629628167 (note that I did some debugging before making an identical re-run)

The CI passes. I don't think this is related to the changes PR #1293, though all the failures are in CI for those changes. I think the issue is more that #1293 happens to have run the CI many times.

braingram:
Fixes #1298

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,6 @@
 The ASDF Standard is at v1.6.0
 
 - Require numpy<1.25 for python 3.8 [#1327]
-- Discard cache of lazy-loaded block data when it is no longer referenced
-  by the tree. [#1280]
 - Add AsdfProvisionalAPIWarning to warn developers of new features that
   may undergo breaking changes but are likely to be included as stable
   features (without this warning) in a future version of ASDF [#1295]
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -1228,28 +1228,11 @@ class Block:
                 self._memmap_data()
                 if not self._memmapped:
                     self._fd.seek(self.data_offset)
-                    # The data variable is important here, otherwise the weakref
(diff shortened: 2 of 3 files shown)


WilliamJamieson:
Let's revert and open a new issue to track adding the feature in #1280. #1280 seemed reasonable, so I would like to figure out and document why this is flakey. It would be great if we could figure out a better solution.

braingram:
Thanks.

I'll open an issue to track re-adding #1280

The CI failure might be related to numpy development dropping 3.8 for the 1.25 release. I will open a separate asdf issue for that.

zacharyburnett (approved):
LGTM

WilliamJamieson:
> The CI failure might be related to numpy development dropping 3.8 for the 1.25 release. I will open a separate asdf issue for that.

Good to know. It will be nice to drop 3.8 support. Maybe we should ahead of the LTS version for ASDF 2 discussed in #1323.

WilliamJamieson (approved):
LGTM.

We should drop dev support and pin numpy on Python 3.8. (Separate issue)

PR #1327, opened 2023-01-20
require numpy < 1.25 for python 3.8

### LINKED ISSUE #1326: Python 3.8 devdeps CI failing
This may be related to numpy development on version 1.25 which drops support for python 3.8.

braingram:
Numpy 1.25 is currently being developed and is dropping support for python 3.8.

Fixes #1326

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ dependencies = [
   "jmespath>=0.6.2",
   "jsonschema>=4.0.1",
   "numpy>=1.18",
+  'numpy<1.25,>=1.18; python_version < "3.9"',
   "packaging>=16",
   "pyyaml>=3.10",
   "semantic_version>=2.8",
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,4 +10,4 @@ git+https://github.com/yaml/pyyaml.git
 git+https://github.com/python-jsonschema/jsonschema
 
 scipy>=0.0.dev0
-numpy>=0.0.dev0
+numpy>=0.0.dev0 ; python_version > "3.8"
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@
 
 The ASDF Standard is at v1.6.0
 
+- Require numpy<1.25 for python 3.8 [#1327]
 - Discard cache of lazy-loaded block data when it is no longer referenced
   by the tree. [#1280]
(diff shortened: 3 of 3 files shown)


zacharyburnett (approved):
Will this need to be maintained manually, or is it possible to set a metapackage like `oldest-supported-numpy`?

requirements-dev.txt
```diff
@@ -10,4 +10,4 @@ git+https://github.com/yaml/pyyaml.git
 git+https://github.com/python-jsonschema/jsonschema
 
 scipy>=0.0.dev0
-numpy>=0.0.dev0
+numpy>=0.0.dev0 ; python_version > "3.8"
```
WilliamJamieson, line 13: Can you put an upper pin on numpy for Python 3.8 in the pyproject.toml

braingram: That's a great suggestion. However, I'd rather have a public change to the requirements in a separate PR with a corresponding changelog entry. If you'd rather have those here I'll make the changes.

braingram:
> oldest-supported-numpy

As far as my limited understanding of this extends, yes I think we'll have to manually maintain this to test against the nightly releases.

Perhaps there is some pip magic that would allow it to only install the oldest wheel including the dev version (numpy appears to have stopped making nightly wheels for 3.8 and our devdeps job is failing when attempting to build numpy).

requirements-dev.txt
WilliamJamieson: I think we can just retitle this PR and add a changelog. That way all the numpy/Python 3.8 stuff is all organized together.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
   "jmespath>=0.6.2",
   "jsonschema>=4.0.1",
   "numpy>=1.18",
+  'numpy<1.25,>=1.18; python_version < "3.9"',
```
WilliamJamieson, line 17: Oldest supported numpy version is 1.20 (and its support ends January 31). We should bump the minimum numpy soon.

braingram: Do you want that change included in this PR?

WilliamJamieson: No

braingram: What other changes are required to merge this PR? It's currently showing that you have 'changes requested'.

WilliamJamieson (approved):

PR #1328, opened 2023-01-21
Add new plugin type for custom schema validators

eslavich:
This resurrects #1050 so that we can merge it to master and include in a 2.x release.

### DIFF
--- a/asdf/core/_extensions.py
+++ b/asdf/core/_extensions.py
@@ -1,0 +1,23 @@
+from asdf.extension import ManifestExtension
+
+from ._validators import ndarray
+
+VALIDATORS = [
+    ndarray.NdimValidator(),
+    ndarray.MaxNdimValidator(),
+    ndarray.DatatypeValidator(),
+]
+
+
+MANIFEST_URIS = [
+    "asdf://asdf-format.org/core/manifests/core-1.0.0",
+    "asdf://asdf-format.org/core/manifests/core-1.1.0",
+    "asdf://asdf-format.org/core/manifests/core-1.2.0",
+    "asdf://asdf-format.org/core/manifests/core-1.3.0",
+    "asdf://asdf-format.org/core/manifests/core-1.4.0",
+    "asdf://asdf-format.org/core/manifests/core-1.5.0",
+    "asdf://asdf-format.org/core/manifests/core-1.6.0",
+]
+
+
+EXTENSIONS = [ManifestExtension.from_uri(u, validators=VALIDATORS) for u in MANIFEST_URIS]
--- a/asdf/core/_integration.py
+++ b/asdf/core/_integration.py
@@ -1,0 +1,21 @@
+from asdf.resource import JsonschemaResourceMapping
+
+
+def get_extensions():
+    """
+    Get the extension instances for the core extensions.  This method is registered with the
+    asdf.extensions entry point.
+
+    Returns
+    -------
+    list of asdf.extension.Extension
+    """
+    from . import _extensions
+
+    return _extensions.EXTENSIONS
+
+
+def get_json_schema_resource_mappings():
+    return [
+        JsonschemaResourceMapping(),
+    ]
--- a/asdf/core/_validators/ndarray.py
+++ b/asdf/core/_validators/ndarray.py
@@ -1,0 +1,28 @@
+from asdf.extension import Validator
+from asdf.tags.core.ndarray import validate_datatype, validate_max_ndim, validate_ndim
+
+
+class NdimValidator(Validator):
+    schema_property = "ndim"
+    # The validators in this module should really only be applied
+    # to ndarray-* tags, but that will have to be a 3.0 change.
+    tags = ["**"]
+
+    def validate(self, expected_ndim, node, schema):
+        yield from validate_ndim(None, expected_ndim, node, schema)
+
+
+class MaxNdimValidator(Validator):
+    schema_property = "max_ndim"
+    tags = ["**"]
+
+    def validate(self, max_ndim, node, schema):
+        yield from validate_max_ndim(None, max_ndim, node, schema)
+
+
+class DatatypeValidator(Validator):
+    schema_property = "datatype"
+    tags = ["**"]
+
+    def validate(self, expected_datatype, node, schema):
+        yield from validate_datatype(None, expected_datatype, node, schema)
--- a/asdf/extension/_manager.py
+++ b/asdf/extension/_manager.py
@@ -249,7 +249,7 @@ class ValidatorManager:
         self._jsonschema_validators_by_schema_property = {}
         for schema_property in self._validators_by_schema_property:
             self._jsonschema_validators_by_schema_property[schema_property] = self._get_jsonschema_validator(
-                schema_property
+                schema_property,
             )
 
     def validate(self, schema_property, schema_property_value, node, schema):
--- a/asdf/extension/_manager.py
+++ b/asdf/extension/_manager.py
@@ -52,7 +52,7 @@ class ExtensionManager:
 
             validators.update(extension.validators)
 
-        self._validator_manager = ValidatorManager(validators)
+        self._validator_manager = _get_cached_validator_manager(tuple(validators))
 
     @property
     def extensions(self):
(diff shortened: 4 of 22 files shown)


CagtayFabry:
I will try to run this against the previous test branch again soon @eslavich

WilliamJamieson:
@eslavich is it possible to break this into smaller chunks?

asdf/core/_extensions.py
```diff
@@ -0,0 +1,23 @@
+from asdf.extension import ManifestExtension
```
braingram, line 1:
```suggestion
from ..extension import ManifestExtension
```

asdf/core/_integration.py
```diff
@@ -0,0 +1,21 @@
+from asdf.resource import JsonschemaResourceMapping
```
braingram, line 1:
```suggestion
from ..resource import JsonschemaResourceMapping
```

asdf/core/_validators/ndarray.py
```diff
@@ -0,0 +1,28 @@
+from asdf.extension import Validator
+from asdf.tags.core.ndarray import validate_datatype, validate_max_ndim, validate_ndim
```
braingram, line 2:
```suggestion
from ...extension import Validator
from ...tags.core.ndarray import validate_datatype, validate_max_ndim, validate_ndim
```

asdf/extension/_manager.py
```diff
@@ -1,6 +1,7 @@
 from functools import lru_cache
 
-from asdf.util import get_class_name
+from asdf.tagged import Tagged
+from asdf.util import get_class_name, uri_match
```
braingram, line 4:
```suggestion
from ..tagged import Tagged
from ..util import get_class_name, uri_match
```

asdf/extension/_manager.py
```diff
@@ -214,3 +225,80 @@ def get_cached_extension_manager(extensions):
...
+
+        return _validator
+
+
+def _validator_matches(validator, node):
+    if any(t == "**" for t in validator.tags):
+        return True
+
+    if not isinstance(node, Tagged):
+        return False
```
braingram, line 302: We should add a test to cover these cases since the numpy validators no longer test them.

asdf/extension/_manager.py
```diff
@@ -214,3 +225,80 @@ def get_cached_extension_manager(extensions):
 @lru_cache
 def _get_cached_extension_manager(extensions):
     return ExtensionManager(extensions)
+
+
+class ValidatorManager:
+    """
+    Wraps a list of validators and indexes them by schema property.
```
braingram, line 232:
```suggestion
    Wraps a list of custom validators and indexes them by schema property.
```

asdf/extension/_manager.py
```diff
@@ -214,3 +225,80 @@ def get_cached_extension_manager(extensions):
...
+
+        self._validators_by_schema_property = {}
+        for validator in self._validators:
+            if validator.schema_property not in self._validators_by_schema_property:
+                self._validators_by_schema_property[validator.schema_property] = set()
+            self._validators_by_schema_property[validator.schema_property].add(validator)
+
+    def validate(self, schema_property, schema_property_value, node, schema):
+        """
+        Validate an ASDF tree node against a schema property.
```
braingram, line 251:
```suggestion
        Validate an ASDF tree node against custom validators for a schema property.
```

asdf/extension/_manager.py
```diff
@@ -214,3 +225,80 @@ def get_cached_extension_manager(extensions):
...
+        result = {}
+
+        for schema_property in self._validators_by_schema_property:
+            result[schema_property] = self._get_jsonschema_validator(schema_property)
+
+        return result
+
+    def _get_jsonschema_validator(self, schema_property):
+        def _validator(_, schema_property_value, node, schema):
+            return self.validate(schema_property, schema_property_value, node, schema)
```
braingram, line 292: Jsonschema appears to cache validators. The closure here captures self (ValidationManager) which I think is only holding references to the custom validators. It's probably safe to leave this as is but if we're concerned about collection of these objects we might be able to convert self to a weakref. Are there references that might warrant this change?

docs/asdf/extending/extensions.rst
```diff
@@ -154,6 +154,26 @@ Tag handles can be defined in the ``yaml_tag_handles`` property of an extension:
...
+
+    class FooValidator(Validator):
+        # ...
+        pass
+
+
+    class FooExtension(Extension):
+        extension_uri = "asdf://example.com/example-project/extensions/foo-1.0.0"
+        validators = [FooValidator()]
+
```
braingram, line 176: Should we link to the extending validators or other documentation here in case the reader wants to know more?

docs/asdf/extending/extensions.rst
```diff
@@ -154,6 +154,26 @@ Tag handles can be defined in the ``yaml_tag_handles`` property of an extension:
...
+and are included in an extension via the ``validators`` property:
+
+.. code-block:: python
+
+    from asdf.extension import Extension, Validator
+
+
+    class FooValidator(Validator):
+        # ...
+        pass
```
braingram, line 170:
```suggestion
    class FooValidator(Validator):
        schema_property = "bar"
        tags = ["asdf://example.com/example-project/tags/foo-1.0.0"]
        
        def validate(self, value, node, schema):
            if value != "baz":
                yield ValidationError("bar is not baz for foo object")
```
Should we flush this out to a complete example or link to the other example in the docs?

braingram (changes requested):
Overall this looks great.

I made a few minor documentation suggestions, changed some imports to relative and highlighted one test coverage gap (because of the change in ndarray validator scope).

asdf/extension/_manager.py
braingram: I opened a PR on your branch that should add coverage to these lines:
https://github.com/eslavich/asdf/pull/3

CagtayFabry:
> @braingram I think I've addressed all of your comments. The weldx failures are due to them using the old-style validators in their new-style extension:
> 
> https://github.com/BAMWelDX/weldx/blob/master/weldx/asdf/extension.py#L62-L66

Would it help to remove the code you linked from the new style extension to see if the legacy implementation still works @eslavich ?

CagtayFabry:
I have removed the incompatible validators from the new style extension, can you run the weldx downstream test again @eslavich ?

braingram (approved):

asdf/extension/_manager.py
braingram: Excellent! CacheInfo is super useful. I ran the suit and logged CacheInfo for all the lru_cache uses I could find (except for the legacy stuff). Here's the output (run on your source branch for this PR):
```bash
INFO     root:conftest.py:25 _get_cached_extension_manager: CacheInfo(hits=1356, misses=365, maxsize=128, currsize=128)
INFO     root:conftest.py:25 _get_cached_validator_manager: CacheInfo(hits=379, misses=4, maxsize=128, currsize=4)
INFO     root:conftest.py:25 _create_validator: CacheInfo(hits=3100, misses=10, maxsize=128, currsize=10)
INFO     root:conftest.py:25 _load_schema: CacheInfo(hits=0, misses=0, maxsize=128, currsize=0)
INFO     root:conftest.py:25 load_custom_schema: CacheInfo(hits=0, misses=0, maxsize=128, currsize=0)
INFO     root:conftest.py:25 _load_schema_cached: CacheInfo(hits=0, misses=0, maxsize=128, currsize=0)
INFO     root:conftest.py:25 _compile_uri_match_pattern: CacheInfo(hits=15555414, misses=148, maxsize=1024, currsize=148)
```
I'm not sure if the number of misses for `_get_cached_extension_manager` is reasonable given that the tests fiddle with extensions but that's outside the scope of this PR. If it looks suspicious feel free to open an issue for it (or let me know and I can open one).

docs/asdf/extending/extensions.rst
braingram: Thanks!

braingram:
> @eslavich is it possible to break this into smaller chunks?

@WilliamJamieson do you have a proposal for breaking this up or are you happy with it as is?

If you're happy with it I will rebase and merge this PR.

PR #1329, opened 2023-01-23
Update ruff to v0.0.231

WilliamJamieson:
We are currently using `ruff v0.0.223` and `ruff v0.0.230` was recently released. It updates and adds new checks. This PR updates ASDF to conform with these ahead of the pre-commit.ci bot attempting to update `ruff`.

### DIFF
--- a/asdf/fits_embed.py
+++ b/asdf/fits_embed.py
@@ -87,7 +87,7 @@ class _EmbeddedBlockManager(block.BlockManager):
                     return f"{FITS_SOURCE_PREFIX}{hdu.name},{hdu.ver}"
 
             msg = "FITS block seems to have been removed"
-            raise ValueError(msg)
+            raise ValueError(msg)  # noqa: TRY004
 
         return super().get_source(block)
 
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -94,7 +94,7 @@ def relative_uri(source, target):
     if relative == ".":
         relative = ""
 
-    return patched_urllib_parse.urlunparse(["", "", relative] + extra)
+    return patched_urllib_parse.urlunparse(["", "", relative, *extra])
 
 
 class _TruncatedReader:
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -608,12 +608,12 @@ class GenericFile(metaclass=util.InheritDocstrings):
         try:
             while reader.read(self.block_size) != b"":
                 pass
-            return True
         except DelimiterNotFoundError as e:
             if exception:
                 raise e
-
             return False
+
+        return True
 
     def fast_forward(self, size):
         """
@@ -986,10 +986,12 @@ def _http_to_temp(init, mode, uri=None):
                 fd.write(chunk)
                 chunk = response.read(block_size)
         fd.seek(0)
-        return RealFile(fd, mode, close=True, uri=uri or init)
+
     except Exception:
         fd.close()
         raise
+
+    return RealFile(fd, mode, close=True, uri=uri or init)
 
 
 def get_uri(file_obj):
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -608,9 +608,10 @@ class GenericFile(metaclass=util.InheritDocstrings):
         try:
             while reader.read(self.block_size) != b"":
                 pass
-        except DelimiterNotFoundError as e:
+        except DelimiterNotFoundError:
             if exception:
-                raise e
+                raise
+
             return False
 
         return True
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -1134,7 +1134,7 @@ def get_file(init, mode="r", uri=None, close=False):
             return InputStream(init, mode, uri=uri, close=close)
 
         msg = f"File '{init}' could not be opened in 'rw' mode"
-        raise ValueError(msg)
+        raise ValueError(msg)  # noqa: TRY004
 
     if mode == "w" and (hasattr(init, "write") and hasattr(init, "seek") and hasattr(init, "tell")):
         return MemoryIO(init, mode, uri=uri)
--- a/asdf/search.py
+++ b/asdf/search.py
@@ -203,7 +203,7 @@ class AsdfSearchResult:
         return AsdfSearchResult(
             self._identifiers,
             self._node,
-            filters=self._filters + [_filter],
+            filters=[*self._filters, _filter],
             parent_node=self._parent_node,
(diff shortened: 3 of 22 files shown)


braingram:
Any idea what's causing the roman_datamodels failure? I don't see a similar failure in a recent run: https://github.com/asdf-format/asdf/actions/runs/3983859659

WilliamJamieson:
> Any idea what's causing the roman_datamodels failure? I don't see a similar failure in a recent run: [asdf-format/asdf/actions/runs/3983859659](https://github.com/asdf-format/asdf/actions/runs/3983859659)

The failure is unrelated, it fails on main, see: https://github.com/asdf-format/asdf/actions/runs/3997794871/jobs/6859591670 which I just run against main.

WilliamJamieson:
`roman_datamodels` will have failures if new schemas are added to `rad` but do not include their support code yet. Looks like spacetelescope/rad#203 added a new schema yesterday (Jan 23, 2023) and `roman_datamodels` depends on the dev version of `rad`, see https://github.com/spacetelescope/roman_datamodels/blob/3613c4edada8e01350477dd762d44a14dac8d003/pyproject.toml#L23. So our CI is being broken by at least this.

WilliamJamieson:
I just ran the `roman_datamodels` CI on its `main` branch. It is also failing now:
https://github.com/spacetelescope/roman_datamodels/actions/runs/3997859004

asdf/generic_io.py
```diff
@@ -608,13 +608,14 @@ def seek_until(
...
             while reader.read(self.block_size) != b"":
                 pass
-            return True
         except DelimiterNotFoundError as e:
             if exception:
                 raise e
-
             return False
 
+        else:
```
braingram, line 616: Is this else necessary?

asdf/generic_io.py
```diff
@@ -986,11 +987,14 @@ def _http_to_temp(init, mode, uri=None):
                 fd.write(chunk)
                 chunk = response.read(block_size)
         fd.seek(0)
-        return RealFile(fd, mode, close=True, uri=uri or init)
+
     except Exception:
         fd.close()
         raise
 
+    else:
```
braingram, line 995: Is this else necessary?

asdf/search.py
```diff
@@ -97,12 +97,14 @@ def _maybe_compile_pattern(self, query):
...
         try:
             result = a == b
-            if isinstance(result, bool):
-                return result
 
+        except Exception:  # noqa: BLE001
             return False
 
-        except Exception:  # noqa: BLE001
+        else:
```
braingram, line 104: Is this else necessary?

asdf/fits_embed.py
```diff
@@ -87,7 +87,7 @@ def get_source(self, block):
                     return f"{FITS_SOURCE_PREFIX}{hdu.name},{hdu.ver}"
 
             msg = "FITS block seems to have been removed"
-            raise ValueError(msg)
+            raise TypeError(msg)
```
braingram, line 90: What rule is requiring this change? This exception (and others changed in this PR) could reach and be caught be code using this library. Changing the type risks breaking their error handling code.

braingram (changes requested):
Do you expect that we will have style changes like this about every week as ruff updates rules?

asdf/generic_io.py
WilliamJamieson: Probably not

asdf/generic_io.py
WilliamJamieson: Probably not

asdf/search.py
WilliamJamieson: probably not

WilliamJamieson:
> Do you expect that we will have style changes like this about every week as ruff updates rules?

For the next bit as they have a high release cadence. I can slow this down by simply limiting how often our pre-commit gets updated.

asdf/fits_embed.py
WilliamJamieson: Its `TRY004` (ruff, `TC004` in original rule set). See https://github.com/guilatrova/tryceratops/blob/main/docs/violations/TC004.md.

I can revert all the `TRY004` changes and ignore them. I would like to keep it on to avoid this mistake in the future.

braingram: That rule does not appear to be correctly applied. For this instance `block` is the correct type but it's value is incorrect in that it refers to a hdu that is no longer present. Why should this be a `TypeError` and not a `ValueError`? The

WilliamJamieson: You are correct.

Unfortunately, It is a deterministic pattern checker/fixer so it just looks for an `isinstance` followed by `ValueError` and changes the error message there. Thanks for catching this. I decided to revert all these changes and suppress them instead. I think the check is fine to leave on as it will hint that one should look more closely at what one is doing.

Looking at this case, I don't think this should be an autofix.

braingram: Thanks. That all sounds good to me.

braingram:
> > Do you expect that we will have style changes like this about every week as ruff updates rules?
> 
> For the next bit as they have a high release cadence. I can slow this down by simply limiting how often our pre-commit gets updated.

I don't have any objection to slowing these down although it would be nice to keep timely updates from the other tools if that's possible.

braingram (approved):
Looks good to me. I added one comment about the `from None` but that shouldn't hold this up unless you think it's worth changing.

asdf/tags/core/ndarray.py
```diff
@@ -379,9 +379,10 @@ def __setitem__(self, *args):
         # originates from the call to __repr__ inside the traceback report.
         try:
             self._make_array().__setitem__(*args)
-        except Exception as e:
+        except Exception:
             self._array = None
-            raise e from None
+
+            raise
```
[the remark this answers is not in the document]
WilliamJamieson: I have no idea what the `from None` was for. Its what python assumes if you don't add the `from` statement.

WilliamJamieson:
> I don't have any objection to slowing these down although it would be nice to keep timely updates from the other tools if that's possible.

@braingram Looking at the `ruff` cadence, I think we should slow down adoption of new rules because the project basically releases everyday with additional rules. They are very receptive to tweaks to rules, see charliermarsh/ruff#2158 with regard to the `TypeError` discussion we had above. This effectively means new linting rules get added and released super quickly, so they are not always fully flushed out.

I think what I am going to do is keep the pre-commit cadence the same, but wait that some time (maybe a month) has passed before enabling new rules. This means that the checks will hopefully be more stable before I mess with trying to correct them. I will create a project here to track this work.

Issue #1335, opened 2023-01-27
Deprecate AsdfFile.blocks and other public access to the block manager


braingram:
This gives the appearance of a public/stable block manager api.

https://github.com/asdf-format/asdf/blob/ae2ec6a4a7c761faa68a47441c6c779382aed3ea/asdf/asdf.py#L656-L661

As we will likely make breaking changes to the block manager we should add a deprecation warning to this and other public access to the block manager.

PR #1336, opened 2023-01-30
add AsdfDeprecationWarning to AsdfFile.blocks

braingram:
Public use of the block manager will be removed in ASDF 3.0 to allow for major changes to it's API. It is currently used by the legacy extensions (also removed in ASDF 3.0). The new converters do not currently have block storage access.

To provide a sense for possible breaking changes to the block manager and block storage access, the current experimental branch that adds block storage access to converters passes all block manager calls through the serialization context:
https://github.com/asdf-format/asdf/commit/4f6215608431f9f9e563fda9dc79d22c87a930bf
Adding block storage access to the converter is scheduled for ASDF 3.0 and the API and implementation details are not yet finalized.

Fixes issue #1335

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -136,7 +136,7 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data
+    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data -W "ignore:The property AsdfFile.blocks has been deprecated:asdf.exceptions.AsdfDeprecationWarning:asdf.asdf:661"
 
 [testenv:asdf-astropy]
 changedir={envtmpdir}
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ The ASDF Standard is at v1.6.0
 - Add AsdfProvisionalAPIWarning to warn developers of new features that
   may undergo breaking changes but are likely to be included as stable
   features (without this warning) in a future version of ASDF [#1295]
+- Add AsdfDeprecationWarning to AsdfFile.blocks [#1336]
 
 2.14.3 (2022-12-15)
 -------------------
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -658,6 +658,12 @@ class AsdfFile:
         """
         Get the block manager associated with the `AsdfFile`.
         """
+        warnings.warn(
+            "The property AsdfFile.blocks has been deprecated and will be removed "
+            "in asdf-3.0. Public use of the block manager is strongly discouraged "
+            "as there is no stable API",
+            AsdfDeprecationWarning,
+        )
         return self._blocks
 
     def set_array_storage(self, arr, array_storage):
@@ -682,8 +688,8 @@ class AsdfFile:
 
             - ``inline``: Store the data as YAML inline in the tree.
         """
-        block = self.blocks[arr]
-        self.blocks.set_array_storage(block, array_storage)
+        block = self._blocks[arr]
+        self._blocks.set_array_storage(block, array_storage)
 
     def get_array_storage(self, arr):
         """
@@ -693,7 +699,7 @@ class AsdfFile:
         ----------
         arr : numpy.ndarray
         """
-        return self.blocks[arr].array_storage
+        return self._blocks[arr].array_storage
 
     def set_array_compression(self, arr, compression, **compression_kwargs):
         """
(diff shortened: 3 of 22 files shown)


braingram:
The error in asdf-astropy is interesting and may be a bug in asdf.
https://github.com/asdf-format/asdf/actions/runs/4046535979/jobs/6959444733

The tests triggered by the fits schema:
https://github.com/astropy/asdf-astropy/blob/main/asdf_astropy/resources/schemas/fits/fits-1.0.0.yaml
result in an attempt to serialize and deserialize an object with tag `tag:astropy.org:astropy/fits/fits-1.0.0`. This correctly uses `to_yaml_tree` and `from_yaml_tree` from the [converter in asdf-astropy](https://github.com/astropy/asdf-astropy/blob/3ab3c86ba84dc6030814ef3fac1c06c093ee1e36/asdf_astropy/converters/fits/fits.py#L74-L76). However, the astropy.io.misc.asdf uses the same tag for FitsType which defines a reserve_blocks (I'm not sure why it needs to reserve blocks). https://github.com/astropy/astropy/blob/070fc3ce0e5bfeeecc0582844dc8a968f254dd7a/astropy/io/misc/asdf/tags/fits/fits.py#L68-L72
This results in FitsType.reserve_blocks being called on an object that is then serialized with AstropyFitsConverter.

It seems to me that finding a converter should prevent the use of the type similar to what is done during serialization:
https://github.com/asdf-format/asdf/blob/da32b4037d012bd07f1d85d5d356332920aaa5ee/asdf/yamlutil.py#L256-L266
but no such check is done:
https://github.com/asdf-format/asdf/blob/da32b4037d012bd07f1d85d5d356332920aaa5ee/asdf/block.py#L558-L562

WilliamJamieson:
> However, the astropy.io.misc.asdf uses the same tag for FitsType which defines a reserve_blocks (I'm not sure why it needs to reserve blocks). [astropy/astropy@`070fc3c`/astropy/io/misc/asdf/tags/fits/fits.py#L68-L72](https://github.com/astropy/astropy/blob/070fc3ce0e5bfeeecc0582844dc8a968f254dd7a/astropy/io/misc/asdf/tags/fits/fits.py#L68-L72) This results in FitsType.reserve_blocks being called on an object that is then serialized with AstropyFitsConverter.

This is irrelevant as `astropy.io.misc.asdf` is deprecated, and we have already agreed with astropy that `astropy.io.misc.asdf` will not support ASDF 3.0. Thus we don't need to bend over backwards to support it. Astropy can be specifically updated to skip the ASDF related tests for ASDF < 3 and ignore the deprecation warnings for ASDF 2.15.

braingram:
> This is irrelevant as `astropy.io.misc.asdf` is deprecated, and we have already agreed with astropy that `astropy.io.misc.asdf` will not support ASDF 3.0. Thus we don't need to bend over backwards to support it. Astropy can be specifically updated to skip the ASDF related tests for ASDF < 3 and ignore the deprecation warnings for ASDF 2.15.

That all sounds good. I opened a PR to fix the issue as it looks like a bug in asdf where it shouldn't be calling reserve_blocks on an object that will be serialized with a converter.

I looked into ignoring the warnings for specific astropy tests and I think the following should suffice if we're happy with the format of the warnings:
https://github.com/astropy/astropy/commit/460f2bdce9e3f1e0ce4159b4421691beb47ae372

tox.ini
```diff
@@ -136,7 +136,7 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data
+    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data -W "ignore:The property AsdfFile.blocks has been deprecated*:asdf.exceptions.AsdfDeprecationWarning:asdf.asdf"
```
braingram, line 139: @WilliamJamieson Should we do this to keep our CI from failing until astropy is updated?

braingram: Nevermind, this worked locally so will require some testing to see why it's not working in the ci.

tox.ini
```diff
@@ -136,7 +136,7 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data
+    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data -W "ignore:The property AsdfFile.blocks has been deprecated:asdf.exceptions.AsdfDeprecationWarning:asdf.asdf:661"
```
braingram, line 139: @WilliamJamieson What are your thoughts on this as a way to keep our downstream CI from failing until astropy is updated to ignore this new deprecation?

WilliamJamieson: This is fine

WilliamJamieson (approved):
This might be out of scope, but as we add these deprecation warnings, we should create a distinct page in the docs which calls out all of them. We may want to include notes on timelines for removal (maybe link to mitigation strategies too).

PR #1339, opened 2023-01-30
More ruff clean-ups

WilliamJamieson:
Ruff has made a few tweaks (some of which I requested in charliermarsh/ruff#2202 and charliermarsh/ruff#2158) to rules we are already enforcing (in particular `SIM300` has been update to detect even more versions of yoda conditions which need to be ignored because we are specifically testing that we can handle those conditions).

This PR cleans up all unnecessary noqa and adds some new `SIM300` noqa. It also ignores:

- `TRY002`
- `TRY301`
- `TRY400`

Which we will address at a later date, per https://github.com/asdf-format/asdf/pull/1329#issuecomment-1405301731.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
       exclude: "asdf/extern/.*"
 
 - repo: https://github.com/charliermarsh/ruff-pre-commit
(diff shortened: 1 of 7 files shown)


braingram (approved):
Looks good to me.

PR #1344, opened 2023-01-30
don't call reserve_blocks on types that have converters

braingram:
This PR fixes reserve_blocks to treat objects in the tree (nodes) in a way consistent with
yamlutil.custom_tree_to_tagged_tree. Conversion of nodes in a custom tree will preferentially use Converters if a type/tag is supported by both old style types and new style converters. Reserve blocks always assumed old style types. This resulted in serialization calling old style reserve_blocks and then new style converters for nodes supported by both types and converters.

### DIFF
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -556,10 +556,12 @@ class BlockManager:
         reserved_blocks = set()
 
         for node in treeutil.iter_tree(tree):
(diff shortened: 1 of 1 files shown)


braingram:
This (testing locally) fixes the failures in asdf-astropy tests in this PR which revealed the issue:
https://github.com/asdf-format/asdf/pull/1336

WilliamJamieson (approved):

PR #1349, opened 2023-02-06
add crds environment variables to stdatamodels ci

braingram:
This PR fixes the stdatamodels downstream CI by pre-syncing the required CRDS context.


braingram:
@zacharyburnett I'm hoping you can work your magic here :)

There are a few things I wasn't able to sort out that would improve this PR.

The crds cache was complicated by the reusable workflow and had to be run in a separate job. This creates a cache that only contains 440kB of data (likely only the mapping files):
https://github.com/asdf-format/asdf/actions/caches
this is unlike the >100 MB stdatamodels caches which likely contain reference files downloaded during the tests:
https://github.com/spacetelescope/stdatamodels/actions/caches
Do you see any easy way to have the cache be rewritten by the OpenAstronomy job or to integrate the crds sync with that job? Note, I'm not sure if this will be needed if the test changes in this PR are accepted: https://github.com/spacetelescope/stdatamodels/pull/117 However I'm not sure if the jwst tests would benefit.

The second issue is there are several places where the CRDS_PATH (etc) are defined. Avoiding this duplication would be preferred but I'm not experienced enough with actions or tox to know how to pass these through the reusable workflow, to tox or where it best makes sense for them to be defined.

Thanks in advance for your help (and the tip about using /tmp)!

zacharyburnett:
> The crds cache was complicated by the reusable workflow and had to be run in a separate job. This creates a cache that only contains 440kB of data (likely only the mapping files): https://github.com/asdf-format/asdf/actions/caches this is unlike the >100 MB stdatamodels caches which likely contain reference files downloaded during the tests: https://github.com/spacetelescope/stdatamodels/actions/caches Do you see any easy way to have the cache be rewritten by the OpenAstronomy job or to integrate the crds sync with that job? Note, I'm not sure if this will be needed if the test changes in this PR are accepted: [spacetelescope/stdatamodels#117](https://github.com/spacetelescope/stdatamodels/pull/117) However I'm not sure if the jwst tests would benefit (by not double fetching files if they're used in both stdatamodels and jwst tests).

@jdavies-st informed me that `crds sync` will not retrieve the correct files, so the cache will be incorrect. Unfortunately the best way to cache seems to be to let the tests run and then cache whatever CRDS files they pulled. Thus the `cache_crds` job is detrimental rather than beneficial

jdavies-st:
Yeah, you definitely don't want to run `crds sync`. In it's minimal form, it will just get the mappings, which are useless on their own, and maximally it will get the context snapshot, reffiles and all, which will take hours and gather up half a TB of data, and will of course timeout on github actions.

The thing to do is only do the caching on the jobs that use CRDS, which here is just `jwst`.  Meanwhile your hitting the cache for all 4 jobs: gwcs, jwst, stdatamodels, roman_datamodels, but the other 3 don't use anything from CRDS (or do so minimally?) and they complete first, i.e. long before the jwst job finishes, so they set the cache.  And it is empty.

Only cache for jwst, or if you want to cache for all of them since they are all in own job, have the cache key synced to the upstream package name, so that only jwst is setting the jwst cache.

braingram:
Thanks @zacharyburnett and @jdavies-st for your help untangling this.

Unfortunately stdatamodels now requires CRDS data for a few of it's tests.
https://github.com/spacetelescope/stdatamodels/blob/46e96286d7c23eaf7bed5f1baa1b441f470dacee/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py#L166

with the current code this requires a call to `crds sync` because it explicitly loads the context from cache:
https://github.com/spacetelescope/stdatamodels/blob/46e96286d7c23eaf7bed5f1baa1b441f470dacee/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py#L177

One solution I could see for this is to:
- finalize this PR to fix the stdatamodels tests to not require `crds sync`: https://github.com/spacetelescope/stdatamodels/pull/117
- remove the cache from this PR (and also from stdatamodels)

Proposals for other options to fix our CI are appreciated. Thanks!

jdavies-st:
It looks like the test itself will get the mappings?  If so, then make sure the github actions cache is hit for the stdatamodels job separately from the jwst job.  Is the test supposed to pull the current default context?  If, so then have the cache key based off the current default context and the job env for stdatamodels, so it gets its own unique cache.

First time it runs, no cache, but then after the test suite is run, that test will have pulled over the mappings for a particular context.  Then you cache at the end of the job all those files using a key based on the crds context (and the stdatamodels env).  Next time it runs with the same context, all those files are already there, and the test will use the cached versions of the mappings.

braingram:
> It looks like the test itself will get the mappings? If so, then make sure the github actions cache is hit for the stdatamodels job separately from the jwst job. Is the test supposed to pull the current default context? If, so then have the cache key based off the current default context and the job env for stdatamodels, so it gets its own unique cache.

Thanks for taking a look at this. I am no CRDS expert but the test uses `crds.get_cached_mapping` https://github.com/spacetelescope/stdatamodels/blob/46e96286d7c23eaf7bed5f1baa1b441f470dacee/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py#L177
This appears to only attempt to load the file from disk. I was never able to get this to fetch the context (tested off the vpn so it doesn't have access to pre-cached crds and with CRDS_PATH set).

> 
> First time it runs, no cache, but then after the test suite is run, that test will have pulled over the mappings for a particular context. Then you cache at the end of the job all those files using a key based on the crds context (and the stdatamodels env). Next time it runs with the same context, all those files are already there, and the test will use the cached versions of the mappings.

I think something like this might work (if the stdatamodels tests fetched all the necessary files) but it's complicated by the reusable workflow. As you pointed out only some of the tox environments require the cache and at the moment all of the stsci ones are hitting it (so any one of them has a chance to invalidate/update it).

WilliamJamieson (approved):

PR #1350, opened 2023-02-06
Enable CRDS for stdatamodels downstream

WilliamJamieson:
This should fix the downstream tests for stdatamodels.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -209,6 +209,11 @@ allowlist_externals=
     git
     bash
 extras=
+setenv=
+    CRDS_SERVER_URL = https://jwst-crds.stsci.edu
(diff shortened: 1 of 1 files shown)


zacharyburnett (approved):
LGTM

braingram (changes requested):
This fails to fix the issue:
https://github.com/asdf-format/asdf/actions/runs/4105988581/jobs/7083632898#step:9:844

zacharyburnett:
I have had success in using `/tmp/crds_cache` instead of `$HOME/crds_cache` in other workflows; perhaps we could set that in `ci.yml` and then use `pass_env: CRDS`

WilliamJamieson:
I think this is superseded by #1349

PR #1359, opened 2023-02-09
add deprecation to CustomType

braingram:
CustomType is deprecated. This PR adds a warning and test for the warning.

### DIFF
--- a/asdf/tags/core/tests/test_history.py
+++ b/asdf/tags/core/tests/test_history.py
@@ -7,7 +7,7 @@ from jsonschema import ValidationError
 
 import asdf
 from asdf import types, util
-from asdf.exceptions import AsdfWarning
+from asdf.exceptions import AsdfDeprecationWarning, AsdfWarning
 from asdf.tags.core import HistoryEntry
 from asdf.tests import helpers
 from asdf.tests.helpers import assert_no_warnings, yaml_to_asdf
@@ -201,20 +201,22 @@ history:
 
 
 def test_metadata_with_custom_extension(tmpdir):
-    class FractionType(types.CustomType):
-        name = "fraction"
-        organization = "nowhere.org"
-        version = (1, 0, 0)
-        standard = "custom"
-        types = [fractions.Fraction]
+    with pytest.warns(AsdfDeprecationWarning, match=".*subclasses the deprecated CustomType.*"):
 
-        @classmethod
-        def to_tree(cls, node, ctx):
-            return [node.numerator, node.denominator]
+        class FractionType(types.CustomType):
+            name = "fraction"
+            organization = "nowhere.org"
+            version = (1, 0, 0)
+            standard = "custom"
+            types = [fractions.Fraction]
 
-        @classmethod
-        def from_tree(cls, tree, ctx):
-            return fractions.Fraction(tree[0], tree[1])
+            @classmethod
+            def to_tree(cls, node, ctx):
+                return [node.numerator, node.denominator]
+
+            @classmethod
+            def from_tree(cls, tree, ctx):
+                return fractions.Fraction(tree[0], tree[1])
 
     class FractionExtension(CustomExtension):
         @property
--- a/asdf/tests/objects.py
+++ b/asdf/tests/objects.py
@@ -1,0 +1,40 @@
+from asdf import CustomType, util
+
+from .helpers import get_test_data_path
+
+
+class CustomTestType(CustomType):
+    """This class is intended to be inherited by custom types that are used
+    purely for the purposes of testing. The methods ``from_tree_tagged`` and
+    ``from_tree`` are implemented solely in order to avoid custom type
+    conversion warnings.
+    """
+
+    @classmethod
+    def from_tree_tagged(cls, tree, ctx):
+        return cls.from_tree(tree.data, ctx)
+
+    @classmethod
+    def from_tree(cls, tree, ctx):
+        return tree
+
+
+class CustomExtension:
+    """
+    This is the base class that is used for extensions for custom tag
+    classes that exist only for the purposes of testing.
+    """
+
+    @property
+    def types(self):
+        return []
+
+    @property
+    def tag_mapping(self):
+        return [("tag:nowhere.org:custom", "http://nowhere.org/schemas/custom{tag_suffix}")]
+
+    @property
+    def url_mapping(self):
+        return [
+            ("http://nowhere.org/schemas/custom/", util.filepath_to_url(get_test_data_path("")) + "/{url_suffix}.yaml"),
+        ]
--- a/asdf/tests/objects.py
+++ b/asdf/tests/objects.py
@@ -1,22 +1,26 @@
+import pytest
+
 from asdf import CustomType, util
+from asdf.exceptions import AsdfDeprecationWarning
 
 from .helpers import get_test_data_path
 
+with pytest.warns(AsdfDeprecationWarning, match=".*subclasses the deprecated CustomType.*"):
 
-class CustomTestType(CustomType):
-    """This class is intended to be inherited by custom types that are used
-    purely for the purposes of testing. The methods ``from_tree_tagged`` and
-    ``from_tree`` are implemented solely in order to avoid custom type
-    conversion warnings.
-    """
+    class CustomTestType(CustomType):
+        """This class is intended to be inherited by custom types that are used
+        purely for the purposes of testing. The methods ``from_tree_tagged`` and
+        ``from_tree`` are implemented solely in order to avoid custom type
+        conversion warnings.
+        """
 
-    @classmethod
-    def from_tree_tagged(cls, tree, ctx):
-        return cls.from_tree(tree.data, ctx)
+        @classmethod
+        def from_tree_tagged(cls, tree, ctx):
+            return cls.from_tree(tree.data, ctx)
 
-    @classmethod
-    def from_tree(cls, tree, ctx):
-        return tree
+        @classmethod
+        def from_tree(cls, tree, ctx):
+            return tree
 
 
 class CustomExtension:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,6 +112,7 @@ open_files_ignore = ['test.fits', 'asdf.fits']
 filterwarnings = [
     'error',
     'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes',
+    'ignore:.*subclasses the deprecated CustomType:asdf.exceptions.AsdfDeprecationWarning:asdf.types',
     'ignore:numpy.ndarray size changed:RuntimeWarning',
 ]
 # Configuration for pytest-doctestplus
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,8 +112,8 @@ open_files_ignore = ['test.fits', 'asdf.fits']
 filterwarnings = [
     'error',
     'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes',
-    'ignore:.*subclasses the deprecated CustomType:asdf.exceptions.AsdfDeprecationWarning:asdf.types',
     'ignore:numpy.ndarray size changed:RuntimeWarning',
+    'ignore:.*from astropy.io.misc.asdf.types subclasses the deprecated CustomType.*:asdf.exceptions.AsdfDeprecationWarning',
 ]
 # Configuration for pytest-doctestplus
 text_file_format = 'rst'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ filterwarnings = [
     'error',
     'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes',
     'ignore:numpy.ndarray size changed:RuntimeWarning',
-    'ignore:.*from astropy.io.misc.asdf.types subclasses the deprecated CustomType.*:asdf.exceptions.AsdfDeprecationWarning',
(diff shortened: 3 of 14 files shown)


braingram:
The asdf pytest plugin will import asdf.tests.helpers which will create CustomTestType which will trigger the warning. So every tested schema will emit a warning.
https://github.com/asdf-format/asdf/blob/master/pytest_asdf/plugin.py#L224

asdf-astropy downstream failing in part due to pulling in asdf.tests:
https://github.com/astropy/asdf-astropy/blob/main/asdf_astropy/converters/fits/tests/test_fits.py#L4

specutils is subclassing CustomType:
https://github.com/astropy/specutils/blob/main/specutils/io/asdf/types.py#L20
This is used for spectra tags:
https://github.com/astropy/specutils/blob/57dd2b28e5122e00e28333043192d30ec9f78a9d/specutils/io/asdf/tags/spectra.py#L28
I see no Converter replacement for these.

A cursory look at the sunpy and dkist logs show mostly warnings from dependencies (astropy.io.misc.asdf and from testing schema).

pyproject.toml
```diff
@@ -95,6 +95,7 @@ open_files_ignore = ['test.fits', 'asdf.fits']
 filterwarnings = [
     'error',
     'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes',
+    'ignore:.*subclasses the deprecated CustomType:asdf.exceptions.AsdfDeprecationWarning:asdf.types',
```
WilliamJamieson, line 98: How difficult would it be for us explicitly catch these warnings in our tests? This will mark the tests that need to be adjusted/removed when `CustomType` is fully removed.

WilliamJamieson (changes requested):

pyproject.toml
braingram: The issue I ran into trying per-test ignores is that warnings turn into errors during test collection and no tests get run. Is there a way to work around this?
One thought I had was to remove the `error` filter from pyproject.toml and add a pytest fixture that turned warnings into errors but this seemed more confusing than just adding the warning here.
Other PRs adding deprecations have similar issues so finding an elegant fix for this would be good.

WilliamJamieson: Adding this deprecation warning will require way more thought than what is being done here. The deprecation warning will be raised by ASDF every single time it tries to load entry points, which is unacceptable to me.

WilliamJamieson: Note that for most of the things raising the deprecation error on startup this can be solved by adding a fixture which loads and catches the warning.

braingram: > Adding this deprecation warning will require way more thought than what is being done here. The deprecation warning will be raised by ASDF every single time it tries to load entry points, which is unacceptable to me.

Perhaps I am not understanding what you are describing. I am only seeing the warning the first time the entry points are loaded:
```python
import warnings

import asdf.entry_points

with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    _ = asdf.entry_points.get_extensions()

_ = asdf.entry_points.get_extensions()
```
The above run with `python -X dev` shows no warnings for the second call to get_extensions. Would you provide an example of what behavior your are objecting to?

The deprecation warning is raised once per subclass of CustomType to inform the developer which classes require updating.

WilliamJamieson: I amend my above statement. The issue is when astropy's extension gets loaded, it issues a bunch of warnings these will all need to be ignored.

pyproject.toml
```diff
@@ -113,6 +113,7 @@ filterwarnings = [
     'error',
     'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes',
     'ignore:numpy.ndarray size changed:RuntimeWarning',
+    'ignore:.*from astropy.io.misc.asdf.types subclasses the deprecated CustomType.*:asdf.exceptions.AsdfDeprecationWarning',
```
WilliamJamieson, line 116: I think we should filter this specific warning when we load the entry points in `asdf.entry_points`. `astropy.io.misc.asdf` is deprecated and will be removed in `astropy` 6, meaning that `astropy` will not want to resolve this warning.

Since `astropy` includes an entry point which raises this warning, this warning will be raised each time ASDF is used while `astropy` is installed. Normally, I wouldn't care too much about doing this, but since our primary users all use ASDF and `astropy` together this effectively means everything in our downstream will need to suppress this specific warning (in many cases this will be turned into an error), which `astropy` will not be fixing. So we will be breaking our downstream packages' CI unnecessarily.

If we restrict the filter to just `astropy.io.misc.asdf` then this suppression will only effect `astropy` (which is won't be upgrading anyways).

tox.ini
```diff
@@ -133,7 +133,9 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data -W "ignore:The property AsdfFile.blocks has been deprecated:asdf.exceptions.AsdfDeprecationWarning:asdf.asdf:661"
+    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data \
+        -W "ignore:The property AsdfFile.blocks has been deprecated:asdf.exceptions.AsdfDeprecationWarning:asdf.asdf:661" \
```
WilliamJamieson, line 137: This was fixed by astropy/astropy#14358 so it is unnecessary.

tox.ini
```diff
@@ -133,7 +133,9 @@ commands_pre=
...
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:504"
```
WilliamJamieson, line 138: This will be fixed by the above request on the entry point.

tox.ini
```diff
@@ -148,7 +150,8 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest asdf-astropy
+    pytest asdf-astropy \
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:504"
```
WilliamJamieson, line 154: This will be fixed by the above request on the entry point.

tox.ini
```diff
@@ -163,7 +166,8 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest specutils
+    pytest specutils \
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:504"
```
WilliamJamieson, line 170: This will be fixed by the above request on the entry point.

tox.ini
```diff
@@ -263,7 +267,8 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest sunpy/sunpy/io
+    pytest sunpy/sunpy/io \
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:504"
```
WilliamJamieson, line 271: This will be fixed by the above request on the entry point.

tox.ini
```diff
@@ -278,4 +283,5 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest dkist
+    pytest dkist \
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:504"
```
WilliamJamieson, line 287: This will be fixed by the above request on the entry point.

WilliamJamieson (changes requested):
Looks good aside from some final comments about suppressing the warning when loading the entry-point.

tox.ini
braingram: Made this change.

tox.ini
braingram: This is still required because specutils contains a subclass of CustomType.

WilliamJamieson: Really we should not be suppressing warnings in our downstream. This should then be an expected failure.

asdf/tests/objects.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @property
+    def tag_mapping(self):
+        return [("tag:nowhere.org:custom", "http://nowhere.org/schemas/custom{tag_suffix}")]
+
+    @property
+    def url_mapping(self):
+        return [
+            ("http://nowhere.org/schemas/custom/", util.filepath_to_url(get_test_data_path("")) + "/{url_suffix}.yaml"),
+        ]
```
WilliamJamieson, line 44: Is there any reason these needed to be moved out of `asdf.tests.__init__`?

tox.ini
WilliamJamieson: I just opened astropy/specutils#1011 to track the expected failure in the downstream.

tox.ini
braingram: I removed this line.

tox.ini
braingram: This is still needed as the tests fail on collection without it. Here is one example:
```
plugins: remotedata-0.4.0, xdist-3.2.0, mock-3.10.0, filter-subpackage-0.1.2, astropy-header-0.2.2, doctestplus-0.12.1, hypothesis-6.68.1, astropy-0.10.0, asdf-2.14.4.dev225+gf580ac36.d20230215, cov-4.0.0, openfiles-0.5.0, arraydiff-0.5.0
collected 289 items / 25 errors

====================================================================================================================== ERRORS =======================================================================================================================
________________________________________________________________________________________________ ERROR collecting astropy/io/misc/asdf/extension.py _________________________________________________________________________________________________
astropy/astropy/io/misc/asdf/extension.py:12: in <module>
    from .tags.coordinates.angle import *
astropy/astropy/io/misc/asdf/tags/coordinates/angle.py:3: in <module>
    from astropy.io.misc.asdf.tags.unit.quantity import QuantityType
astropy/astropy/io/misc/asdf/tags/unit/quantity.py:4: in <module>
    from astropy.io.misc.asdf.types import AstropyAsdfType
astropy/astropy/io/misc/asdf/types.py:37: in <module>
    class AstropyType(CustomType, metaclass=AstropyTypeMeta):
astropy/astropy/io/misc/asdf/types.py:25: in __new__
    cls = super().__new__(mcls, name, bases, attrs)
../../../asdf/types.py:122: in __new__
    new_cls = super().__new__(cls, name, bases, attrs)
../../../asdf/types.py:504: in __init_subclass__
    warnings.warn(
E   asdf.exceptions.AsdfDeprecationWarning: AstropyType from astropy.io.misc.asdf.types subclasses the deprecated CustomType class. Please see the new extension API https://asdf.readthedocs.io/en/stable/asdf/extending/converters.html
```

tox.ini
braingram: I removed this line.

tox.ini
braingram: To clarify, do you want this line removed and the downstream specutils job to fail?

asdf/tests/objects.py
braingram: Import of `asdf.tests.helpers` like is done in asdf-astropy would result in warnings that created errors during test collection.
https://github.com/astropy/asdf-astropy/search?q=asdf.tests.helpers

tox.ini
WilliamJamieson: Yes. The same for the astropy job. We are expecting those jobs to fail.

asdf/tags/core/tests/test_history.py
```diff
@@ -201,20 +201,22 @@ def test_strict_extension_check():
...
-        @classmethod
-        def to_tree(cls, node, ctx):
-            return [node.numerator, node.denominator]
-
-        @classmethod
-        def from_tree(cls, tree, ctx):
-            return fractions.Fraction(tree[0], tree[1])
+    with pytest.warns(AsdfDeprecationWarning, match=".*subclasses the deprecated CustomType.*"):
+
+        class FractionType(types.CustomType):
```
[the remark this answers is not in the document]
braingram: Perhaps. I would prefer to limit the amount of refactoring that is done to our tests to limit the potential bugs introduced. If refactoring the tests to all subclass `CustomTestType` results in overall fewer changes than what is already in this PR than I would likely consider it a better option.

tox.ini
braingram: I do not think we should allow these to fail when we have a simple way to keep them running to find errors for the other changes we are making. If we do not include these warning filters they will fail during test collection and run no tests. This means that any change we make in asdf that might break these downstream packages would go unnoticed (instead of appearing in our CI with the addition of these warning filters).

WilliamJamieson: If you want to go this route then you should track the "fixes" in the downstream packages and when they are merged, removing them from ASDF.

asdf/tests/objects.py
WilliamJamieson: This doesn't make sense to me. You are literally wrapping the `CustomTestType` in something which catches the warning. All that would need to be done if this was left in `asdf.tests.__init__` would be to add this wrapper there instead.

asdf/tags/core/tests/test_history.py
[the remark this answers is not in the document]
WilliamJamieson: This would eliminate all the `pytest.warns` contexts from every test file. The only place that would be left is in `asdf.tests.objects` (or `asdf.tests.__init__` if it moves back there).

tox.ini
braingram: That sounds good to me.

asdf/tests/objects.py
braingram: The change was introduced prior to the refactoring that you requested adding `pytest.warns` for every use of `CustomType` in the tests.

braingram:
@WilliamJamieson I believe I have addressed all your requested changes. Are there any I missed?

WilliamJamieson (approved):
These are fine

PR #1360, opened 2023-02-10
Add oldest dependencies testing

WilliamJamieson:
Add testing against the oldest pinned versions of our dependencies.

Note that per https://github.com/asdf-format/asdf/pull/1331#discussion_r1096980876 I moved the script for generating the minimum dependencies out of ASDF and into its own package, see: https://github.com/WilliamJamieson/minimum_dependencies.

Closes #1331, as this supersedes it.


braingram (approved):
Looks good to me.

PR #1362, opened 2023-02-13
deprecate asdf.resolver submodule

braingram:
asdf.resolver is only used by legacy extensions.

This PR adds a deprecation warning to the submodule and creates a hidden `_resolver` submodule for internal use.


braingram:
Ugh, something odd is happening with the CI for this PR. The jobs failed to start, then the downstream CI label never took effect.

pyproject.toml
```diff
@@ -114,6 +114,7 @@ filterwarnings = [
     'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.asdftypes',
     'ignore:numpy.ndarray size changed:RuntimeWarning',
     'ignore:.*from astropy.io.misc.asdf.*subclasses the deprecated CustomType.*:asdf.exceptions.AsdfDeprecationWarning',
+    'ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.resolver:7'
```
braingram, line 117: @WilliamJamieson Can I get your thoughts on how to handle this deprecation? Based on the previous PR having this line is not preferred which makes the approach taken in this PR (move `resolver.py`  to `_resolver.py` and add a stub in `resolver.py` that issues a warning on direct use) unacceptable.

asdf/tests/test_resolver.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
+from asdf._resolver import Resolver, ResolverChain
```
WilliamJamieson, line 3: Can this be moved to a local import down in `assert_extension_correctness`. Eventually, that method will need to be deprecated too. I would rather not expose the "new" private API in this module.

WilliamJamieson (changes requested):
One small request.

asdf/tests/helpers.py
```diff
@@ -23,12 +23,12 @@
 
 import asdf
 from asdf import generic_io, versioning
+from asdf._resolver import Resolver, ResolverChain
```
WilliamJamieson, line 26: I meant for my comment to be here. Please move this to a local import in `assert_extension_correctness`. Since that is the only method that uses these in this module anyway, I would rather remove the public API all together from this.

Note `assert_estension_correctness` will need to be deprecated with the rest of the legacy extension stuff.

WilliamJamieson (approved):

asdf/tests/helpers.py
braingram: Thanks for this catch. I made the change.

I'll open another PR for the `assert_extension_correctness` deprecation.

WilliamJamieson:

PR #1365, opened 2023-02-15
Make the warning only issue on the very first subclass of `CustomType`

WilliamJamieson:
This makes the warning make sense, by only issuing it on immediate children of `CustomType`. It will make astropy/astropy#14395 much simpler.

### DIFF
--- a/asdf/types.py
+++ b/asdf/types.py
@@ -44,21 +44,6 @@ def _from_tree_tagged_missing_requirements(cls, tree, ctx):
     raise TypeError(msg)
 
 
-def _find_first_child_class(sub_class, parent_class):
-    err = None
-    for b in sub_class.__bases__:
-        if b == parent_class:
-            return sub_class
-        try:
-            return _find_first_child_class(b, parent_class)
-        except ValueError as e:
-            err = e
-    if err:
-        raise err
-    msg = f"Failed to find {parent_class} in {sub_class}"
-    raise ValueError(msg)
-
-
 class ExtensionTypeMeta(type):
     """
     Custom class constructor for tag types.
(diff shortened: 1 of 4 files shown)


asdf/types.py
```diff
@@ -500,10 +500,13 @@ class CustomType(ExtensionType, metaclass=ExtensionTypeMeta):
...
+
+        # Create a warning for the first subclass of a CustomType class
         scls = _find_first_child_class(cls, CustomType)
-        warnings.warn(
-            f"{scls.__name__} from {scls.__module__} subclasses the deprecated CustomType class. "
-            "Please see the new extension API "
-            "https://asdf.readthedocs.io/en/stable/asdf/extending/converters.html",
-            AsdfDeprecationWarning,
-        )
+        if scls.__name__ == cls.__name__:
```
braingram, line 506:
```suggestion
        if CustomType in cls.__bases__:
```
Can you give this a try (which would allow for removal of `_find_first_child_class`)? I believe it accomplishes the same result.

WilliamJamieson: Local check showed this worked. I'll push a fix removing `_find_first_child_class` completely.

braingram:
2 of the downstream CI jobs are failing with this PR.

WilliamJamieson:
Those should be expected.

WilliamJamieson:
The tox.ini suppression is very very specific (down to the line #) that has changed. This is part of the reason for my objections here: https://github.com/asdf-format/asdf/pull/1359#discussion_r1106447876. I will fix it because that decision has been made.

braingram:
> The tox.ini suppression is very very specific (down to the line #) that has changed. This is part of the reason for my objections here: [#1359 (comment)](https://github.com/asdf-format/asdf/pull/1359#discussion_r1106447876). I will fix it because that decision has been made.

I have no objection to removing or updating the line number.

tox.ini
```diff
@@ -134,7 +134,7 @@ commands_pre=
     pip freeze
 commands=
     pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data \
-        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:504"
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types:.*"
```
braingram, line 137:
```suggestion
        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.types"
```

braingram (approved):
This looks good to me. I would like to see the jwst job pass before merging.

WilliamJamieson:

PR #1372, opened 2023-02-16
Move label check to using `any_of` instead of `one_of`

WilliamJamieson:
Looking at the documentation for the labeler action used, `any_of` is a better choice because It may become necessary to label with more than one backport label. This change supports that logic.


WilliamJamieson:
The failure is again due to the `pull_request_target` permissions issue.

PR #1389, opened 2023-02-17
run jwst downstream tests in parallel

braingram:
testing speed of jwst tests run in parallel

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -176,6 +176,8 @@ commands =
     pytest gwcs
 
 [testenv:jwst]
+deps=
+    pytest-xdist
 change_dir = {env_tmp_dir}
 allowlist_externals =
     git
@@ -193,7 +195,7 @@ commands_pre =
     pip install -r {env_tmp_dir}/requirements.txt
     pip freeze
 commands =
-    pytest jwst
+    pytest --numprocesses auto jwst
 
 [testenv:stdatamodels]
 change_dir = {env_tmp_dir}


braingram:
Comparing the run of the jwst downstream in this PR (run in parallel on 2 cores with pytest-xdist):
https://github.com/asdf-format/asdf/actions/runs/4207382736/jobs/7302102634
vs one in a recent PR (run serially):
https://github.com/asdf-format/asdf/actions/runs/4207209567/jobs/7301726071
Running the jwst tests in parallel (which jwst [appears to do](https://github.com/spacetelescope/jwst/blob/master/tox.ini#L4) with [similar performance](https://github.com/spacetelescope/jwst/actions/runs/4195427374/jobs/7274982870)) results in a roughly 1.5x speed up (28 vs 43 minutes).

@WilliamJamieson @zacharyburnett should we consider running at least jwst tests (and possibly other parts of the ci) in parallel?

If so, I'm happy to take this PR out of draft or replace (or have it replaced) by a more concerted effort.

zacharyburnett:
You can probably move this into a global tox factor by adding `pytest-xdist` to the `[testenv]` `deps` and also adding 
```
commands=
    pytest --remote-data \
    xdist: -n auto \
    {posargs}
```
Then we can run any of the toxenvs in parallel with `-xdist`

zacharyburnett:
> You can probably move this into a global tox factor in `[testenv]` by adding `pytest-xdist` to `deps` and also adding
> 
> ```diff
>   commands=
> -    pytest --remote-data
> +    pytest --remote-data \
> +    xdist: -n auto \
> +    {posargs}
> ```
> 
> Then we can run any of the toxenvs in parallel with `-xdist`

never mind, I didn't see that the `commands` are overridden; this should be fine for the downstream toxenvs, and then the global one can be another PR

braingram:
closing this, superceded by: #1396

WilliamJamieson:
> closing this, superceded by: #1396

#1396 does not superseed this PR, as it does not parallelize the JWST job.

tox.ini
```diff
@@ -199,7 +201,7 @@ commands_pre=
     pip install -r {envtmpdir}/requirements.txt
     pip freeze
 commands=
-    pytest jwst
+    pytest -n auto jwst
```
WilliamJamieson, line 204:
```suggestion
    pytest --numprocesses auto jwst
```

WilliamJamieson (changes requested):
Why not add the parallelization to all the downstream jobs?

WilliamJamieson (approved):

PR #1394, opened 2023-02-20
Fix bug with running tests in ASDF from path outside asdf

WilliamJamieson:
The use of relative paths for `--ignore` in the `addopts` configuration for `pytest` means that the tests will fail if pytest is triggered from a directory outside the `asdf` directory. This is because those paths are relative to where the `pytest` command is executed. This is causing the CI to fail in asdf-astropy: https://github.com/astropy/asdf-astropy/actions/runs/4220556775/jobs/7334281038 because warnings are being collected.

This PR fixes this by moving the `--ignore` command-line arguments to the `conftest.py`

### DIFF
--- a/asdf/conftest.py
+++ b/asdf/conftest.py
@@ -7,6 +7,8 @@ import pytest
 
 from asdf.tests.httpserver import HTTPServer, RangeHTTPServer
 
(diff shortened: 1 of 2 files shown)


braingram (approved):
LGTM

PR #1399, opened 2023-02-20
Add concurrency cancellations to workflows

WilliamJamieson:
To give us more available runners, jobs triggered by PRs should cancel previous jobs triggered by that PR (from the same workflow), so that only one instance of each job is running per PR. Currently, if a new commit is pushed to a PR, but the CI has not completed then that instance of the CI will continue to run in addition to the new set of CI which will be triggered by the new commit. This ends up causing us to continue to run no-longer-relevant CI jobs to completion, which uses up job runners.

This PR adds concurrency rules based on: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow to our PR related CI jobs. This should start freeing up runners for asdf.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -4,6 +4,14 @@ on:
   pull_request:
     types: [labeled, unlabeled, opened, synchronize, reopened]
 
+# Only cancel in-progress jobs or runs for the current workflow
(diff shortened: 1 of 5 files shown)


zacharyburnett (approved):
LGTM

Issue #1333, opened 2023-01-26
Add deprecations for 'legacy' extension system


braingram:
The 'legacy' extensions should raise deprecation warnings in 2.15 as they will be removed in 3.0. Warnings should be issued for:

- [x] [CustomType](https://github.com/asdf-format/asdf/blob/cbc066dad91f236236dc04c6bd14a08ae84f01cb/asdf/types.py#L400): https://github.com/asdf-format/asdf/pull/1359
- [x] Resolver (see [comment below](https://github.com/asdf-format/asdf/issues/1333#issuecomment-1405703407)): https://github.com/asdf-format/asdf/pull/1362
- [x] [asdf.types](https://github.com/asdf-format/asdf/blob/cbc066dad91f236236dc04c6bd14a08ae84f01cb/asdf/types.py#L1) (this includes [format_tag](https://github.com/asdf-format/asdf/blob/cbc066dad91f236236dc04c6bd14a08ae84f01cb/asdf/types.py#L15)) https://github.com/asdf-format/asdf/pull/1401
- [x] asdf.tests.helpers.assert_extension_correctness: https://github.com/asdf-format/asdf/pull/1388
- [x] [asdf.extension.get_default_resolver](https://github.com/asdf-format/asdf/blob/master/asdf/extension/_legacy.py#L254): https://github.com/asdf-format/asdf/pull/1409
- [x] [asdf.extension.get_cached_asdf_extension_list](https://github.com/asdf-format/asdf/blob/master/asdf/extension/_legacy.py#L167): https://github.com/asdf-format/asdf/pull/1409
- [x] [asdf.extension.default_extensions](https://github.com/asdf-format/asdf/blob/43a9f381ef939931ced73ebf68dc07de68de6ce0/asdf/extension/_legacy.py#L251): https://github.com/asdf-format/asdf/pull/1409
- [x] [asdf.type_index](https://github.com/asdf-format/asdf/blob/master/asdf/type_index.py#L184): https://github.com/asdf-format/asdf/pull/1403
- [x] [asdf.AsdfFile.run_hook and similar](https://github.com/asdf-format/asdf/blob/master/asdf/asdf.py#L1412): https://github.com/asdf-format/asdf/pull/1417
- [x] [asdf.types.AsdfType](https://github.com/asdf-format/asdf/blob/a143b77856e93bd1df6aeff7672c0467ae1c5cde/asdf/types.py#L394): https://github.com/asdf-format/asdf/pull/1401
[...]

PR #1401, opened 2023-02-20
deprecate asdf.types

braingram:


asdf/reference.py
```diff
@@ -12,7 +12,7 @@
 import numpy as np
 
 from . import generic_io, treeutil, util
-from .types import AsdfType
+from ._types import AsdfType
```
WilliamJamieson, line 15: Is `AsdfType` intended to be part of the private API? If so we might want to consider doing `from . import _types` instead, and accessing `AsdfType` via the `_type` namespace. This would apply in the other public modules that use `AsdfType`.

asdf/tags/core/complex.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 
 from asdf import util
-from asdf.types import AsdfType
+from asdf._types import AsdfType
```
WilliamJamieson, line 4: Same `AsdfType` discussion applies here.

WilliamJamieson (dismissed):
I have one question about `AsdfType`, see below.

asdf/reference.py
braingram: AsdfType is currently part of the 'public' API at least with respect to the documentation: https://asdf.readthedocs.io/en/stable/api/asdf.types.AsdfType.html
This form of import was chosen to mimic the existing interface (where AsdfType is also exposed via `asdf.reference`).

asdf/tags/core/complex.py
braingram: Same as in `asdf/reference.py`.

asdf/reference.py
WilliamJamieson: The developer docs say not

> This should not be used for user-defined extensions.

which indicates to me that this really shouldn't be public.

WilliamJamieson: I personally, don't think it should be public.

braingram: I agree that AsdfType should not be public. This PR deprecates `asdf.types` so that we can eventually remove that submodule to start cleaning up the public API.

A follow up PR(s) will deprecate the public classes in `asdf.types` (AsdfType, ExtensionType) (I added these to the list of legacy api deprecations: https://github.com/asdf-format/asdf/issues/1333#issue-1558531069).

A more general clean up of our public API is very much called for but outside the scope of this work.

If making the changes you called our are required to get this PR approved I will make them. Are these required changes for this PR?

WilliamJamieson: > I agree that AsdfType should not be public. This PR deprecates `asdf.types` so that we can eventually remove that submodule to start cleaning up the public API.

> A follow up PR(s) will deprecate the public classes in `asdf.types` (AsdfType, ExtensionType) (I added these to the list of legacy api deprecations: [#1333 (comment)](https://github.com/asdf-format/asdf/issues/1333#issue-1558531069)).

I fail to understand how `AsdfType` is not effectively deprecated as part of the public API by simply deprecating the public module which contains it. Are you suggesting that `AsdfType` should be deprecated in the same fashion as `CustomType`, meaning depreciation of subclasses?

I guess this is fine if we plan on removing `AsdfType` from the public namespace of this public module before the `2.15` release.

braingram: I made the changes you asked about.

To clarify, you consider this sufficient to deprecate `AsdfType`. Is the same true for `ExtensionType`?

WilliamJamieson: After grepping the codebase, I think `ExtensionType` should fall under the same category.

WilliamJamieson (approved):

WilliamJamieson:

WilliamJamieson:

PR #1405, opened 2023-02-20
Remove concurrency cancel for `Label Checker` job

WilliamJamieson:
Concurrency groups act funny on `pull_request_target` triggered workflows (`github.ref` becomes the branch being merged into not a reference to the PR branch). Label Checker is a very small and fast workflow, so we don't need concurrency cancelations with it

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -9,14 +9,6 @@ on:
       - labeled
       - unlabeled
 
-# Only cancel in-progress jobs or runs for the current workflow
-#   This cancels the already triggered workflows for a specific PR without canceling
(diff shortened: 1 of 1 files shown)


zacharyburnett (approved):

pllim (approved):
If you have someone who keeps spamming label events, you should remove their write access and problem solved. I agree you don't need auto cancel. 😉

WilliamJamieson:
> If you have someone who keeps spamming label events, you should remove their write access and problem solved. I agree you don't need auto cancel. 😉

Especially, when they cancel other PR's Label workflow.

PR #1408, opened 2023-02-20
Use reusable changelog workflow

WilliamJamieson:
The meeseeks backports shouldn't have to have a `no-changelog-entry-needed` label for their CI. This PR solves this by moving to the old workflow that was used by astropy for its changelog and is still used by jdaviz. This workflow happens to include skips for meeseeks.

Note the original ASDF workflow was a copy of the astropy one from before it was refactored into the reusable workflow, being used here. So they should operate exactly the same, except for the additional meeseeks logic.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -17,11 +17,8 @@ jobs:
     name: Confirm changelog entry
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


pllim:
You are welcome to use it... 😬

WilliamJamieson:
Label check CI is expected to fail because I fixed the milestone name, this is a `pull_request_target` issue.

zacharyburnett (approved):
LGTM

PR #1417, opened 2023-02-21
deprecate AsdfFile attrs and methods that use legacy extension api

braingram:
Deprecate:
- run_hook
- run_modifying_hook
- url_mapping
- tag_mapping
- type_index
- resolver
- extension_list

### DIFF
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -129,7 +129,7 @@ class AsdfFile:
         self._user_extensions = self._process_user_extensions(extensions)
         self._plugin_extensions = self._process_plugin_extensions()
         self._extension_manager = None
-        self.__extension_list = None
+        self._extension_list_ = None
 
         if custom_schema is not None:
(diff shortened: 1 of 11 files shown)


asdf/asdf.py
```diff
@@ -206,7 +206,7 @@ def version(self, value):
         self._user_extensions = self._process_user_extensions(self._user_extensions)
         self._plugin_extensions = self._process_plugin_extensions()
         self._extension_manager = None
-        self._extension_list = None
+        self.__extension_list = None
```
WilliamJamieson, line 209: Don't use `__` prefixes like this, this causes names to get mangled, see https://docs.python.org/3/tutorial/classes.html#private-variables, which can potentially break things for users.

You should pick a different name such as `_extension_list_` or `_private_extension_list` instead.

braingram: Changed to `_extension_list_`.

WilliamJamieson (approved):

WilliamJamieson:

PR #1433, opened 2023-02-22
move format_tag to testing.helpers

### LINKED ISSUE #1404: Find suitable new location for `format_tag`
This is currently in the soon the be deprecated `asdf.types`:
https://github.com/asdf-format/asdf/blob/f726fe8a2ccaa3faf33713738510aa103836cfa3/asdf/types.py#L17-L29

It does not depend on the legacy type system and could be moved to a new location.

I've only found 1  use of this in `test_schema.py`:
https://github.com/asdf-format/asdf/blob/f726fe8a2ccaa3faf33713738510aa103836cfa3/asdf/tests/test_schema.py#L639

braingram:
fixes #1404


WilliamJamieson (approved):

WilliamJamieson:
Looks like this was used in astropy... I'll make another PR to astropy directly.

PR #1434, opened 2023-02-22
Resolve `PT011` error code

### LINKED ISSUE #1342: Resolve `PT011` error code
This is that `pytest.raises` should check the error message. This is considered "best practice", but there are a lot of these. We should fix them, but it is not a "high priority".

WilliamJamieson:
Fixes #1342

Adds`match` checks to all the `pytest.raises` checks, so that error messages are checked.


braingram (approved):

WilliamJamieson:
close/reopen for RTD

Issue #1363, opened 2023-02-13
Deprecate tests.helpers


braingram:
Move tests.helpers into (and combine with) testing.helpers.

Related to this (and especially if it makes deprecating tests.helpers easier) rename tests to _tests to discourage public use but still allow for testing of the installed/built package.

braingram:
Opening a new issue for tests renaming

PR #1440, opened 2023-02-23
deprecate asdf.tests.helpers

braingram:
The current users of asdf.tests.helpers (outside this repo) are (with 1 example from each):
- astropy.io.misc.asdf
- specutils: https://github.com/astropy/specutils/blob/75a6cba8f7465878823d25e1ddf3ea47c0e73051/specutils/io/asdf/tags/tests/test_spectra.py#L11
- asdf-astropy: https://github.com/astropy/asdf-astropy/blob/706f4ec19d6509966f363c3ade6a4cfd46c4a15d/asdf_astropy/converters/unit/tests/test_unit.py#L5

The uses in the deprecated astropy.io.misc.asdf are ignored.

specutils uses `assert_roundtrip_tree` to test that object serialization then deserialization results in an equivalent object. `asdf.testing.roundtrip_object` can be used but requires either using the [`assert_equal` class method defined on the spectra](https://github.com/astropy/specutils/blob/75a6cba8f7465878823d25e1ddf3ea47c0e73051/specutils/io/asdf/tags/spectra.py#L67), implementation of an equality operator or a similar test to verify that the deserialized object is 'equal' (it currently relies on a feature of `assert_roundtrip_tree` which checks for the `assert_equal` class method of the custom type).

asdf-astropy uses `yaml_to_asdf`. An alternative, drop in replacement exists in `asdf.testing.helpers`.


tox.ini
```diff
@@ -158,7 +160,8 @@ commands_pre =
     pip freeze
 commands =
     pytest specutils \
-        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.entry_points"
+        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.entry_points" \
```
WilliamJamieson, line 163:
```suggestion
```

Fixed by astropy/specutils#1024.

asdf/tests/_helpers.py
```diff
@@ -0,0 +1,499 @@
...
+            try:
+                with generic_io.get_file(schema_location) as f:
+                    yaml.safe_load(f.read())
+            except Exception as err:  # noqa: BLE001
+                msg = (
+                    f"{extension_type.__name__} supports tag, {check_type.yaml_tag}, "
+                    f"which resolves to schema at {schema_location}, but "
+                    "schema cannot be read."
+                )
+                raise AssertionError(msg) from err
```
WilliamJamieson, line 499: Clearly downstream packages are using some of this. We cannot just hide it away like this without providing some of it back to our downstream. I thought the plan was to move this to `asdf.testing.helpers`? If so those changes need to be part of the scope of this PR.

WilliamJamieson (changes requested):

WilliamJamieson:
> specutils uses `assert_roundtrip_tree` to test that object serialization then deserialization results in an equivalent object. `asdf.testing.roundtrip_object` can be used but requires either using the [`assert_equal` class method defined on the spectra](https://github.com/astropy/specutils/blob/75a6cba8f7465878823d25e1ddf3ea47c0e73051/specutils/io/asdf/tags/spectra.py#L67), implementation of an equality operator or a similar test to verify that the deserialized object is 'equal' (it currently relies on a feature of `assert_roundtrip_tree` which checks for the `assert_equal` class method of the custom type).
> 
> asdf-astropy uses `yaml_to_asdf`. An alternative, drop in replacement exists in `asdf.testing.helpers`.

If this is the case, please open PRs on `specutils` and `asdf-astropy` implementing these fixes. I would like to review those changes first before I consider the changes in this PR.

Also, have you checked the if the new warnings are raised in the rest of the downstream tests? The three packages you list here are simply the ones which turn warnings into errors. I would like to cover off these changes in the rest of the downstream.

WilliamJamieson (approved):
After an offline conversation with @braingram, it was explained that his analysis indicates that   they only downstream uses of `asdf.tests.helpers` were in `astropy`, `specutils`, and `asdf-astropy` which already have solution paths planned.

Given this analysis, I retract my previous comments (aside from removing the single suppression from the downstream tox).

These changes look good to me otherwise.

asdf/tests/_helpers.py
WilliamJamieson: Offline conversation resolved this question

tox.ini
WilliamJamieson: This is the only change I really would like (simply to avoid yet another PR).

braingram: Thanks for the fix. I removed this warning filter.

braingram:
Fixes #1363

braingram:
> If this is the case, please open PRs on `specutils` and `asdf-astropy` implementing these fixes. I would like to review those changes first before I consider the changes in this PR.

@WilliamJamieson I opened PRs:
https://github.com/astropy/asdf-astropy/pull/168
https://github.com/astropy/specutils/pull/1025

I converted them both to draft in case we'd like to rename testing.helpers. I'm inclined to leave it as is but don't feel strongly. If you'd like to change the name I don't have any objection.

PR #1445, opened 2023-02-24
Module shadowing

braingram:
To add deprecation warnings, several public sub modules were 'shadowed' by moving them to a private file (prefixed with `_`) and adding a stub public module in their place that issued a deprecation warning then replaced itself with the private module. Here is one example:
https://github.com/asdf-format/asdf/blob/6b4701277bb25c1d4e8583b88bc2b125be9a36d6/asdf/types.py#L1-L26

This example illustrates one issue with this approach where certain objects needed special handling to allow doc references to work.

An alternative approach was used for `asdf.tests.helpers`.
https://github.com/asdf-format/asdf/blob/6b4701277bb25c1d4e8583b88bc2b125be9a36d6/asdf/tests/helpers.py#L1-L26

That has several benefits including:
- easier control over what attributes are exposed
- a hook to allow deprecation warnings to be issued for every attribute access and not just during module import (which can sometimes result in retrieval of a cached import and no deprecation warning).

This PR updates `asdf.types`, `asdf.type_index` and `asdf.resolver` module shadowing to align with the approach taken with `asdf.tests.helpers`.

### DIFF
--- a/asdf/resolver.py
+++ b/asdf/resolver.py
@@ -1,16 +1,27 @@
-import sys
 import warnings
 
 from . import _resolver
 from .exceptions import AsdfDeprecationWarning
 
-warnings.warn(
-    "asdf.resolver is deprecated "
-    "Please see Resources "
-    "https://asdf.readthedocs.io/en/stable/asdf/extending/resources.html",
-    AsdfDeprecationWarning,
-)
 
-# overwrite the hidden module __file__ so pytest doesn't throw an ImportPathMismatchError
-_resolver.__file__ = __file__
-sys.modules[__name__] = _resolver
+def _warn():
+    warnings.warn(
+        "asdf.resolver is deprecated "
+        "Please see Resources "
+        "https://asdf.readthedocs.io/en/stable/asdf/extending/resources.html",
+        AsdfDeprecationWarning,
+    )
+
+
+_warn()
+
+
+def __getattr__(name):
+    attr = getattr(_resolver, name)
+    _warn()
+    attr.__module__ = __name__
+    return attr
+
+
+def __dir__():
+    return dir(_resolver)
--- a/asdf/resolver.py
+++ b/asdf/resolver.py
@@ -19,7 +19,8 @@ _warn()
 def __getattr__(name):
     attr = getattr(_resolver, name)
     _warn()
-    attr.__module__ = __name__
+    if hasattr(attr, "__module__"):
+        attr.__module__ = __name__
     return attr
 
 
--- a/asdf/types.py
+++ b/asdf/types.py
@@ -20,5 +20,6 @@ __all__ = _types.__all__  # noqa: PLE0605
 def __getattr__(name):
     attr = getattr(_types, name)
     _warn()
-    attr.__module__ = __name__
+    if hasattr(attr, "__module__"):
+        attr.__module__ = __name__
     return attr
--- a/asdf/types.py
+++ b/asdf/types.py
@@ -1,26 +1,24 @@
-import sys
 import warnings
 
 from . import _types
 from .exceptions import AsdfDeprecationWarning
 
-warnings.warn(
-    "asdf.types is deprecated "
-    "Please see the new extension API "
-    "https://asdf.readthedocs.io/en/stable/asdf/extending/converters.html",
-    AsdfDeprecationWarning,
-)
+
+def _warn():
+    warnings.warn(
+        "asdf.types is deprecated "
+        "Please see the new extension API "
+        "https://asdf.readthedocs.io/en/stable/asdf/extending/converters.html",
+        AsdfDeprecationWarning,
+    )
 
 
-# overwrite the hidden module __file__ so pytest doesn't throw an ImportPathMismatchError
-_types.__file__ = __file__
-# overwrite the module name for each defined class to allow doc references to work
-for class_ in [
-    _types.ExtensionTypeMeta,
-    _types.AsdfTypeMeta,
-    _types.ExtensionType,
-    _types.AsdfType,
-    _types.CustomType,
-]:
-    class_.__module__ = __name__
-sys.modules[__name__] = _types
+_warn()
+__all__ = _types.__all__  # noqa: PLE0605
+
+
+def __getattr__(name):
+    attr = getattr(_types, name)
+    _warn()
+    attr.__module__ = __name__
+    return attr
--- a/asdf/types.py
+++ b/asdf/types.py
@@ -20,6 +20,6 @@ __all__ = _types.__all__  # noqa: PLE0605
 def __getattr__(name):
     attr = getattr(_types, name)
     _warn()
-    if hasattr(attr, "__module__"):
+    if hasattr(attr, "__module__") and name != "format_tag":
         attr.__module__ = __name__
     return attr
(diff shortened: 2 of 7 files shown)


braingram:
The new warnings in astropy:
https://github.com/asdf-format/asdf/actions/runs/4265734101/jobs/7425422437#step:9:300
were likely not seen because of the entry point warning filtering for astropy.io.misc hiding the initial types deprecation warning:
https://github.com/asdf-format/asdf/blob/6b4701277bb25c1d4e8583b88bc2b125be9a36d6/asdf/entry_points.py#L68-L71
Because asdf._types replaced asdf.types subsequent use of asdf.types (even with warnings cleared) did not emit a new deprecation warning. With this PR, the [use of asdf.types.format_tag](https://github.com/astropy/astropy/blob/b5b3f40c5aad1323369c1f0327ad177e3e001d3a/astropy/io/misc/asdf/tags/tests/helpers.py#L25) causes a new asdf.types deprecation to be issued (the new warning seen here).

braingram:
The errors in roman_datamodels appear unrelated:
https://github.com/asdf-format/asdf/actions/runs/4265734101/jobs/7425422199#step:9:247
and may be due to:
https://github.com/spacetelescope/roman_datamodels/pull/125
Independent tests are needed.

braingram:
The weldx errors:
https://github.com/asdf-format/asdf/actions/runs/4265734101/jobs/7425423059#step:9:432
are due to additional issues of the existing deprecation warning. This time the warning is occurring during schema validation which is turned into an error by pytest_asdf:
https://github.com/asdf-format/asdf/blob/6b4701277bb25c1d4e8583b88bc2b125be9a36d6/pytest_asdf/plugin.py#L249-L251

braingram:
I added a warning filter (treating `AsdfDeprecationWarning` as `default` instead of `error`) during schema validation. This should fix the weldx failures (but not the root cause which is they are using old style validators).

braingram:
It is unclear to me why the docs fail to properly include `asdf.testing.helpers.format_tag` (and instead generate a warning and list, but fail to link to the function) with these changes but the modifications in commit https://github.com/asdf-format/asdf/pull/1445/commits/9f28dbb58d80d59bf065294e5c24a8afa3addb92 allow them to build.

The warnings issued (without the changes in the above linked commit) are:
```
asdf/docs/asdf/developer_api.rst:220: WARNING: autosummary: stub file not found 'asdf.testing.helpers.format_tag'. Check your autosummary_generate setting.
asdf/docs/asdf/developer_api.rst:224:<autosummary>:1: WARNING: py:obj reference target not found: asdf.testing.helpers.format_tag
```
```

**Comment by CagtayFabry** [comment]
> The weldx errors: https://github.com/asdf-format/asdf/actions/runs/4265734101/jobs/7425423059#step:9:432 are due to additional issues of the existing deprecation warning. This time the warning is occurring during schema validation which is turned into an error by pytest_asdf:
> 
> https://github.com/asdf-format/asdf/blob/6b4701277bb25c1d4e8583b88bc2b125be9a36d6/pytest_asdf/plugin.py#L249-L251

thank you for looking into this
once 2.15 is out I plan to build everything on `>=2.15` and remove deprecated calls (including the legacy validation extension)

**Comment by WilliamJamieson** [comment]
Why is astropy failing in this branch but not with the changes on `master`?

**Comment by braingram** [comment]
This comment describes the changes with respect to astropy: https://github.com/asdf-format/asdf/pull/1445#issuecomment-1444386535

In brief, the change in module shadowing creates more warnings with this PR.

For this example:
```python
import warnings

import asdf

with warnings.catch_warnings():
    warnings.simplefilter('ignore')
    import asdf.types

asdf.types.format_tag('a', 'b', 'c', 'd')
```

run against master produces
```
AsdfDeprecationWarning: asdf.types.format_tag is deprecated. Please use asdf.testing.helpers.format_tag
```
with this PR there is an additional warning:
```
AsdfDeprecationWarning: asdf.types.format_tag is deprecated. Please use asdf.testing.helpers.format_tag
AsdfDeprecationWarning: asdf.types is deprecated Please see the new extension API https://asdf.readthedocs.io/en/stable/asdf/extending/converters.html
```

WilliamJamieson:
> In brief, the change in module shadowing creates more warnings with this PR.

Yes, as noted by https://github.com/asdf-format/asdf/pull/1445#issuecomment-1448659885 you changed the warning message for `format_tag`. While annoying, its easily fixed by astropy/astropy#14474, which covers both versions of the warning.

braingram:
> > In brief, the change in module shadowing creates more warnings with this PR.
> 
> Yes, as noted by [#1445 (comment)](https://github.com/asdf-format/asdf/pull/1445#issuecomment-1448659885) you changed the warning message for `format_tag`. While annoying, its easily fixed by [astropy/astropy#14474](https://github.com/astropy/astropy/pull/14474), which covers both versions of the warning.

I see no change in the format of the warning for format_tag. Can you provide an example?

asdf/resolver.py
```diff
@@ -1,16 +1,28 @@
...
+
+_warn()
+
+
+def __getattr__(name):
+    attr = getattr(_resolver, name)
+    _warn()
+    if hasattr(attr, "__module__"):
+        attr.__module__ = __name__
+    return attr
```
WilliamJamieson, line 24: Can you add a test to `test_deprecated.py` to actually test the functionality of this method?

asdf/resolver.py
```diff
@@ -1,16 +1,28 @@
...
+
+
+def __dir__():
+    return dir(_resolver)
```
WilliamJamieson, line 28: Add tests for the functionality of this method to `test_deprecated.py`

asdf/types.py
```diff
@@ -1,26 +1,25 @@
...
+_warn()
+__all__ = _types.__all__  # noqa: PLE0605
+
+
+def __getattr__(name):
+    attr = getattr(_types, name)
+    _warn()
+    if hasattr(attr, "__module__") and name != "format_tag":
+        attr.__module__ = __name__
+    return attr
```
WilliamJamieson, line 25: Are there tests fully covering this?

WilliamJamieson (changes requested):
Requests for additional tests.

WilliamJamieson:
> I see no change in the format of the warning for format_tag. Can you provide an example?

With the `master` branch of `asdf` the message for `format_tag` emitted in astropy is filtered by https://github.com/astropy/astropy/blob/38145e2c2c6a024eeec010e91dd95407fa3d17c4/astropy/io/misc/asdf/tags/tests/helpers.py#L20-L25. That is it has the form:
```python
r"asdf.types.format_tag is deprecated.*",
```
The origin point of the deprecation message is here: https://github.com/asdf-format/asdf/blob/056c3ffc8639eb864553fbb967abb3d49a09cf17/asdf/_types.py#L21-L24

With the changes in this PR two warning messages get raised for the `from asdf.types import format_tag`, first the one originating here: https://github.com/braingram/asdf/blob/125915dc78c831cec96eab86c0ee913c6bb1b37e/asdf/types.py#L8-L13 followed by the one being filtered for in `astropy`. This means you have changed the warnings being issued by `format_tag`.

It is a bit redundant to issue two warnings for `format_tag`, especially with the first message being different from the original message being issued.

WilliamJamieson:
> I see no change in the format of the warning for format_tag. Can you provide an example?

astropy/astropy#14474 was just merged, and it modified the filter to catch both warning messages together so the question is moot.

WilliamJamieson (approved):

PR #1451, opened 2023-02-28
respect umask when determining file permissions

### LINKED ISSUE #1450: asdf produces files with file permissions that differ from default files
```python
with open('test', 'wb') as f:
    pass
```
produces a file with
`-rw-r--r--`

```python
af = asdf.AsdfFile({})
af.write_to('test.asdf')
```
produces a file with
`-rw-------`

braingram:
fixes #1450

### DIFF
--- a/asdf/generic_io.py
+++ b/asdf/generic_io.py
@@ -27,6 +27,10 @@ from .util import patched_urllib_parse
 __all__ = ["get_file", "get_uri", "resolve_uri", "relative_uri"]
 
 
+_FILE_PERMISSIONS_DEFAULT_UMASK = 0o22
+_FILE_PERMISSIONS_ALL = 0o777
+_FILE_PERMISSIONS_NO_EXECUTE = 0o666
+
 _local_file_schemes = ["", "file"]
 if sys.platform.startswith("win"):  # pragma: no cover
     import string
@@ -399,6 +403,7 @@ class GenericFile(metaclass=util.InheritDocstrings):
         """
         if self._close:
             self._fd.close()
+            self._fix_permissions()
 
     def truncate(self, size=None):
         """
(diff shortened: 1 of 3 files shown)


asdf/generic_io.py
```diff
@@ -816,10 +817,42 @@ def flush_memmap(self):
...
+            if not os.path.exists(fn):
+                return
+            # there is no way to read the umask without setting it
+            # so set it to the typical default 0o22
+            umask = os.umask(0o22)
+            if umask != 0o22:
+                # restore the read value if it differs from the default
+                os.umask(umask)
+            permissions = 0o777 if os.path.isdir(fn) else 0o666
+            os.chmod(self._fd._filename, permissions - umask)
```
WilliamJamieson, line 843: I don't like the use of "magic" octal numbers. Yes they "look-like" `chmod` arguments, but those are annoying to figure out and often require me looking up how to decode the values. This obfuscates the code enough that it is hard to read.

Instead can you make all these magic octal numbers module constants like `_FILE_PERMISSIONS_FOO = 0o22` where `FOO` describes what the value means?

Alternately, you could create a `FilePermissions` [`Enum`](https://docs.python.org/3/library/enum.html) subclass, which again encodes what the "magic" values mean.

asdf/tests/test_generic_io.py
```diff
@@ -821,3 +822,19 @@ def test_fsspec_http(httpserver):
         gf = generic_io.get_file(f)
         r = gf.read(len(ref))
         assert r == ref, (r, ref)
+
+
+@pytest.mark.skipif(
+    sys.platform.startswith("win"),
+    reason="Cannot test write file permissions on windows",
+)
+@pytest.mark.parametrize("umask", [0o22, 0o2, 0o0, 0o44, 0o66])
```
WilliamJamieson, line 831: Again can you do as I asked before in order document what each of these octal numbers mean?

Also, without figuring out these 5 values, does this cover all possible permutations of permissible file permissions?

WilliamJamieson (changes requested):
The changes look good, I just request better documentation of some of the values involved.

PR #1454, opened 2023-02-28
Drop pytest-openfiles dependency

WilliamJamieson:
astropy/pytest-openfiles#44 starts the retirement process of `pytest-openfiles`, which will complete in November 2023. `pytest-openfiles` can be entirely replaced by catching `ResourceWarnings` from emitted by `pytest >= 7`. Since ASDF turns all warnings into errors, if we require `pytest>=7` for testing, then we no-longer need `pytest-openfiles`.

This PR removes `pytest-openfiles` from our testing and pins `pytest` to above version 7.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,8 @@ The ASDF Standard is at v1.6.0
 - Add AsdfDeprecationWarning to asdf_extensions entry point [#1361]
 - Deprecate asdf.tests.helpers [#1440]
 - respect umask when determining file permissions for written files [#1451]
+- Remove the retired ``pytest-openfiles`` dependency and replace it with ``pytest>=7``
+  instead. [#1454]
 
 2.14.3 (2022-12-15)
 -------------------
--- a/asdf/tests/test_fits_embed.py
+++ b/asdf/tests/test_fits_embed.py
@@ -380,7 +380,7 @@ def test_verify_with_astropy(tmp_path, dtype):
 def test_dangling_file_handle(tmp_path):
     """
     This tests the bug fix introduced in #533. Without the bug fix, this test
-    will fail when running the test suite with pytest-openfiles.
+    will fail with a ResourceWarning about unclosed file handles with pytest >= 7.
     """
     import gc
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,9 +46,8 @@ tests = [
   "gwcs>=0.18.3",
   "lz4>=0.10",
   "psutil",
-  "pytest>=6",
+  "pytest>=7",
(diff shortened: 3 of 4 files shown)


braingram:
Do we have a test to verify this is a working replacement?

Testing with pytest 7.2.1 and the following test (added to `test_generic_io`)
```python
_d = {}


def test_open_file_error(tmp_path):
    _d['foo'] = generic_io.get_file(tmp_path / 'test', mode='w')
```

I see no error unless I run pytest with `--open-files`.

WilliamJamieson:
See astropy/pytest-openfiles#44, it should raise a `ResourceWarning` if its real. If you have a problem with this please raise it in `pytest-openfiles`. Otherwise, we need to drop it before it is retired.

braingram (changes requested):
I'm against dropping this at the moment.

We are planning for likely 2 major releases before November including a major refactoring of our code (dropping the deprecated features in 3.0). Given that pytest missed catching the test I mentioned it appears the plugin is more capable and will be useful for developing these releases.

WilliamJamieson:
> I'm against dropping this at the moment.
> 
> We are planning for likely 2 major releases before November including a major refactoring of our code (dropping the deprecated features in 3.0). Given that pytest missed catching the test I mentioned it appears the plugin is more capable and will be useful for developing these releases.

I disagree, first your example doesn't actually open a file pointer, it simply passes a file name which gets wrapped as `pytest` is tracking the calls of `tracemalloc`, this is even noted in the docstring of the function you are using: https://github.com/asdf-format/asdf/blob/3c83a49de6ab7455a4cd2bcc4aea65704a582db3/asdf/generic_io.py#L1057-L1059

Now if one passes an open file in the test case, then on the `pytest` exit `python` raises the `ResourceWarning`. This is due to the fact that the construction of your test case is poor because until the program ends, `_d` is in-fact holding a reference to the file which can be closed as the entire test module will still be open in memory. It is only when the program ends that a dangling file pointer is left. This only occurs outside the scope of `pytest`.

Second, this means that the `pytest-openfiles` flag on the test in that case is due to the buggy problem described in astropy/pytest-openfiles#32, though in this case there is not even a proper fixture in place . Indeed, this issue is why `astropy` is dropping support for `pytest-openfiles`. If the dictionary was a proper fixture, then `pytest` would in fact emit the `ResourceWarning` at the point of closing the fixture, or the fixture would clean its file pointers up. In either case the issue is covered off.

In any event, if you are against dropping `pytest-openfiles` at this time due to your concerns then please open an issue detailing your concerns with `pytest-openfiles`. I will close this PR in favor of #1458.

PR #1460, opened 2023-03-02
Rename tests to _tests and consolidate tests to new _tests directory

### LINKED ISSUE #1441: Rename tests to _tests to discourage public use
Splitting issue #1363

braingram:
Fixes #1441


asdf/_tests/__init__.py
```diff
@@ -0,0 +1,29 @@
+"""
+This packages contains affiliated package tests.
```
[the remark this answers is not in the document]
braingram: That does seem unhelpful. I removed this comment.

braingram:
> Can we move `asdf/conftest.py` to `asdf/_tests/conftest.py` as well? Now that the tests themselves are quarantined to that directory, I don't think we need it where it is.

Good idea!
I consolidated `conftest.py`, parts of `asdf/conftest.py` and `asdf/_tests/conftest.py`.

`asdf/conftest.py` contains a list of legacy modules that are ignored to allow for error-free test collection (these raise `AsdfDeprecationWarning` on import which turns into an error and breaks collection). This could be removed entirely when these public modules are removed (ASDF 3.0). I like having the ignores there (instead of in the pyproject.toml) as they will be included in the installed package. Moving them to `asdf/_tests/conftest.py` (and prepending `../`) doesn't appear to work possibly because pytest walks the directories before visiting subdirectories.

conftest.py
```diff
@@ -1,18 +0,0 @@
...
-    This fixture creates a temporary current working directory
-    for the test session, so that docstring tests that write files
-    don't clutter up the real cwd.
-    """
-    original_cwd = os.getcwd()
-    try:
-        os.chdir(tmpdir_factory.mktemp("cwd"))
-        yield
-    finally:
-        os.chdir(original_cwd)
```
WilliamJamieson, line 18: You moved this to the `asdf._tests.conftest` this means it will only be scoped to tests in `asdf._tests` (yes it says session, but the tests in the session it applies to are the ones in its directory or below); however, there are tests in many other directories above this one like the `asdf.commands._tests`. This will also effect running the doctests (if there are any) in `asdf/docs`.

braingram: Thanks for the catch. I moved that fixture back.

WilliamJamieson (approved):

Issue #1348, opened 2023-02-03
Add upper pins to dependencies for LTS release


braingram:
@WilliamJamieson mentioned that dependabot might be useful for this.

WilliamJamieson:
After reading @eslavich's suggested article in https://github.com/asdf-format/asdf/pull/1461#issuecomment-1455251472. I am convinced that upper pins on dependencies is a fundamentally bad idea, which will create more issues later on than it solves.

An alternate approach something like what I mentioned in https://github.com/asdf-format/asdf/pull/1461#issuecomment-1456363662:

> Maybe a solution is to add a requirements-tested.txt to 2.15.x when we sunset the 2.15.x version of ASDF? This way someone would have a configuration they could install if it starts breaking.

That is we include, some sort of "lock" file which is updated in the `2.15.x` branch which lays out (lower and) upper pins based on what versions we have actually tested for. Then hypothetically, future users can use this as a basis for generating working environments for working with legacy/unsupported versions of ASDF.

I think we can write some scripting to generate such a lock file from installing ASDF (without) pins. This could be run in CI to to periodically create PRs updating this file.

braingram:
Closing per discussion here and https://github.com/asdf-format/asdf/pull/1461#issuecomment-1456900199

PR #1461, opened 2023-03-02
add upper pins to dependencies 

braingram:
these are added to packages using the following strategy:
- development version (<1.0) are left unpinned as it seems overly
  conservative to pin these to a minor version and as they're <1.0
  there is little guarantee that a minor version won't introduce
  a breaking change. Packages effected are:
    - asdf-transform-schemas
    - asdf-unit-schemas
    - jmespath
- packages that seem to follow semantic versioning were pinned
  to the current working major version. These are:
    - asdf-standard
    - importlib-metadata
    - importlib_resources
    - jsonschema
    - pyyaml
    - semantic_version
- packages that don't follow semantic versioning were pinned
  to the current working version. These are:
    - numpy
    - packaging


WilliamJamieson:
This PR should not be against the `master` branch as it unnecessarily makes working with development versions of the ASDF dependencies difficult, as pip will install the upper pinned version of these dependencies over any newer versions being worked with every time ASDF is reinstalled.

The point of #1348 was to add upper pins to the LTS release only. This release will occur from the `2.15.x` branch, not the `master` branch of ASDF.

braingram:
Thanks for looking at this.

That all sounds great. I opened this to test how this would work with our CI and as expected it breaks the numpy devdeps testing.

Other than the branch target, what do you think of the strategy and pins?

WilliamJamieson:
Looks like you need to rebase your changes on `2.15.x` as well.

WilliamJamieson:
Fixed it

pyproject.toml
```diff
@@ -19,18 +19,17 @@ dynamic = [
   'version',
 ]
 dependencies = [
-  "asdf-standard>=1.0.1",
+  "asdf-standard<2,>=1.0.1",
```
WilliamJamieson, line 22: These should be exact pins to the latest releases of these packages. `dependabot` should auto increment this via opening PRs.

WilliamJamieson (changes requested):

WilliamJamieson:
> Some interesting reading on this subject: [iscinumpy.dev/post/bound-version-constraints](https://iscinumpy.dev/post/bound-version-constraints/)

Yeah this is the balancing act. Eventually the 2.15.x version will stop working with at least one of the newly released versions of the dependencies at which point we will either need to do a hot fix pinning the dependency or basically leave it to users to do a custom install with that older version. If we upper pin then we can guarantee (up to our level of CI testing) that version of ASDF on PyPi will "function" (well assuming a dependency of a dependency doesn't break something). 

The general idea, was that on this version we would upper pin and then let dependabot open PRs bumping pinned versions as new releases occur, this would run our CI on it and tell us if everything works. We would then immediately release a new bugfix version of 2.15.x with the bumped dependency upon merging these PRs (so fairly regular bugfix releases). Since the release system is now very automated, it should not be too much work to do this.

On the other hand we run our CI against the dev version of most of our major dependencies, https://github.com/asdf-format/asdf/blob/master/.github/workflows/scheduled.yml will run that CI regularly on 2.15.x (at least until we drop support for 2.15.x vresions). So someone should get alerted if 2.15.x will be broken by one of its dependencies.

Maybe a solution is to add a `requirements-tested.txt` to `2.15.x` when we sunset the 2.15.x version of ASDF? This way someone would have a configuration they could install if it starts breaking.

WilliamJamieson:
Given my comment https://github.com/asdf-format/asdf/issues/1348#issuecomment-1456897699, I think we should close this PR and #1348, and open a new issue detailing how to handle the issues this PR is trying to solve.

PR #1464, opened 2023-03-06
Remove legacy extension api

braingram:


braingram:
weldx downstream job failure for this job looks like it should be fixed by weldx PR: https://github.com/BAMWelDX/weldx/pull/863

CagtayFabry:
> weldx downstream job failure for this PR looks like it should be fixed by weldx PR: [BAMWelDX/weldx#863](https://github.com/BAMWelDX/weldx/pull/863)

Yes we will release a new version soon based on the new extension API which should fix the downstream job

CagtayFabry:
feel free to rerun the downstream welx test again now with weldx 0.6.5 @braingram

pllim:
This might have broken 2 of the astropy release branches: v5.2.x and v5.3.x

astropy v6.0.dev seems compatible but that is not going to be released until the end of 2023. Any chances this can be reverted and then wait 6 months? It is the v5.3.x that I worry about.

Example logs:

* https://github.com/astropy/astropy/actions/runs/4920084061/jobs/8788506464
* https://github.com/astropy/astropy/actions/runs/4920080182/jobs/8788497391

cc @WilliamJamieson

pllim:
p.s. I opened https://github.com/astropy/astropy/issues/14795

braingram:
> astropy v6.0.dev seems compatible but that is not going to be released until the end of 2023. Any chances this can be reverted and then wait 6 months? It is the v5.3.x that I worry about.

I would prefer to find a solution that does not revert and delay these changes for 6 months.

One option might be to install asdf-astropy and require it for building the docs on v5.3.x and v5.2.x (and update the documentation for the modeling example to describe that asdf-astropy is now required for saving astropy models to a file). Similar changes appear to have been included in https://github.com/astropy/astropy/pull/14668

pllim:
That would be too confusing for astropy users, as we have not removed deprecated code yet in those branches. I can always pin maxversion of asdf for those branches but I have a feeling that @nden won't like that.

pllim:
Also, what is the rush of removing code here so much so soon?

braingram:
> That would be too confusing for astropy users, as we have not removed deprecated code yet in those branches. I can always pin maxversion of asdf for those branches but I have a feeling that @nden won't like that.

I must be missing something. How would adding a dependency for building the docs be confusing to an astropy user?

pllim:
We are breaking the promise of requiring something that should not be required.

pllim:
Didn't y'all write a whole long-term support policy a few months ago? I thought downstream would have another year, which wouldn't be a problem for us by then because astropy 6.0 should come out in 6 months.

braingram:
> Didn't y'all write a whole long-term support policy a few months ago? I thought downstream would have another year, which wouldn't be a problem for us by then because astropy 6.0 should come out in 6 months.

It was agreed that ASDF does not have the resources to support formal LTS releases. Would committing to supporting 2.15 for 6 months meet your needs?

The currently released version of ASDF (2.15) should allow the doctest you linked to pass. However the run you linked installed the 3.0 development branch (main) of ASDF. This will continue to fail even if we can agree to support 2.15.

> It would be very easy for me to pin `asdf<3` for `astropy 5.x` series, but it means STScI cannot use asdf 3 with stable astropy for another 6 months.

Is that the solution you prefer?

pllim:
> Is that the solution you prefer?

If you cannot revert this, then that is what I must do. It is wrong to introduce breaking changes in v5.3.x after the freeze. We are already in RC.

Thanks for your inputs!

PR #1467, opened 2023-03-08
remove astropy and specutils from downstream

braingram:
astropy was included here because of astropy.io.misc.asdf which is deprecated and uses the legacy extension api (which will be removed during this development cycle). The remaining usage of asdf in astropy is through asdf-astropy.

specutils includes a legacy extension that may be moved to asdf-astropy. Per advice from WilliamJamieson specutils is being removed from downstream testing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -114,22 +114,6 @@ commands_pre =
 commands =
     pytest asdf-coordinates-schemas
 
-[testenv:astropy]
-change_dir = {env_tmp_dir}
-allowlist_externals =
-    git
-    bash
-extras =
-commands_pre =
-    bash -c "pip freeze -q | grep 'asdf @' > {env_tmp_dir}/requirements.txt"
-    git clone https://github.com/astropy/astropy.git
-    pip install -e astropy[test]
-    pip install -r {env_tmp_dir}/requirements.txt
-    pip freeze
-commands=
-    pytest astropy/astropy/io/misc/asdf --open-files --run-slow --remote-data \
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -143,24 +127,7 @@ commands_pre =
...
-extras =
-commands_pre =
-    bash -c "pip freeze -q | grep 'asdf @' > {env_tmp_dir}/requirements.txt"
-    git clone https://github.com/astropy/specutils.git
-    pip install -e specutils[test]
-    pip install -r {env_tmp_dir}/requirements.txt
-    pip freeze
-commands =
-    pytest specutils \
-        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.tests.helpers"
```
WilliamJamieson, line 163: Leave this in because we may want to add specutils back once its extension has been modernized. This will let us easily test that.

WilliamJamieson (changes requested):

tox.ini
braingram: Sounds good.

WilliamJamieson (approved):

tox.ini
```diff
@@ -143,8 +127,7 @@ commands_pre =
     pip install -r {env_tmp_dir}/requirements.txt
     pip freeze
 commands =
-    pytest asdf-astropy \
-        -W "ignore::asdf.exceptions.AsdfDeprecationWarning:asdf.tests.helpers"
```
WilliamJamieson, line 147: Note, this change should be "backported" to 2.15.x.

PR #1468, opened 2023-03-08
Move array storage settings to config

braingram:
This PR moves:
- `all_array_storage`
- `all_array_compression`
- `all_array_compression_kwargs`

into `asdf.config.AsdfConfig`. These were previously passed around as arguments to several functions during `AsdfFile` writes. By moving them into the config the `AsdfFile` can continue to accept them as arguments to `AsdfFile.write_to` and `AsdfFile.update` (by using the arguments to set the values in the `AsdfConfig` context created within those functions) and now users can also adjust these settings in the config directly.

This helps to make these storage settings available to all functions called within the `AsdfConfig` contexts created during write without having to pass them around as arguments (this proved important for the upcoming deferred block and Converter block storage support).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ The ASDF Standard is at v1.6.0
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Drop support for ASDF-in-FITS. [#1288]
+- Add ``all_array_storage``, ``all_array_compression`` and
+  ``all_array_compression_kwargs`` to ``asdf.config.AsdfConfig`` [#1468]
 
 2.15.0 (unreleased)
 -------------------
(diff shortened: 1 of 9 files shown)


perrygreenfield:
Let me see if I understand correctly. The main advantage is to remove the necessity of passing `ctx` as an argument to many of these functions to clean up the interface? Otherwise the changes look good to me. I just want to make sure I understand the the goal.

braingram:
> Let me see if I understand correctly. The main advantage is to remove the necessity of passing `ctx` as an argument to many of these functions to clean up the interface? Otherwise the changes look good to me. I just want to make sure I understand the the goal.

Exactly!

It also allows everything within the context to have access to the setting. For example the eventual ndarray converter will benefit from access to these settings during a call to `to_yaml_tree`. Currently `to_yaml_tree` only receives a `SerializationContext` (as the `ctx` argument) and not a reference to the `AsdfFile` (where the settings are located prior to this PR). With this PR the `all_array_storage` (etc) settings are available within the scope of the call to `Converter.to_yaml_tree`.

PR #1469, opened 2023-03-09
Block data callback

braingram:
This adds a `data_callback` argument to `asdf.block.Block` that allow the block to dynamically load data during write. This is work towards a deferred block implementation where this callback will be used during the deferred write.

### DIFF
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -839,7 +839,11 @@ class Block:
         ],
     )
 
-    def __init__(self, data=None, uri=None, array_storage="internal", memmap=True, lazy_load=True):
+    def __init__(self, data=None, uri=None, array_storage="internal", memmap=True, lazy_load=True, data_callback=None):
+        self._data_callback = data_callback
+        if self._data_callback is not None and data is not None:
+            msg = "Block.__init__ cannot contain non-None data and a non-None data_callback"
+            raise ValueError(msg)
         self._data = data
         self._uri = uri
         self._array_storage = array_storage
@@ -1032,6 +1036,9 @@ class Block:
             If `True`, validate the data against the checksum, and
             raise a `ValueError` if the data doesn't match.
         """
+        if self._data_callback is not None:
+            msg = "read called on a Block with a data_callback"
+            raise RuntimeError(msg)
         offset = None
         if fd.seekable():
             offset = fd.tell()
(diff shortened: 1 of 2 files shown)


braingram:
Failures are unrelated and addressed in #1470

WilliamJamieson (changes requested):
This needs a changelog as `Block` is a public class.

braingram:
> This needs a changelog as `Block` is a public class.

Based on this comment I took `Block` to not be a public class:
https://github.com/asdf-format/asdf/blob/b8e1759aa9f013283942140c94732a170eb9a703/asdf/block.py#L799-L801

I do not think this should be public and would rather not include a changelog entry which might cause users to think it's public.

Are there changes that should be made to make this more explicit?

WilliamJamieson:
> Based on this comment I took `Block` to not be a public class:
> 
> https://github.com/asdf-format/asdf/blob/b8e1759aa9f013283942140c94732a170eb9a703/asdf/block.py#L799-L801
> 
> I do not think this should be public and would rather not include a changelog entry which might cause users to think it's public.
> 

According to [`pep8`](https://peps.python.org/pep-0008/#public-and-internal-interfaces) this is a public interface to the `asdf.block` module, which itself is public. Note that unless specifically specified `__all__` is assumed to be all module level objects without a prefixed `_`. Therefore I consider `Block` to be public.

The docstring/comment you listed is not easily accessible without actually diving into the source code, so I think its unreasonable to expect users to know that it should remain private.

> Are there changes that should be made to make this more explicit?

The obvious solution is to make either the `block` module private or the `Block` class private.

asdf/block.py
```diff
@@ -1005,6 +1009,9 @@ def read(self, fd, past_magic=False, validate_checksum=False):
             If `True`, validate the data against the checksum, and
             raise a `ValueError` if the data doesn't match.
         """
+        if self._data_callback is not None:
```
perrygreenfield, line 1012: Shouldn't the docstring for this method also point this out?

perrygreenfield (commented):
Only needs an update to a docstring in my opinion.

perrygreenfield (approved):
LGTM

asdf/block.py
braingram: Thanks! I added some descriptions to the docstring. Let me know if this looks good to you.

PR #1470, opened 2023-03-09
np.product is deprecated in numpy 1.25

pllim:
This is from https://github.com/astropy/astropy/actions/runs/4376653449/jobs/7659080296 for https://github.com/astropy/astropy/pull/14510


braingram:
Thanks!

There appears to be one more usage of `product` here:
https://github.com/asdf-format/asdf/blob/40eb2eaa7830911b0e9dd746a60d94d68e1feef7/asdf/tags/core/ndarray.py#L332

Would you mind adding that change to this PR?

pllim:
Done.

braingram (approved):
Thanks! LGTM when the CI passes.

PR #1473, opened 2023-03-10
add key support to block manager

braingram:
This PR adds an optional `key` argument to `BlockManager.add` and adds `BlockManger.find_or_create_block` to support providing a hashable  `key` when adding a block (different from the default `id(block._data)` used when a key is not provided. This is useful when `id(block._data)` is inconsistent, unavailable, expensive or otherwise inconvenient to use. This is work towards a deferred block system that will add block storage support to the converter.

Note that `BlockManager.__getitem__` was specifically not updated as it would require another local import and provides no additional functionality over the more explicit `find_or_create_block.

### DIFF
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -91,9 +91,9 @@ class BlockManager:
         if block_set is not None:
             if block in block_set:
                 block_set.remove(block)
-                if block._data is not None and id(block._data) in self._data_to_block_mapping:
-                    del self._data_to_block_mapping[id(block._data)]
-
+                for key, blk in list(self._data_to_block_mapping.items()):
+                    if blk is block:
+                        del self._data_to_block_mapping[key]
         else:
             msg = f"Unknown array storage type {block.array_storage}"
             raise ValueError(msg)
--- a/asdf/block.py
+++ b/asdf/block.py
@@ -53,7 +53,7 @@ class BlockManager:
         """
         return sum(len(x) for x in self._block_type_mapping.values())
 
-    def add(self, block):
+    def add(self, block, key=None):
         """
         Add an internal block to the manager.
         """
@@ -63,9 +63,9 @@ class BlockManager:
             # in the middle of the list.
             self.finish_reading_internal_blocks()
 
-        self._add(block)
+        self._add(block, key=key)
 
-    def _add(self, block):
+    def _add(self, block, key=None):
         block_set = self._block_type_mapping.get(block.array_storage, None)
         if block_set is not None:
             if block not in block_set:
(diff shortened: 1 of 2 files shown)


WilliamJamieson:
> On a general note there is a lot of ambiguity about what is considered `public` in ASDF. We've discussed cleaning this up in ASDF 3.0 but there are features that we need to add before the cleanup can occur. This will force us to make compromises and in this instance I'd rather not add a changelog to further confuse users if `BlockManager` is public.

See https://github.com/asdf-format/asdf/pull/1469#issuecomment-1473881294

perrygreenfield:
At the moment, this seems inconsistent with #1468

braingram:
> At the moment, this seems inconsistent with #1468

Thanks! I rebased this to main now that #1468 is merged and fixed the conflicts.

asdf/block.py
```diff
@@ -78,8 +78,10 @@ def _add(self, block):
             msg = "Can not add second streaming block"
             raise ValueError(msg)
 
-        if block._data is not None:
-            self._data_to_block_mapping[id(block._data)] = block
+        if block._data is not None or key is not None:
+            if key is None:
+                key = id(block._data)
+            self._data_to_block_mapping[key] = block
```
perrygreenfield, line 84: Does this imply that if created with a key, that the id cannot be used as a key for that block? May that cause problems with future code that presumes that id should always work?

asdf/block.py
```diff
@@ -747,6 +749,30 @@ def find_or_create_block_for_array(self, arr, ctx):
 
         return block
 
+    def find_or_create_block(self, key, ctx):
```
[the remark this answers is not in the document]
braingram: Thanks for looking for inconsistencies. I believe this was fixed during the rebase. I'm not seeing it when viewing the `files changed` for this PR and the comment shows an 'outdated' label.

asdf/block.py
braingram: Yes that is correct. If a `key` is provided then `id(block._data)` will not be used to generate a `key` and an attempt to get a block using `id(block._data)` will fail (if it does not match the `key`).

Is there a current (or future) use that you think might conflict with this? I'm expecting that when we update ndarray to a converter that we will continue to use `id(block._data)` (equal to `id(base_array)`). The `key` is more useful for objects where the data might not have a consistent location in memory (such as zarr, which reloads the data on each read, creating a new id, and instead a key based on the chunk key is more consistent). Here's an example usage in the `asdf-zarr` prototype: https://github.com/braingram/asdf_zarr/blob/deferred_block/src/asdf_zarr/converter.py#L143

perrygreenfield (approved):
LGTM

WilliamJamieson (approved):

PR #1474, opened 2023-03-12
Move built-in tags to converters (except ndarray and integer)

eslavich:
This moves the following built-in tags to converters:

- tag:stsci.edu:asdf/core/asdf-*
- tag:stsci.edu:asdf/core/complex-*
- tag:stsci.edu:asdf/core/constant-* (whatever that is???)
- tag:stsci.edu:asdf/core/extension_metadata-*
- tag:stsci.edu:asdf/core/externalarray-*
- tag:stsci.edu:asdf/core/history_entry-*
- tag:stsci.edu:asdf/core/software-*
- tag:stsci.edu:asdf/core/subclass_metadata-*

### DIFF
--- a/asdf/_tests/test_versioning.py
+++ b/asdf/_tests/test_versioning.py
@@ -267,4 +267,4 @@ def test_spec_equal():
     assert spec != (1, 1, 0)
     assert (1, 1, 0) != spec
     assert spec == (1, 3, 0)
-    assert (1, 3, 0) == spec  # noqa: SIM300
+    assert (1, 3, 0) == spec
--- a/asdf/core/_converters/complex.py
+++ b/asdf/core/_converters/complex.py
@@ -2,9 +2,8 @@ import re
 
 import numpy as np
 
-from asdf.extension import Converter
 from asdf import util
-
+from asdf.extension import Converter
 
 _REPLACEMENTS = {
     re.compile("i(?!nf)"): "j",


asdf/core/_converters/complex.py
```diff
@@ -0,0 +1,22 @@
...
+
+_REPLACEMENTS = {
+    re.compile("i(?!nf)"): "j",
+    re.compile("I(?!NF)"): "J",
+}
+
+
+class ComplexConverter(Converter):
+    tags = ["tag:stsci.edu:asdf/core/complex-1.0.0"]
+    types = ["builtins.complex", "numpy.complex64", "numpy.complex128", "numpy.complex256"]
```
braingram, line 13: The old code dynamically inspected numpy to get all complex types. I think the explicit list introduced here is more readable. However, on my system I do not have a `complex256` and I do have a `cdoublelong` (returned from `util.iter_subclasses` like in the old code) which appears to be 128 bit but fails isinstance checks against complex128 and complex64.

Also, it looks like this and the previous code did not roundtrip numpy objects. On my system the builtin complex seems capable of representing the largest (complex128) values that numpy supports so this might not be an issues but I wonder if large complex256 will become infs. Does your system support complex256 and can you test this?

compatibility_tests/test_file_compatibility.py
```diff
@@ -26,7 +27,11 @@
 # Minimum library version to produce files read by the current
 # version of the code.  Earlier versions aren't able to generate
 # files for all the ASDF Standard versions that they claim to support.
-MIN_VERSION_OLD_FILES = Version("2.3.0")
+MIN_VERSION_OLD_FILES = Version("2.5.0")
+
+# Minimum library version that is capable of correctly writing
+# files that follow ASDF Standard 1.0.0.
+MIN_VERSION_1_0_0_STANDARD = Version("2.7.0")
```
braingram, line 34: What changes required this restriction?

braingram (changes requested):
Thanks for taking this on!

I added a bunch of questions. My main concerns are over compatibility and API consistency. Functionally things look good (except for maybe complex which looks... complex) but I'd like more information on some of the test removals and compatibility issues.

asdf/_tests/test_api.py
```diff
@@ -529,3 +529,21 @@ def test_none_values(tmp_path):
...
+
+    af = asdf.AsdfFile()
+    af.write_to(buff, version="1.0.0")
+    buff.seek(0)
+    content = buff.read()
+    assert b"!core/asdf-1.0.0" in content
+    assert b"!core/asdf-1.1.0" not in content
+
+    buff.seek(0)
+    af.write_to(buff, version="1.2.0")
```
braingram, line 545: To check I understand this test, is it setting the ASDF standard to 1.0.0 then checking that core tag is version 1.0.0, as listed in the [1.0.0 manifest](https://github.com/asdf-format/asdf-standard/blob/main/src/asdf_standard/resources/manifests/asdf-format.org/core/core-1.0.0.yaml). Then setting the standard to 1.2.0 to check that the correct 1.1.0 core tag is used, as listed in the [1.2.0 manifest](https://github.com/asdf-format/asdf-standard/blob/main/src/asdf_standard/resources/manifests/asdf-format.org/core/core-1.2.0.yaml)?

asdf/_tests/test_versioning.py
```diff
@@ -274,104 +268,3 @@ def test_spec_equal():
...
-        ("1.0.0", "tag:stsci.edu:asdf/wcs/wcs-1.0.0"),
-    ]:
-        request.node.add_marker(
-            pytest.mark.xfail(reason="astropy does not yet explicitly support older schema versions", strict=True),
-        )
-
-
-@pytest.mark.parametrize(("version", "schema_type", "tag"), list(_standard_versioned_tags()))
-@pytest.mark.usefixtures("_xfail_version_map_support_cases")
-def test_version_map_support(version, schema_type, tag):
```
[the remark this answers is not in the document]
braingram: Good point.

Looking at the changes in #1464 it does look like complete support is lacking for all asdf-standard versions. defunit appears to be unsupported by every version so I'm not sure what's going on there and if this test is meant to reveal those types of issues it's obviously failing (as it didn't catch that nothing appears to support defunit).

I do think that the current arrangement (asdf only partially supporting the standard and requiring specific extra packages to be installed to fully support the standard) is confusing. I'll open an issue for this so we can discuss what can/should be done and link back to this conversation. https://github.com/asdf-format/asdf/issues/1511

For now I'm ok with removing this test, we can always re-add it if we decided that supporting the standard with our test dependencies is a useful target.

asdf/_tests/test_yaml.py
```diff
@@ -201,7 +201,7 @@ def check_asdf(asdf):
 def test_explicit_tags():
     yaml = f"""#ASDF {asdf.versioning.default_version}
 %YAML 1.1
---- !<tag:stsci.edu:asdf/core/asdf-1.0.0>
```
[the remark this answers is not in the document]
braingram: Thanks!

asdf/core/_converters/complex.py
braingram: Given the varied support across machines I'm thinking matching the old type (using `iter_subclasses`) is preferable so that we don't miss types.

Your point about tightening up the standard is a good one and deserves a separate issue. Do you think it makes sense to open one in ASDF or ASDF-standard?

Failure to roundtrip that preserves precision seems fine to me, failures to roundtrip that lose precision (and/or convert numbers to inf/nan) seem unacceptable. I'm all ears for ideas on how to best handle this and must admit that I don't fully grasp all the places where precision loss can occur (on write/conversion to new type, then to yaml, on read/conversion from yaml then to new type, ect).

asdf/tags/core/__init__.py
```diff
@@ -1,20 +1,16 @@
...
 from .ndarray import NDArrayType
 
 __all__ = [
     "AsdfObject",
+    "Constant",
     "Software",
     "HistoryEntry",
     "ExtensionMetadata",
     "SubclassMetadata",
-    "ConstantType",
```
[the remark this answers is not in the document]
braingram: All great points. I'm good with removing them for the reasons you described.

compatibility_tests/test_file_compatibility.py
braingram: Thanks! How about adding another line to the comment with something like:

```suggestion
# ASDF library versions older than this version would sometimes write newer
# tags that were not listed in the supported standard
MIN_VERSION_1_0_0_STANDARD = Version("2.7.0")
```

braingram (changes requested):
Thanks for updating this and apologies again for not seeing the update.
I added a few more requests:
- adding a comment to describe the `MIN_VERSION_1_0_0_STANDARD` addition
- possibly switching complex types back to using `iter_subclasses`
Otherwise the comments are mostly questions to make sure I'm understanding the changes.
Feel free to mark me again for a review request when this is ready for another look.

braingram (approved):
LGTM! Out of an abundance of caution I added the downstream label. I don't expect it to fail but we might as well see how it fairs.

PR #1475, opened 2023-03-12
Move large integer support to Converter and deprecate IntegerType

eslavich:
I separated this from the others in #1474 because it's a little more involved.  The Converter implementation is able to automatically detect and handle integers that are too large for the standard's limit, which means the IntegerType wrapper can be deprecated.

### DIFF
--- a/asdf/_tests/core/_converters/test_integer.py
+++ b/asdf/_tests/core/_converters/test_integer.py
@@ -11,7 +11,7 @@ from asdf.testing import helpers
 
 
 @pytest.fixture(autouse=True)
-def set_random_seed():  # noqa: PT004
+def set_random_seed():
     # Make sure tests are deterministic
     random.seed(0)
 


asdf/core/_converters/integer.py
```diff
@@ -0,0 +1,67 @@
...
+from asdf.tags.core import IntegerType
+
+
+class IntegerConverter(Converter):
+    tags = [
+        "tag:stsci.edu:asdf/core/integer-1.0.0",
+        "tag:stsci.edu:asdf/core/integer-1.1.0",
+    ]
+
+    types = [
```
braingram, line 14: Related to https://github.com/asdf-format/asdf/pull/1474#discussion_r1138969192

I wonder if this PR and #1474 might benefit from a more flexible way of matching types to converters. I have not looked at this recently (so please correct any mistakes  I make as I'm typing from memory) but I believe the types are only used during conversion from a custom to tagged tree (for writing or validation). If the goal of this converter is (in part) to support all 'inty' types from numpy, inspecting those at runtime should be fine (given that we only need to support the objects that are supported by the system running this code).

One other comment and thing to discuss is that this doesn't round trip types. For example:
```python
t = {'a': numpy.uint8(255), 'b': numpy.uint8(255)}
rt = asdf.testing.helpers.roundtrip_object(tree)
rt['a'] + rt['b'] == 510
t['a'] + t['b'] == 254   # and a RuntimeWarning: overflow encountered in scalar add
```

asdf/tags/core/integer.py
```diff
@@ -16,87 +15,38 @@ class IntegerType(_types.AsdfType):
...
+        warnings.warn(
+            "IntegerType is deprecated.  Large integer values are now handled automatically "
+            "and need no special wrapper class.  Use asdf.get_config().array_inline_threshold "
+            "to customize the array storage type.",
+            AsdfDeprecationWarning,
+        )
+
+        if storage_type is not None:
+            warnings.warn(
+                "The storage_type argument to IntegerType is now ignored.  Use "
```
braingram, line 43: Would it be possible to retain this feature with the changes you added to the deferred block prototype (https://github.com/braingram/asdf/commit/90ddacc0a51df30bd348c99cb5597e72a14a58dd)?
I have not yet opened a PR to bring over those changes because I was going to include them with the 'adding block storage to the converter' PR that depends on #1468 #1469 #1473. If it would allow for this feature than perhaps we should create a PR for that feature (adding storage/compression settings for non-ndarray objects) first.

asdf/tags/core/integer.py
```diff
@@ -111,7 +61,7 @@ def __eq__(self, other):
         if isinstance(other, IntegerType):
             return self._value == other._value
 
-        msg = f"Can't compare IntegralType to unknown type: {type(other)}"
+        msg = f"Can't compare IntegerType to unknown type: {type(other)}"
```
braingram, line 64: Thanks for catching the typo. Would you add a test that covers this line?

braingram (changes requested):
There are some great changes in this PR (both in scope and in impact). I added a bunch of comments/questions.

PR #1476, opened 2023-03-13
Deprecate AsdfFile.extensions and the 'extensions' argument to AsdfFile and asdf.open

eslavich:
The AsdfConfig object provides a way to add extensions at runtime, so we don't need to continue to allow custom extensions on the AsdfFile.

### DIFF
--- a/asdf/_tests/tags/core/tests/test_ndarray.py
+++ b/asdf/_tests/tags/core/tests/test_ndarray.py
@@ -54,7 +54,7 @@ class CustomExtension:
 
 
 @pytest.fixture(autouse=True)
-def configure_custom_extension():
+def configure_custom_extension():  # noqa: PT004
     with asdf.config_context() as config:
         config.add_extension(CustomExtension())
         yield
--- a/asdf/_tests/test_schema.py
+++ b/asdf/_tests/test_schema.py
@@ -535,8 +535,8 @@ custom: !<tag:nowhere.org:custom/default-1.0.0>
             assert ff.tree["custom"]["j"]["l"] == 362
 
         buff.seek(0)
-        with config_context() as config:
-            config.legacy_fill_schema_defaults = False
+        with config_context() as config2:
+            config2.legacy_fill_schema_defaults = False
             with asdf.open(buff) as ff:
                 assert "a" not in ff.tree["custom"]
                 assert "c" not in ff.tree["custom"]["b"]


pyproject.toml
```diff
@@ -114,6 +114,7 @@ remote_data_strict = true
 filterwarnings = [
     'error',
     'ignore:numpy.ndarray size changed:RuntimeWarning',
+    'ignore:.*extensions.*deprecated:asdf.exceptions.AsdfDeprecationWarning',
```
braingram, line 117: Presumably this is needed in part because the test code uses the deprecated feature. Here's one example:
https://github.com/asdf-format/asdf/blob/19d2cbb4486559cd5b44fa400823456d9f6bd05c/asdf/_tests/test_schema.py#L307

As these deprecated uses will need to be fixed when the feature is removed it seems preferable to make these changes as part of this PR and to remove this general warning filter.

braingram (changes requested):

asdf/asdf.py
```diff
@@ -69,6 +69,7 @@ def __init__(
             if possible and if created from `asdf.AsdfFile.open`.
 
         extensions : object, optional
+            DEPRECATED
```
braingram, line 72: Is there a concise link we can add here that will render in the docs, something like (I'm not sure if this will render right so the docs should be checked if this suggestion is used).
```suggestion
            DEPRECATED See :ref:`extending_extensions_installing_asdf_config`
```

asdf/asdf.py
```diff
@@ -119,6 +120,13 @@ def __init__(
         else:
             self._version = versioning.AsdfVersion(validate_version(version))
 
+        if extensions is not None:
+            warnings.warn(
+                "The 'extensions' argument to AsdfFile is deprecated.  Use asdf.get_config().add_extension to "
+                "add an extension at runtime.",
+                AsdfDeprecationWarning,
```
braingram, line 127: Similar to the other suggestion:
```suggestion
                "The 'extensions' argument to AsdfFile is deprecated.  Please see :ref:`extending_extensions_installing_asdf_config`.",
                AsdfDeprecationWarning,
```

asdf/asdf.py
```diff
@@ -226,6 +234,15 @@ def extensions(self):
         -------
         list of asdf.extension.ExtensionProxy
         """
+        warnings.warn(
+            "AsdfFile.extensions is deprecated.  Use asdf.get_config().extensions to "
+            "get a list of enabled extensions.",
+            AsdfDeprecationWarning,
```
braingram, line 240: Similar to other suggestions.
```suggestion
            "AsdfFile.extensions is deprecated.  Please see :ref:`extending_extensions_installing_asdf_config`.",
            AsdfDeprecationWarning,
```

asdf/asdf.py
```diff
@@ -238,6 +255,15 @@ def extensions(self, value):
         ----------
         value : list of asdf.extension.AsdfExtension or asdf.extension.Extension
         """
+        warnings.warn(
+            "AsdfFile.extensions is deprecated.  Use asdf.get_config().add_extension to "
+            "add an extension at runtime.",
+            AsdfDeprecationWarning,
```
braingram, line 261: ditto
```suggestion
            "AsdfFile.extensions is deprecated.  Please see :ref:`extending_extensions_installing_asdf_config`.",
            AsdfDeprecationWarning,
```

asdf/asdf.py
```diff
@@ -1863,6 +1890,13 @@ def open_asdf(
     if not _compat:
         mode = _check_and_set_mode(fd, mode)
 
+    if extensions is not None:
+        warnings.warn(
+            "The 'extensions' argument to asdf.open is deprecated.  Use asdf.get_config().add_extension to "
+            "add an extension at runtime.",
+            AsdfDeprecationWarning,
```
braingram, line 1897: ditto
```suggestion
            "The 'extensions' argument to asdf.open is deprecated. Please see :ref:`extending_extensions_installing_asdf_config`.",
            AsdfDeprecationWarning,
```

braingram (changes requested):
Overall looks good. I made some minor warning message suggestions (adding what should be a link to the docs).

I also scanned the downstream run logs for more `AsdfDeprecationWarning` messages and don't see any new ones (there are some in WeldX from other deprecations which they said they'll fix once 2.15 is released and some in JWST because of the unreleased changes in stpipe). Are there any uses for these arguments/methods that you're aware of?

PR #1477, opened 2023-03-13
Deprecate flags that control warnings in AsdfFile and asdf.open

eslavich:
We have these flags that we propagate around that control behavior of certain warnings:

- ignore_version_mismatch
- ignore_unrecognized_tag
- ignore_implicit_conversion
- strict_extension_check
- ignore_missing_extensions

but there is a cleaner and more conventional way: Python warning filters.  For example, any of the "ignore" flags can be replaced with an ignore filter:

```python
warnings.simplefilter("ignore", AsdfConversionWarning)
```

The `strict_extension_check` flag can be replaced with an error filter:
```python
warnings.simplefilter("error", AsdfMissingExtensionWarning)
```

The `ignore_version_mismatch` flag can be dropped with no replacement, since Converter does not fall back to earlier versions of a tag.

The `ignore_implicit_conversion` warning should be an error, I think, since when it occurs we are losing the original class of an immutable object in the process of a tree conversion.  It seems to me that we should ask the user to deal with that rather than assume that converting a tuple object to a list is a reasonable thing to do.

I added the `AsdfMissingExtensionWarning` class so that users can easily filter on it.  This change should be backwards compatible because `AsdfMissingExtensionWarning` subclasses the original warning class (`AsdfWarning`).

### DIFF
--- a/pytest_asdf/plugin.py
+++ b/pytest_asdf/plugin.py
@@ -37,15 +37,15 @@ def pytest_addoption(parser):
     )
     parser.addini(
         "asdf_schema_ignore_unrecognized_tag",
-        "Set to true to disable warnings when tag serializers are missing",
+        "DEPRECATED. Set to true to disable warnings when tag serializers are missing",
         type="bool",
-        default=False,
+        default=None,
     )
     parser.addini(
         "asdf_schema_ignore_version_mismatch",
-        "Set to true to disable warnings when missing explicit support for a tag",
+        "DEPRECATED. Set to true to disable warnings when missing explicit support for a tag",
         type="bool",
-        default=True,
+        default=None,
     )
     parser.addoption("--asdf-tests", action="store_true", help="Enable ASDF schema tests")
 
@@ -59,8 +59,8 @@ class AsdfSchemaFile(pytest.File):
         fspath,
         skip_examples=False,
         validate_default=True,
-        ignore_unrecognized_tag=False,
-        ignore_version_mismatch=False,
+        ignore_unrecognized_tag=None,
+        ignore_version_mismatch=None,
         skip_tests=None,
         xfail_tests=None,
         **kwargs,
@@ -201,8 +201,8 @@ class AsdfSchemaExampleItem(pytest.Item):
         schema_path,
         example,
         example_index,
-        ignore_unrecognized_tag=False,
-        ignore_version_mismatch=False,
+        ignore_unrecognized_tag=None,
+        ignore_version_mismatch=None,
         **kwargs,
     ):
         if hasattr(super(), "from_parent"):
@@ -302,7 +302,12 @@ def pytest_collect_file(path, parent):
     skip_examples = parent.config.getini("asdf_schema_skip_examples")
     validate_default = parent.config.getini("asdf_schema_validate_default")
     ignore_unrecognized_tag = parent.config.getini("asdf_schema_ignore_unrecognized_tag")
+    if ignore_unrecognized_tag == []:
+        ignore_unrecognized_tag = None
+
     ignore_version_mismatch = parent.config.getini("asdf_schema_ignore_version_mismatch")
+    if ignore_version_mismatch == []:
+        ignore_version_mismatch = None
 
     skip_tests = _parse_test_list(parent.config.getini("asdf_schema_skip_tests"))
     xfail_tests = _parse_test_list(parent.config.getini("asdf_schema_xfail_tests"))


WilliamJamieson (changes requested):
These changes are incomplete. These flags are in use by the `pytest-asdf` plugin. This is what is causing many of the failures in the downstream tests.

Please update the `pytest-asdf` plugin to not use any of these flags.

asdf/asdf.py
```diff
@@ -112,6 +121,29 @@ def __init__(
...
+            warnings.warn(
+                "The ignore_version_mismatch argument to AsdfFile will be removed in 3.0.",
+                AsdfDeprecationWarning,
+            )
+        else:
+            ignore_version_mismatch = True
+
+        if ignore_unrecognized_tag is not None:
+            warnings.warn(
+                "The ignore_unrecognized_tag argument to AsdfFile will be removed in 3.0. "
```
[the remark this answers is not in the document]
braingram: Is there migration documentation this can link to? Also, I would say let's not commit to a version for removal here and instead just say it's deprecated.

asdf/asdf.py
```diff
@@ -344,7 +376,8 @@ def _check_extensions(self, tree, strict=False):
                 if Version(installed.package_version) < Version(extension.software["version"]):
                     msg = (
                         f"File {filename}was created with extension {extension_description}, "
-                        f"but older package ({installed.package_name}=={installed.package_version}) is installed."
+                        f"but older package ({installed.package_name}=={installed.package_version}) is installed. "
+                        "This warning will be removed in 3.0."
```
braingram, line 380: Can we keep this warning and just not force it to be turned into an error? We could add another `AsdfWarning` subclass (perhaps `AsdfExtensionVersionWarning`) and link to the documentation on how to convert these warnings into errors.

asdf/asdf.py
```diff
@@ -861,11 +894,29 @@ def _open_asdf(
...
-        strict_extension_check=False,
-        ignore_missing_extensions=False,
+        strict_extension_check=None,
+        ignore_missing_extensions=None,
         **kwargs,
     ):
         """Attempt to populate AsdfFile data from file-like object"""
+        if strict_extension_check is not None:
+            warnings.warn(
+                "The strict_extension_check argument to asdf.open will be removed in 3.0. "
```
braingram, line 904: Same as above about a migration link and removing the version.

asdf/asdf.py
```diff
@@ -861,11 +894,29 @@ def _open_asdf(
...
+                "To raise an exception when an extension is missing, use a warning filter "
+                "to convert asdf.exceptions.AsdfMissingExtensionWarning to an error.",
+                AsdfDeprecationWarning,
+            )
+        else:
+            strict_extension_check = False
+
+        if ignore_missing_extensions is not None:
+            warnings.warn(
+                "The ignore_missing_extensions argument to asdf.open will be removed in 3.0. "
```
braingram, line 914: Same here about link and removing version.

pytest_asdf/plugin.py
```diff
@@ -59,8 +59,8 @@ def from_parent(
         fspath,
         skip_examples=False,
         validate_default=True,
-        ignore_unrecognized_tag=False,
-        ignore_version_mismatch=False,
```
braingram, line 63: This doesn't match the default so switching to None will change this behavior:
https://github.com/asdf-format/asdf/blob/b8e1759aa9f013283942140c94732a170eb9a703/asdf/asdf.py#L1772
I'm not immediately understanding the context so does this change matter?

pytest_asdf/plugin.py
```diff
@@ -37,15 +37,15 @@ def pytest_addoption(parser):
...
-        "Set to true to disable warnings when tag serializers are missing",
+        "DEPRECATED. Set to true to disable warnings when tag serializers are missing",
         type="bool",
-        default=False,
+        default=None,
     )
     parser.addini(
         "asdf_schema_ignore_version_mismatch",
-        "Set to true to disable warnings when missing explicit support for a tag",
+        "DEPRECATED. Set to true to disable warnings when missing explicit support for a tag",
```
braingram, line 46: Can we link to documentation here?

braingram (changes requested):
A quick look at the downstream errors suggests they're all related to new warnings being turned into errors.
https://github.com/asdf-format/asdf/actions/runs/4410156009/jobs/7727325251?pr=1477#step:10:656
https://github.com/asdf-format/asdf/actions/runs/4407455592/jobs/7721154361#step:9:413

There are uses of these arguments in the downstream (here are some of the ones in stdatamodels: https://github.com/spacetelescope/stdatamodels/search?q=ignore_version_mismatch). Are these all compatible with the changes in this PR? If there's no urgency to this PR would you be willing to open PRs to fix the downstream uses first?

PR #1479, opened 2023-03-13
rename master branch to main

braingram:
After this PR is merged the `master` branch can be renamed to `main`.

We should probably merge https://github.com/asdf-format/asdf-standard/pull/370 first as `asdf-standard` has a hard coded reference to `master`. It's possible github will redirect these (based on https://github.com/github/renaming, but i'm not sure if the ci jobs are considered "web requests").

### DIFF
--- a/docs/asdf/developer_versioning.rst
+++ b/docs/asdf/developer_versioning.rst
@@ -169,14 +169,14 @@ Update the asdf-standard submodule commit pointer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The asdf-standard repository is integrated into the asdf repository
-as a submodule.  To pull in new commits from the remote master (
+as a submodule.  To pull in new commits from the remote main (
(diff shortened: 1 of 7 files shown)


docs/asdf/developer_versioning.rst
```diff
@@ -169,14 +169,14 @@ Update the asdf-standard submodule commit pointer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The asdf-standard repository is integrated into the asdf repository
-as a submodule.  To pull in new commits from the remote master (
+as a submodule.  To pull in new commits from the remote main (
```
WilliamJamieson, line 172: Can you open a matching PR in `asdf-standard`? we can then merge them right after one another and update the branch names after that.

braingram: Sounds good.
https://github.com/asdf-format/asdf-standard/pull/371

WilliamJamieson (approved):

PR #1480, opened 2023-03-13
Bugfix for docs on python 3.11

WilliamJamieson:
`tomli` usage combine with `tomli` pin is incorrect for `python 3.11`. Also, switch to using conda and `python 3.11` for RTD build.


zacharyburnett (approved):
LGTM, good to move docs to 3.11

Issue #1486, opened 2023-03-15
remove upper pin from jsonschema and re-add to devdeps


braingram:
Changes scheduled for jsonschema 4.18 are incompatible with ASDF as described in this issue:
https://github.com/asdf-format/asdf/issues/1485

When those issues are resolved, the upper pin for jsonschema can be removed and it can be re-added to the devdeps jobs.

Issue #1485, opened 2023-03-14
jsonschema 4.18 incompatible with asdf


braingram:
Changes to ref resolution introduced here:
https://github.com/python-jsonschema/jsonschema/pull/1049

are incompatible with asdf uses of jsonschema because of deprecations and apparently further changes.

Ignoring the test failures due to warnings from deprecation of the following:
 - `jsonschema.exceptions.RefResolutionError`
 - `jsonschema.validators.RefResolver`
 - `Validator.resolver`
(by filtering the deprecation warnings) still results in numerous test failures.

```
FAILED asdf/_tests/test_resource.py::test_jsonschema_resource_mapping - FileNotFoundError: [Errno 2] No such file or directory: '/Users/bgraham/projects/230314_jsonschema_ref_resolver/jsonschema/jsonschema/schemas/draft4.json'
FAILED asdf/_tests/test_schema.py::test_load_schema - jsonschema.exceptions._RefResolutionError: Unresolvable JSON pointer: 'definitions/stringArray'
FAILED asdf/_tests/test_schema.py::test_load_schema_with_full_tag - jsonschema.exceptions._RefResolutionError: Unresolvable JSON pointer: 'definitions/stringArray'
FAILED asdf/_tests/test_api.py::test_overwrite - asdf.exceptions.AsdfWarning: Unable to locate schema file for 'tag:stsci.edu:asdf/transform/affine-1.3.0': 'http://stsci.edu/schemas/asdf/transform/affine-1.3.0'
FAILED asdf/_tests/test_schema.py::test_load_schema_with_tag_address - jsonschema.exceptions._RefResolutionError: Unresolvable JSON pointer: 'definitions/stringArray'
FAILED asdf/_tests/test_schema.py::test_load_schema_with_file_url - jsonschema.exceptions._RefResolutionError: Unresolvable JSON pointer: 'definitions/stringArray'
FAILED asdf/_tests/test_schema.py::test_load_schema_with_asdf_uri_scheme - jsonschema.exceptions._RefResolutionError: Unresolvable JSON pointer: 'definitions/local_bar'
FAILED asdf/_tests/test_schema.py::test_nested_array_yaml - jsonschema.exceptions._RefResolutionError: Unresolvable JSON pointer: 'definitions/simpleTypes'
FAILED asdf/_tests/test_types.py::test_custom_reference_cycle - RecursionError: maximum recursion depth exceeded while calling a Python object
[...]

PR #1487, opened 2023-03-15
add upper pin to jsonschema

braingram:
jsonschema 4.18 will introduce changes that are incompatible with asdf. More details can be found in:
https://github.com/asdf-format/asdf/issues/1485

this commit adds an upper pin to jsoncschema and removes it from the devdeps testing

As described in the issue, adding a warning filter was insufficient to allow the devdeps jobs to pass.

An issue was created to track removing the pin:  https://github.com/asdf-format/asdf/issues/1486


WilliamJamieson (changes requested):
I am normally against commenting things out instead of removing them.

In this case, I want it clearly documented why we have stopped the testing against jsonschema dev and why we have introduced an upper pin within the code itself. Otherwise, it becomes unclear when examining the code at first glance why these changes have been made and the intent for them to be temporary changes.

braingram:
> Should an issue with `jsonschema` be opened about this? That clearly indicates the changes are not backwards compatible.

I am suspicious that something is up with the reference resolution in meta schemas but so far I've been unable to come up with a non-asdf minimal case that fails. Given the extensive use of non-public portions of jsonschema within asdf it's possible the changes did not introduce any public backwards incompatibility and the issues we're encountering are our own fault. I'd rather wait to open an issue until we can determine that jsonschema is at fault.

WilliamJamieson (approved):
Hopefully these changes will be temporary

pllim:
I vaguely remember a maxversion pin of jsonschema in the past, and eventually that gave astropy problems when something else required a new jsonschema that asdf cannot use. So hopefully this won't stay forever. 🤞

WilliamJamieson:
@meeseeksdev backport to 2.14.x

PR #1489, opened 2023-03-17
Backport PR #1487: add upper pin to jsonschema

WilliamJamieson:
Backport of #1487 for jsonschema pin


WilliamJamieson:
The weldx and stdatamodels failures look expected (this CI has not been run in awhile).

.github/workflows/ci.yml
```diff
@@ -88,12 +88,12 @@ jobs:
         include:
           - package_name: asdf-standard
             repository: asdf-format/asdf-standard
-            ref: master
+            ref: main
```
braingram, line 91: These can be deleted rather than changed to `main`.

braingram (approved):
One minor comment on the default branch names. Thanks for the quick response.

.github/workflows/ci.yml
WilliamJamieson: False, more changes than just removing the ref are required. The workflow later references this.

I'd rather minimize the changes to what is actually the topic of this PR, namely backporting a pin for jsonschema, rather than refactoring a bunch of unrelated things. The main goal with the other changes is to simply get the CI working well enough to establish that this doesn't break ASDF.

braingram: True... you are :)

Leaving out those changes sounds good. The devdeps errors look to be what was solved in https://github.com/asdf-format/asdf/pull/1470

braingram:
@WilliamJamieson Is this ready to merge? @pllim is asking about a 2.14 release with this to deal with astropy CI failures.

PR #1490, opened 2023-03-19
Attempt to use only public jsonschema interfaces

eslavich:
Just checking if this cuts the mustard with our downstream packages, stay tuned...

### DIFF
--- a/asdf/schema.py
+++ b/asdf/schema.py
@@ -425,6 +425,7 @@ def _make_jsonschema_resolver_or_registry(url_mapping):
 
     if _USE_REFERENCING:
 
+        @lru_cache
         def retrieve_schema(url):
             schema = schema_loader(url)[0]
             return referencing.Resource(schema, specification=referencing.jsonschema.DRAFT4)
@@ -615,6 +616,7 @@ class _Validator:
         visit_repeat_nodes,
         resolver=None,
         registry=None,
+        instance_checks=None,
     ):
         self._ctx = ctx
         self._serialization_context = serialization_context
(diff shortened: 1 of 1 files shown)


WilliamJamieson:
@eslavich this looks like it will work really well, and the changes look nice at first glance. However, I have not had a chance to look over the changes in detail. 

Since you are still testing this could you remove the jsonschema upper pin? It would be nice to go ahead and fix that issue.

braingram:
I think the asdf-astropy (and possibly the dkist errors) will be fixed by:
https://github.com/asdf-format/asdf-standard/pull/374
https://github.com/asdf-format/asdf-transform-schemas/pull/88
I also opened a PR against your branch with:
- a minor change to improve performance during schema validation with 4.18
- re-enabling jsonschema 4.18 as an allowable dependency, re-adding it to devdeps
- adding a devdeps coverage job (to allow the per-jsonschema branches to be tested)

To summarize things so far, it looks like this PR should allow compatibility with jsonschema 4.18. However, the more strict schema parsing in this newer version means that asdf will be unable to validate using schema for the current released versions of:
- asdf-transform-schema
- asdf-standard

There are also the issues you opened in jsonschema that should hopefully be addressed before we merge these changes:
https://github.com/python-jsonschema/jsonschema/issues/1062
https://github.com/python-jsonschema/jsonschema/issues/1061
Specifically https://github.com/python-jsonschema/jsonschema/issues/1061 should allow us to avoid overwriting the private `_resolver` attribute: https://github.com/eslavich/asdf/blob/eslavich-no-patch-iter-errors/asdf/schema.py#L656

We are also now depending on `referencing` and `jsonschema_specifications` (both new dependencies of jsonschema 4.18) and should likely add those to the pyproject.toml (I'm inclined to wait on this until a release version of jsonschema 4.18 to see what the lower pins are at that time).

PR #1495, opened 2023-03-22
drop isort as ruff sorts imports

braingram:
ruff sorts imports:
https://beta.ruff.rs/docs/rules/#isort-i

and currently does so in way that can be inconsistent with isort. Ruff will fix to:
```python
from . import _fizz, bam
from . import _foo as foo

_fizz()
foo()
bam()
```
whereas isort will fix to:
```python
from . import _fizz
from . import _foo as foo
from . import bam

_fizz()
foo()
bam()
```

This PR drops the use of isort.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,11 +47,6 @@ repos:
     - id: ruff
       args: ["--fix"]
 
-- repo: https://github.com/pycqa/isort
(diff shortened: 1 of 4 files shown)


WilliamJamieson:
This should have been done as part of #1299. I guess the tools agreed until recently on all our code.

PR #1500, opened 2023-03-24
Initial benchmarks

braingram:
This is the start of some benchmarks for ASDF to be integrated with https://github.com/spacetelescope/bench

The approach taken here is to include the benchmarks in this repo (rather than have them in a separate repo like [astropy](https://github.com/astropy/astropy-benchmarks/)). This approach was taken to hopefully allow us to keep the benchmarks up-to-date with all the expected 3.0+ API changes.

The general benchmark layout attempts to follow the source layout (`benchmarks/asdf.py` contains benchmarks related to `asdf/asdf.py`) to make associations between functions and benchmarks more obvious.

I think we should feel free to mess with the structure/format/conventions while we are becoming familiar with the tools and how best to integrate this with ASDF.

I will call out some questions I have by adding comments to lines of code.

I would also be curious to hear if there are ways this might integrate with the ASDF CI. Specifically, does it make sense to test the benchmarks (to make sure they follow any code updates, perhaps by adding a `benchmarks` label like `downstream-ci` so we don't run potentially expensive benchmarks on every CI job).


benchmarks/asdf.py
```diff
@@ -0,0 +1,82 @@
...
+        self.asdf_file = asdf.AsdfFile(_utils.build_tree(*key.split('_')))
+
+    def time_AsdfFile_validate(self, key):
+        af = self.asdf_file.validate()
+
+    def peakmem_AsdfFile(self, key):
+        # peakmem includes setup, this is for comparison
+        pass
+
+    def peakmem_AsdfFile_validate(self, key):
```
braingram, line 38: I'm not sure I fully understand `peakmem` vs `mem`. In this instance, I want to test how much memory is consumed during a call to `AsdfFile.validate`. Initially I though to use `mem` but this appears to measure memory of a specific object (which doesn't fit this use). `peakmem` seemed like a better fit but includes the memory consumed during `setup`. Next I thought to use `setup_cache` as that doesn't appear to get included with `peakmem`. However, `setup_cache` can't be parametrized. In the end I added a comparison benchmark `peakmem_AsdfFile` that does nothing and should provide a measurement of memory consumed during `setup`. Is there a better way to accomplish this? It would be nice if there was a way to snapshot the memory used and only look at a delta but I found no way to do this in asv.

zacharyburnett: I'm not as familiar with `asv`'s memory benchmarking, but from what you said it looks like the functionality isn't quite what is needed here. We could also try `memray` or something to that effect.

braingram: That's a great idea!

One thing I'd like to try to catch are issues where a function call (or class instantiation etc) creates objects that are difficult to garbage collect (or are erroneously cached) and create ever increasing memory usage (like https://github.com/spacetelescope/stdatamodels/pull/109). I think `memray` would be helpful for catching these.

zacharyburnett:
I made a PR at `bench`: https://github.com/spacetelescope/bench/pull/11

asv.conf.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": 1,
+    "project": "asdf",
+    "project_url": "https://github.com/asdf-format/asdf",
+    "repo": ".",
+    "branches": [
+        "main"
+    ],
+    "environment_type": "virtualenv",
```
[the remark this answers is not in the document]
zacharyburnett: I don't think so, it should be fine. The `bench` runner is currently running within a Conda environment itself, so it has access to the latest `virtualenv` to use.

braingram:
@zacharyburnett I added a few more benchmarks that I think should serve as a reasonable initial set. Let me know if there are any issues or other benchmarks that might be useful. Thanks!

zacharyburnett:
We can definitely merge this (and the corresponding `bench` PR) in the meantime, as well as run these benchmarks manually on the benchmarking machine.

braingram:
Good to know.

I also saw mention of and added a 'timeraw' benchmarks which I think should let us measure some of the one-time code that runs when ASDF looks for extensions, schema, etc on startup:
https://asv.readthedocs.io/en/stable/writing_benchmarks.html#raw-timing-benchmarks

PR #1518, opened 2023-04-11
limit ruff rules

braingram:
Prior to this PR ASDF included all new ruff rules. The ruff version in kept up-to-date with weekly pre-commit automatic updates. These updates often include new rules that were automatically enabled because of the 'all' rule selection setting. These frequent style updates will cause the code in the main branch to quickly diverge from the older (2.15) branch. The pace of ruff development currently outpaces ASDF and ASDF PRs often fall stale due to style issues after a few weeks.

This PR drops the 'all' rule selection setting and replaces it with a set of rules that matches ASDF prior to the use of ruff.
Using this commit as a reference:
https://github.com/asdf-format/asdf/commit/40131769b51b145dd25e79ab4a2d8ede10604036
ASDF has dropped the following tools, replacing them with ruff:
- yesqa: replaced with "RUF" rules
- pyupgrade: replaced with "UP" rules
- autoflake8 and pyflakes: replaced with "F" rules
- isort: replaced with "I" rules
- bandit: replaced with "S" rules
- pycodestyle: was not previously used, but the "E" rules are default enabled for ruff so I'm including them here with the expectation that they will be stable and require infrequent changes

### DIFF
--- a/asdf/__init__.py
+++ b/asdf/__init__.py
@@ -25,7 +25,7 @@ from ._convenience import info
 from ._types import CustomType
 from ._version import version as __version__
 from .asdf import AsdfFile
-from .asdf import open_asdf as open  # noqa: A001
+from .asdf import open_asdf as open
 from .config import config_context, get_config
 from .stream import Stream
 from .tags.core import IntegerType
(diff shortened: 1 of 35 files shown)


zacharyburnett (approved):
LGTM, it makes sense to freeze our `select` until ruff has a more stable ruleset

braingram:

Issue #1347, opened 2023-02-03
Remove IntegerType value caching


braingram:
IntegerType will cache and reuse arrays on a per-AsdfFile basis to avoid storing duplicate blocks for >1 IntegerTypes that share the same value.

This feature is unused in the coverage survey of downstream packages and we should consider deprecating it for removal:
https://github.com/asdf-format/asdf/blob/7862592cd14dd1e121b3f8fc8c751c2fdedccdd6/asdf/tags/core/integer.py#L60-L68

braingram:
I'm failing to find any documentation of this behavior. I'm closing this issue as it appears to be an internal implementation detail that can be changed with no effect on user code and doesn't require a deprecation warning prior to removal.

PR #1527, opened 2023-04-23
Move IntegerType to converter and drop cache of converted values

eslavich:
This is mostly done, the only missing piece is a way to set the IntegerType's array storage within Converter.


braingram:
> This needs to be redone, I believe we decided to keep the cache, and the converter will need to be implemented using the new select_tag feature.

I'd say we ditch the cache: https://github.com/asdf-format/asdf/issues/1347

PR #1528, opened 2023-04-24
skip astropy dev testing for python 3.8

braingram:
CI for a previous PR revealed that the python 3.8 devdeps job failed due to astropy dropping support for 3.8:
https://github.com/asdf-format/asdf/actions/runs/4789902511/jobs/8518523706

This PR removes astropy from python 3.8 devdeps testing (and continues to use the development version of astropy for python 3.9+).


zacharyburnett (approved):
LGTM, perhaps we can also put the schedule of Python version deprecations on a central page in the `spacetelescope` org as well for visibility (or somewhere like `ins-jwst-community-software`)

braingram:
> LGTM, perhaps we can also put the schedule of Python version deprecations on a central page in the `spacetelescope` org as well for visibility (or somewhere like `ins-jwst-community-software`)

Good idea! Thanks :)

PR #1532, opened 2023-04-27
drop depending on astropy.io.misc.asdf for test

braingram:
Astropy removed astropy.io.misc.asdf in their development version.
https://github.com/astropy/astropy/pull/14668

ASDF has one test that silently depended on the astropy.io.misc.asdf submodule to provide an equality function for an AffineTransform.

The PR removing the public API for the legacy extension already deals with this issue but is unlikely to be quickly merged: https://github.com/asdf-format/asdf/pull/1464

This PR includes the updated test from #1464 to not depend on the types provided by the now non-existent astropy.io.misc.asdf which is causing devdeps jobs to fail (as seen in CI runs for this PR https://github.com/asdf-format/asdf/pull/1531).


braingram:
Closing and opening to retrigger docs build.

nden (approved):

braingram:
Thanks @nden for the review!

PR #1544, opened 2023-05-09
Revert "Remove legacy extension api"

braingram:
Reverts asdf-format/asdf#1464

Astropy is testing against our 3.0 development version. A few tests in their v5.0, v5.2 and v5.3 branches have not been updated.
https://github.com/astropy/astropy/issues/14795

Until those are fixed it's likely easiest to revert the changes here to avoid having them pin ASDF
https://github.com/astropy/astropy/pull/14796


braingram:
Converted to draft to try alternative fix as discussed at: https://github.com/astropy/astropy/issues/14795

PR #1556, opened 2023-06-02
drop python 3.8 support and update CI

braingram:
Use new nightly wheels repository: https://anaconda.org/scientific-python-nightly-wheels/repo

drop testing with development scipy (which is not a dependency)


nden (approved):
lgtm

Issue #1552, opened 2023-05-24
Failure to load array data from files opened with s3fs


braingram:
Tested with the current asdf main branch and a mock s3 the following script fails:

```python
import asdf
import boto3
from moto.server import ThreadedMotoServer
import numpy
import s3fs

ip = "127.0.0.1"
port = 3000
endpoint_url = f"http://{ip}:{port}"

if __name__ == '__main__':
    server = ThreadedMotoServer(ip_address="127.0.0.1", port=3000, verbose=False)
    server.start()

    s3 = boto3.client('s3', endpoint_url=endpoint_url)
    bucket_name = 'test'
    s3.create_bucket(Bucket=bucket_name)

    base_url = f"s3://{bucket_name}"

    fs = s3fs.S3FileSystem(endpoint_url=endpoint_url)

    arr = numpy.arange(42)
    af = asdf.AsdfFile({'arr': arr})

    url = f"{base_url}/test.asdf"
    with fs.open(url, mode="wb") as f:
        af.write_to(f)

    with fs.open(url, mode="rb") as f:
        with asdf.open(f, copy_arrays=True) as af:
            print(af['arr'][0])
```
with:
```
  File ".../asdf/asdf/generic_io.py", line 822, in read_into_array
    return np.fromfile(self._fd, dtype=np.uint8, count=size)
io.UnsupportedOperation: fileno
```

braingram:
Fixed in #1557

PR #1557, opened 2023-06-06
test for valid fileno before creating RealFile

braingram:
fileno is needed for memory mapping and numpy.frombuffer. Test that fileno does not result in an ``io.UnsupportedOperation`` before creating a ``generic_io.RealFile`` and if the error occurs, create a ``generic_io.RandomAccessFile`` instead.

fixes issue #1552


braingram:
Python 3.8 devdeps failures are addressed in: https://github.com/asdf-format/asdf/pull/1556

WilliamJamieson (approved):

PR #1559, opened 2023-06-07
convert Reference to Converter

braingram:
This is draft PR and should not be merged.

I'm not at the moment seeing a way to move `asdf.reference.Reference` to a `Converter` without changes to the the extension API.

The core issue is that `Reference` does not have a tag and instead uses the yaml map tag: "tag:yaml.org,2002:map" (which does not appear in the manifest for the asdf core extension). This causes the converter to be ignored in the extension because of the code in the `ConverterProxy` filters out any tag definitions that aren't in the extension:
https://github.com/asdf-format/asdf/blob/b0c9a50e8d3add337e1271369a3bf834925176a4/asdf/extension/_converter.py#L196-L202
and the manager doesn't include converters with no tags:
https://github.com/asdf-format/asdf/blob/b0c9a50e8d3add337e1271369a3bf834925176a4/asdf/extension/_manager.py#L39

To work around this issue, this PR introduces a special tag `"*"` to be used for converters that should always be included in the extension. The `ReferenceConverter` then exploits `select_tag` to return the standard yaml map tag.

### DIFF
--- a/asdf/core/_converters/reference.py
+++ b/asdf/core/_converters/reference.py
@@ -1,0 +1,18 @@
+from asdf.extension import Converter
+
+
+class ReferenceConverter(Converter):
+    tags = ["*"]
+    types = ["asdf.reference.Reference"]
+
+    def to_yaml_tree(self, obj, tag, ctx):
+        from asdf.generic_io import relative_uri
+
+        uri = relative_uri(ctx.url, obj._uri) if ctx.url is not None else obj._uri
+        return {"$ref": uri}
+
+    def from_yaml_tree(self, node, tag, ctx):
+        pass
+
+    def select_tag(self, obj, tag, ctx):
+        return "tag:yaml.org,2002:map"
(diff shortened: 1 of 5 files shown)


braingram:
The changes in this PR and the `NDArrayConverter` in #1537 can be used to fix the `NdarrayMixin` conversion errors in asdf-astropy (and possibly the `FITS_rec` issue in stdatamodels) which result from `NDArrayConverter` warning when serializing `NdarrayMixin` instances as ndarrays. To fix the error a converter like the following could be added to asdf-astropy:
```python
class NdarrayMixinConverter(Converter):
    tags = ["*"]

    types = ["astropy.table.ndarray_mixin.NdarrayMixin"]

    def to_yaml_tree(self, obj, tag, ctx):
        from asdf.core._converters.ndarray import NDArrayConverter

        return NDArrayConverter.to_yaml_tree(self, obj, tag, ctx)

    def from_yaml_tree(self, node, tag, ctx):
        raise NotImplementedError()

    def select_tag(self, obj, tags, ctx):
        return "tag:stsci.edu:asdf/core/ndarray-1.0.0"
```

This converter could check `obj` during `select_tag` to verify that it can be safely converted as a ndarray. There are a few things about this solution that don't seem optimal including:
- requirement that the converter imports (the now private) `NDArrayConverter`
- `select_tag` returns a hard-coded tag (this would require updating when the tag version changes and there is no guarantee that the returned tag is supported by another extension)
- the `"*"` tag syntax isn't obvious (it could be taken to mean that this converter should be used when deserializing all objects when instead this is never used for deserialization)

PR #1567, opened 2023-06-26
fix docs dependencies

braingram:
Several dependencies of `sphinx-asdf` were being used directly in `asdf`. This PR either includes those dependencies or makes changes to not require them.

- `sphinx-inline-tabs` (used in [Installing extensions via entry points](https://asdf--1567.org.readthedocs.build/en/1567/asdf/extending/extensions.html#installing-extensions-via-entry-points) <- link is to docs built with this PR)
- `myst-parser` a md to rst converter which appears to only have been used to render the `CODE_OF_CONDUCT.md`. This PR converts that file to `rst`. See note below about `CODE_OF_CONDUCT.md`
- `graphviz` to render the [class inheritance diagrams](https://asdf--1567.org.readthedocs.build/en/1567/asdf/developer_api.html#id3)

This PR temporarily has the sphinx-asdf dependency installing from the source branch for https://github.com/asdf-format/sphinx-asdf/pull/69 which removes the above dependencies from sphinx-asdf. Post approval and before merge this dependency on main should be changed back to the released version of sphinx-asdf.

This PR removes the top level `CODE_OF_CONDUCT.md`. This will mean that the 'About' section on the right side of the github repository page will list the organization-wide code of conduct (see: https://github.com/asdf-format/.github/blob/main/CODE_OF_CONDUCT.md). As this file is also included in the asdf docs (and I was able to find no easy way to have sphinx fetch and include an external file) this PR adds `docs/asdf/CODE_OF_CONDUCT.rst` (which is a copy of the organization wide code of conduct with a comment to editors mentioning that it is a copy).


nden (approved):
LGTM

PR #1568, opened 2023-06-26
bump numpy lower pin

braingram:
Numpy 1.21 reached end of life.

We had the lower pin at 1.20, this PR bumps the lower pin to 1.22.


WilliamJamieson (approved):

braingram:
Thanks for the review!

PR #1570, opened 2023-06-27
Add `zstd` as a built-in compression algorithm

mspacek:
Hello! I've been silently watching ASDF for a while now, in our search for a potential scientific data packaging format other than HDF5. I'm very impressed and convinced that ASDF is the one.

While looking into compression, I discovered `zstd`, and think it would be worthy to add zstd as a built-in compression algorithm, to complement `lz4`. `zstd` is the clear successor to `lz4`, written by the same author. The two are similar in their very high decompression speed, but zstd compresses faster at a given compression ratio, or compresses better at a given compression speed. Decompression speed is constant regardless of compression level.

In my own informal tests on a ~1GB 2D uint16 array, I find zstd at level 5 compresses in ~ 13 s and decompresses in ~ 3 s, with a compression ratio of 2.5. The closest I can get with lz4 is at "hi" compression level 9, which takes ~ 90 s to compress and ~ 2.5 s to decompress, with a compression ratio of 1.9. This is on a 7 year old i7-6600U laptop with 16 GB of RAM.

I know there have been a couple of mentions of zstd here in the issues/PRs, especially as a BOSC plugin. I tried setting it up as a plugin, but got stuck pretty quickly, and found it easier to simply add it as a built-in. The code is very analogous to the existing lz4, zlib and bz2 code. I also added a unit test, which seems to run successfully on my system (some other tests fail, esp YAML-related, but I think I may be missing some configuration settings or packages).

Please let me know how I can improve this. It would be great to see this state of the art compression algorithm built into ASDF!

### DIFF
--- a/.github/workflows/s390x.yml
+++ b/.github/workflows/s390x.yml
@@ -73,6 +73,6 @@ jobs:
           run: |
             python3 -m venv --system-site-packages tests
             source tests/bin/activate
-            pip3 install --upgrade pip setuptools gwcs pytest pytest-doctestplus pytest-remotedata
+            pip3 install --upgrade pip setuptools gwcs pytest pytest-doctestplus pytest-remotedata zstandard
             pip3 install -e .[all,tests]
             python3 -m pytest --remote-data
--- a/README.rst
+++ b/README.rst
@@ -162,9 +162,12 @@ It is possible to compress the array data when writing the file:
 
     af.write_to("compressed.asdf", all_array_compression="zlib")
 
-The built-in compression algorithms are ``'zlib'``, and ``'bzp2'``.  The
-``'lz4'`` algorithm becomes available when the `lz4 <https://python-lz4.readthedocs.io/>`__ package
-is installed.  Other compression algorithms may be available via extensions.
+The built-in compression algorithms are ``'zlib'``, and ``'bzp2'``.
+The ``'lz4'`` algorithm becomes available when the
+`lz4 <https://python-lz4.readthedocs.io/>`__ package is installed.
+The ``'zstd'`` algorithm becomes available when the
+`zstandard <https://python-zstandard.readthedocs.io//>`__ package is installed.
+Other compression algorithms may be available via extensions.
 
 .. _end-compress-file:
 
--- a/asdf/_tests/commands/tests/test_defragment.py
+++ b/asdf/_tests/commands/tests/test_defragment.py
@@ -52,3 +52,8 @@ def test_defragment_bzp2(tmpdir):
 def test_defragment_lz4(tmpdir):
     pytest.importorskip("lz4")
     _test_defragment(tmpdir, "lz4")
+
+
+def test_defragment_zstd(tmpdir):
+    pytest.importorskip("zstandard")
+    _test_defragment(tmpdir, "zstd")
--- a/asdf/_tests/test_array_blocks.py
+++ b/asdf/_tests/test_array_blocks.py
@@ -860,9 +860,11 @@ def test_block_allocation_on_validate():
 
 
 @pytest.mark.parametrize("all_array_storage", ["internal", "external", "inline"])
-@pytest.mark.parametrize("all_array_compression", [None, "", "zlib", "bzp2", "lz4", "input"])
+@pytest.mark.parametrize("all_array_compression",
+                         [None, "", "zlib", "bzp2", "lz4", "zstd", "input"])
 @pytest.mark.parametrize("compression_kwargs", [None, {}])
-def test_write_to_update_storage_options(tmp_path, all_array_storage, all_array_compression, compression_kwargs):
+def test_write_to_update_storage_options(tmp_path, all_array_storage,
+                                         all_array_compression, compression_kwargs):
     if all_array_compression == "bzp2" and compression_kwargs is not None:
         compression_kwargs = {"compresslevel": 1}
 
--- a/asdf/_tests/test_compression.py
+++ b/asdf/_tests/test_compression.py
@@ -113,6 +113,13 @@ def test_lz4(tmp_path):
     _roundtrip(tmp_path, tree, "lz4")
 
 
+def test_zstd(tmp_path):
+    pytest.importorskip("zstandard")
+    tree = _get_large_tree()
+
+    _roundtrip(tmp_path, tree, "zstd")
+
+
 def test_recompression(tmp_path):
     tree = _get_large_tree()
     tmpfile = os.path.join(str(tmp_path), "test1.asdf")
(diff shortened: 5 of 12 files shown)


mspacek:
@braingram I think I was just a bit naive about how much boilerplate would be required, and quickly lost patience :) But now that I see it implemented, it doesn't look so complicated.

Is there some criteria for deciding whether a compression algorithm should be an extension vs. built-in? Should lz4 perhaps be removed as a built-in and implemented as an extension instead?

Coincidentally, I noticed that the lz4 code was unnecessarily complex. The `lz4` Python package now makes it very easy to compress data and decompress streamed data, but in the frame format instead of block format. This seems to speed up decompression noticeably, but unfortunately the two formats are incompatible. I've opened up #1571 as a place to consider this.

braingram:
@mspacek Thanks for the quick reply and all the work you've put into this. I wrote a lot below in part to help collect my own thoughts.

> @braingram I think I was just a bit naive about how much boilerplate would be required, and quickly lost patience :) But now that I see it implemented, it doesn't look so complicated.

Perhaps having an example extension that provides a custom `Compressor` would be helpful? I haven't quite thought this through but it might be possible to have an `asdf-compressors` package which provides zstd (and the lz4 frame you mentioned, maybe even roll in the [blosc compression from 'abacusutils'](https://github.com/abacusorg/abacusutils/blob/7ea712cce6990fc59150d698b53a8824375ed82e/abacusnbody/data/asdf.py#L28)). It should be possible to set up the package in a way where the algorithms are all optional to allow users to select which ones they want to support (`pip install asdf-compressors[zstd]` to just install the zstandard compressor). Hopefully it could be flexible enough to be useful but still simple enough to serve as an example for writing a custom `Compressor`.

> Is there some criteria for deciding whether a compression algorithm should be an extension vs. built-in? Should lz4 perhaps be removed as a built-in and implemented as an extension instead?

There is no criteria that I'm aware of.

The standard only refers to `zlib` and `bzp2`: https://asdf-standard.readthedocs.io/en/1.0.3/file_layout.html#compression and I'm inclined to not add more compression algorithms to the core asdf library (to limit the dependencies and complexity) as long as the new algorithms can be supported by the extension system (if not than I'd prefer updating the extension system). I expect that this puts developers in a situation where it's easiest to include the custom `Compressor` in their larger tool (like with abacusutils) making the extension less useful for other users.

I don't believe compression is widely used (the largest ASDF user, [STScI](https://github.com/spacetelescope/) does not use any compression, which puts it lower on my list of priorities) and there are definite concerns about compression and 'archivablity' (as described in the [ASDF paper](https://www.sciencedirect.com/science/article/pii/S2213133715000645#s000105)). These concerns are more relaxed for extensions whereas files produced using only the asdf core should be as archival as possible.

> 
> Coincidentally, I noticed that the lz4 code was unnecessarily complex. The `lz4` Python package now makes it very easy to compress data and decompress streamed data, but in the frame format instead of block format. This seems to speed up decompression noticeably, but unfortunately the two formats are incompatible. I've opened up #1571 as a place to consider this.

Thanks for pointing that out and opening the other PR!

If it sounds good to you. I propose that we plan to have `zstd` and `lz4` frame in extensions (perhaps in an `asdf-compressors` package). Are these algorithms something you expect to need in the near term?

mspacek:
@braingram Indeed, I think an example extension including all the boilerplate files would be very useful. I wasn't sure if one already existed, and I didn't look very hard. I only looked at `asdf/extension/_compressor.py` and naively assumed that all I needed to do was copy and rename and edit it.

An `asdf-compressors` package definitely sounds good!

Given that only `zlib` and `bz2` are part of the core standard, for clarity it would be nice if `lz4` **wasn't** in the core library. Would it be feasible to move the current `lz4` (block format) into an extension? My naive thoughts on backward compatibility make me think that wouldn't be a problem. For the next major release, the move to an extension could be pointed out as a kind of API change, and the docs could be changed from "for lz4 support, please install the `lz4` package" to something like "please install the full `asdf-compressors` package, or just `asdf-compressors[lz4]`". Or am I missing something?

I think it might also be quite useful to stress in the docs just how *fast* lz4 and zstd are, especially at decompression, compared to zlib and bz2. The difference is striking enough that it changes the equation for users deciding whether or not to compress at all.

That said, I'm also not completely convinced that compression is the way to go for our own uses. Certainly it adds an extra step when manually interpreting array data in an .asdf file, such as with a hex editor. On the other hand for ridiculously compressible data, it's a shame to not compress losslessly. For zstd, I've checked that the compressed data are indeed readable using the zstd CLI. I haven't tested it myself, but according to the docs, the lz4 CLI can only handle the frame format, and not the block format: https://github.com/lz4/lz4/blob/dev/doc/lz4_manual.html . That makes the existing lz4 block format rather less archivable in my opinion.

Is there some process in ASDF that could deprecate writing the lz4 block format, and default to the frame format instead? I suppose for backward compatibility, the `lz4` label would unfortunately have to remain as-is, representing the block format, and a new label (`lz4f`?) would have to be created for the lz4 frame format?

I don't need zstd or lz4-frame quite yet, perhaps in the medium term though, so there's no huge rush for an `asdf-compressors` package. But it would be great to get it going fairly soon. I'm happy to help, but I'll need some guidance.

CagtayFabry:
Thank you for the valuable and insightful discussion and implementation @mspacek @braingram 
I can see the compression topic becoming more and more relevant at some point in the future as well.

> Is there some process in ASDF that could deprecate writing the lz4 block format, and default to the frame format instead? I suppose for backward compatibility, the `lz4` label would unfortunately have to remain as-is, representing the block format, and a new label (`lz4f`?) would have to be created for the lz4 frame format?

May I suggest adding this as a possible point to review or discuss for ASDF standard 2.0? (cref https://github.com/asdf-format/asdf-standard/issues/248)

braingram:
@CagtayFabry Thanks for linking that issue. I'll give it a read through.

@mspacek I like the idea of deprecating writing the `lz4` block format (and tag) and moving it to `asdf-compressors` (as it's already an optional dependency). I will take another look through some key downstream packages to see if any of them will be impacted.

Given this PR and https://github.com/asdf-format/asdf/pull/1571 adding lz4 frame and zstd compression should be possible with the existing extension system (a skim through the https://github.com/asdf-format/asdf-standard/issues/248 suggests that other algorithms might require changes).

I opened an issue to discuss moving lz4 to a separate package (asdf-compressors) where we could also add ztd support. https://github.com/asdf-format/asdf/issues/1574

braingram:
Thanks again for opening it @mspacek and please take a look at https://github.com/asdf-format/asdf-compression which adds zstd compression support via an asdf extensions. I'm closing this PR as the support in asdf-compression should hopefully address this need (once it is released on pypi).

Issue #1553, opened 2023-05-25
`AsdfFile.info` loads array data


braingram:
It appears that this line is triggering array data to be loaded for every array displayed in the the info:
https://github.com/asdf-format/asdf/blob/e86f69af6a90051ffc4a45f2de6c669cdaba836d/asdf/_node_info.py#L202

Here is a minimal script showing that info loads array data:
```python
import asdf
import numpy as np

fn = 'test.asdf'
tree = dict([(k, np.arange(ord(k))) for k in 'abc'])
asdf.AsdfFile(tree).write_to(fn)

with asdf.open(fn) as af:
    assert 'unloaded' in str(af['b'])
    af.info()
    assert 'unloaded' not in str(af['b'])
```

PR #1572, opened 2023-06-27
raise AttributeError for __asdf_traverse__ on NDArrayType

braingram:
NDArrayType does not implement this method and if not intercepted (and an AttributeError raised) than AsdfFile.info will load all array data

Fixes: https://github.com/asdf-format/asdf/issues/1553

This PR adds a regression test for this bug.

### DIFF
--- a/asdf/_tests/_issues/test_1553.py
+++ b/asdf/_tests/_issues/test_1553.py
@@ -1,0 +1,19 @@
+import numpy as np
+
+import asdf
+
+
+def test_1553(tmp_path):
+    """
+    AsdfFile.info loads array data
+
+    https://github.com/asdf-format/asdf/issues/1553
+    """
+    fn = tmp_path / "test.asdf"
+    tree = dict([(k, np.arange(ord(k))) for k in "abc"])
+    asdf.AsdfFile(tree).write_to(fn)
+
+    with asdf.open(fn) as af:
+        assert "unloaded" in str(af["b"])
+        af.info()
+        assert "unloaded" in str(af["b"])
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ The ASDF Standard is at v1.6.0
 - Remove deprecated TagDefinition.schema_uri [#1595]
 - Removed deprecated AsdfFile.open and deprecated asdf.open
   AsdfFile.write_to and AsdfFile.update kwargs [#1592]
+- Fix ``AsdfFile.info`` loading all array data [#1572]
 
 2.15.1 (2023-08-07)
 -------------------
--- a/asdf/_tests/_regtests/test_1553.py
+++ b/asdf/_tests/_regtests/test_1553.py
@@ -3,9 +3,9 @@ import numpy as np
 import asdf
 
 
-def test_1553(tmp_path):
+def test_asdf_info_should_not_load_arrays(tmp_path):
     """
-    AsdfFile.info loads array data
+    AsdfFile.info should not load array data
 
     https://github.com/asdf-format/asdf/issues/1553
     """
--- a/asdf/tags/core/ndarray.py
+++ b/asdf/tags/core/ndarray.py
@@ -368,6 +368,13 @@ class NDArrayType(_types._AsdfType):
         # number of array creations in the general case.
         if attr == "__array_struct__":
             raise AttributeError
+        # AsdfFile.info will call hasattr(obj, "__asdf_traverse__") which
+        # will trigger this method, making the array, and loading the array
+        # data. Intercept this and raise AttributeError as this class does
+        # not support that method
+        # see: https://github.com/asdf-format/asdf/issues/1553
+        if attr == "__asdf_traverse__":
+            raise AttributeError
         return getattr(self._make_array(), attr)
 
     def __setitem__(self, *args):


WilliamJamieson (approved):

asdf/_tests/_issues/test_1553.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+import asdf
+
+
+def test_1553(tmp_path):
```
WilliamJamieson, line 6: Use a descriptive name for this test.

WilliamJamieson (dismissed):

asdf/_tests/_issues/test_1553.py
braingram: This follows the convention set up in: https://github.com/asdf-format/asdf/pull/1537

WilliamJamieson: I completely disagree with this approach. Test names should be descriptive of what is going on in a test. That way you have an idea of what is broken when a test fails without having to go and read a docstring and then some issue on github.

braingram:
Following up on discussion with @perrygreenfield and comments by @WilliamJamieson and @pllim on test naming here.

It was discussed that it might be easiest to address the issue in this smaller PR compared instead of #1537 (which makes other large changes).

It was never my intention to discourage engagement or make this project 'hostile' to developers. If these changes did anything of the sort I apologize and am happy to not use the naming structure. I never had strong opinions about it but hopefully I can explain why I chose the structure and we can reach a happy compromise that addresses the valid and helpful concerns that everyone has raised.

During work on #1537 I encountered many issues (see the PR for the full list). Many of these were quite high level (similar to issues that an end-user might raise using the public API). The reproducers for these high level issues are often not good unit tests as they cover a wide surface area of the code. It occurred to me that it might be nice to separate the regression tests added alongside fixing these issues from unit tests (and in the case of asdf the mostly unorganized test suite to hopefully work towards a more organized structure). Separating the unit and regression tests would have a few benefits, a major one being that it would be relatively easy to run coverage on just the unit tests (to reveal obvious gaps). In addition to separating unit and regression tests, it occurred to me that the issues often contain discussion, external links, failed exploration, and other information that often does not make it into the code, commits and comments with the PR that fixes the issue. As this information can be useful if a regression tests starts failing it seems beneficial to make the issue number prominent in a failed test run (and not a comment in the regression test requiring one to open the source for the failed test). Given these reasons I put forward the structure that was proposed in #1537 that regression tests for issues be placed in `_issues/test_<issue number>` and be named `test_<issue number`.

As noted, there are issues with this approach:
- the issue number, in itself, is not information
- migration to a new system (not github) means that issue numbers are not permanent
- no one memorizes issue numbers
- what if the test span multiple modules?
- what if a future issue is opened about the old issue?
- additionally the issue number as show above is redundant if in the file name and test name

I like @WilliamJamieson suggestion of test classes. This seems like a way to link the issue number to the regression tests (and provides obvious grouping for tests related to the issue). I do not mean to dismiss this suggestion and would be happy to go with this approach. What I'm proposing in this PR is to:

- change the test naming to something descriptive (for this PR it's currently `test_asdf_info_loads_arrays`)
- include a brief description and link to the issue in the docstring
- rename `_issues` to `_regtests` to further make these distinct from unit tests
- keep the files as `test_<issue number>.py`. As pytest usually includes both the file name and test name in output this means that both the issue number and descriptive name are apparent on a failed test run

A few reasons for keeping the issue number as the filename are:
- as noted above it is printed on a failed test run and can be useful for finding the wider context contained in the issue discussion
- during work on the issue, where only one issue is in mind, it makes it easy to type/complete while working on the issue `pytest asdf/_tests/_regtests/test_1572.py` or `pytest -k test_1572`
- this might help with migration if the code is moved from github. Each file can be edited to point to the new issue (if the issues can be migrated) or renamed based on whatever strategy is devised at the time.

I don't believe I have yet responded to @pllim [question](https://github.com/asdf-format/asdf/pull/1537#pullrequestreview-1511759757) about what if a new issue is opened for the same problem (also please let me know if I've missed any other questions and concerns). If the original regression test failed to produce the same issue, it seems reasonable to delete it and replace it with a new test (linking back to the new issue, which links to the old issue, probably also including a description of the old and new issues in the docstring of the test). If the new failure is sufficiently distinct, adding a second regression test might be a sensible option.

I am happy to add to the [contributing](https://asdf.readthedocs.io/en/stable/asdf/contributing.html#contributing-code-and-bug-fixes) documentation describing whatever structure we agree upon.

I don't mean to belabor anything but I am not strongly committed to these changes I only found it useful to organize things by issue while attempting to tackle the numerous issues exposed during rewriting the block management. I am happy to adopt test classes as @WilliamJamieson suggested I only wanted to raise the option of keeping the issue name in the filename as I had to pick a file name for the new tests and this seemed like a way to insert the issue number in a way that should be easy to migrate.

pllim:
If you want to separate unit test from regtest, sure, have sub-dirs. Though personally I don't find that distinction useful in practice. It is either run in CI or not. But maybe useful if you want users to only run `pytest tests/unit_tests/` or something.

```
tests/
    unit_tests/
    regtests/
```

Issue number and links are only useful if you decide you want to revisit old discussions. If the test name is descriptive and purpose of the test is clear, I find this need rare. I think something like this would be good enough for most cases.

```python
test_chunk_overflow():
    """Test such and such bug that happens when this and that.
    See <GitHub issue link> for more info.
    """
    # test_code_here
```

In the future, if someone else run into another edge case of this bug, it is immediately clear which function they have to add to. And they also have access to the link if they want to know more. Even if there is no link provided, git blame would take them to the original PR.

pllim:
Thank you for the lengthy clarification! I appreciate that you took time to clarify things.

braingram:
devdeps will likely fail due to pyyaml and cython3 incompatibility: https://github.com/yaml/pyyaml/issues/601

braingram:
@WilliamJamieson does the proposed test structure/naming in this PR address your concern? https://github.com/asdf-format/asdf/pull/1572#issuecomment-1636435463

Issue #1565, opened 2023-06-21
`asdf.asdf.SerializationContext` not found


WilliamJamieson:
When building docs elsewhere, I get the following warning:
```
.from_yaml_tree:1: WARNING: py:class reference target not found: asdf.asdf.SerializationContext
```
when building API documentation using [`automodapi`](https://github.com/astropy/sphinx-automodapi).

Tracing this backwards, I believe the warning ultimately originates from: https://github.com/asdf-format/asdf/blob/c3e358886a41c36b81a2a26bd03aedc2720c9817/asdf/extension/_converter.py#L143

in ASDF. The warning occurs because there is no documentation for `asdf.asdf.SerializationContext` in the ASDF documentation for my documentation to link to. One can see this absence in the ASDF [documentation](https://asdf.readthedocs.io/en/latest/api/asdf.extension.Converter.html#asdf.extension.Converter.from_yaml_tree) too, where the `asdf.asdf.SerializationContext` does not link to anything (which it should).

ASDF needs to extend its API documentation to include the `asdf.asdf.SerializationContext` so that other projects can build documentation for objects subclassing `asdf.extension.Converter`.

braingram:
Agreed!

xref: https://github.com/asdf-format/asdf/issues/1549

Issue #1549, opened 2023-05-15
`SerializationContext` missing from docs


braingram:
`SerializationContext` is mentioned in the documentation for converters:
https://asdf.readthedocs.io/en/stable/api/asdf.extension.Converter.html#asdf.extension.Converter.select_tag
and in the `future plans`:
https://asdf.readthedocs.io/en/latest/asdf/developer_overview.html#future-plans-for-serializationcontext

But it's methods and usage is undocumented.

PR #1573, opened 2023-06-27
move SerializationContext to asdf.extension and add it to docs

braingram:
As far as I can tell, the public usage of SerializationContext occurs exclusively through extensions.

Rather than document `asdf.asdf` (where `SerializationContext` was implemented prior to this PR) which was previously undocumented, this PR moves `SerializationContext` to `asdf.extension._serialization_context` and exposes it publicly at `asdf.extension.SerializationContext`.

Code that imports `SerializationContext` from `asdf.asdf` should continue to work (as `SerializationContext` is imported in `asdf.asdf`) but should be updated to import from the documented public location once this PR makes it into a release.

Link to relevant updated docs: https://asdf--1573.org.readthedocs.build/en/1573/api/asdf.extension.SerializationContext.html#asdf.extension.SerializationContext

Fixes https://github.com/asdf-format/asdf/issues/1565
Fixes https://github.com/asdf-format/asdf/issues/1549

### DIFF
--- a/docs/asdf/developer_api.rst
+++ b/docs/asdf/developer_api.rst
@@ -7,6 +7,11 @@ Developer API
 The classes and functions documented here will be of use to developers who wish
 to create their own custom ASDF types and extensions.
 
+.. automodapi:: asdf.asdf
+    :no-inheritance-diagram:
+    :include: SerializationContext
+    :skip: AsdfConversionWarning, AsdfDeprecationWarning, AsdfObject, AsdfSearchResult, AsdfWarning, DelimiterNotFoundError, Extension, ExtensionMetadata, ExtensionProxy, HistoryEntry, Software, ValidationError, Version
+
 .. automodapi:: asdf.tagged
 
 .. automodapi:: asdf.exceptions
--- a/docs/asdf/developer_api.rst
+++ b/docs/asdf/developer_api.rst
@@ -7,11 +7,6 @@ Developer API
 The classes and functions documented here will be of use to developers who wish
 to create their own custom ASDF types and extensions.
 
-.. automodapi:: asdf.asdf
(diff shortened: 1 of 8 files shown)


docs/asdf/developer_api.rst
```diff
@@ -7,6 +7,11 @@ Developer API
 The classes and functions documented here will be of use to developers who wish
 to create their own custom ASDF types and extensions.
 
+.. automodapi:: asdf.asdf
+    :no-inheritance-diagram:
+    :include: SerializationContext
+    :skip: AsdfConversionWarning, AsdfDeprecationWarning, AsdfObject, AsdfSearchResult, AsdfWarning, DelimiterNotFoundError, Extension, ExtensionMetadata, ExtensionProxy, HistoryEntry, Software, ValidationError, Version
```
braingram, line 13: automodapi does not appear to pass `:include:` to automodsumm when generating class tables. This resulted in many warnings as the table contained rows linking to undocumented classes. I don't know why it also wants to include everything imported but this ugly list of skips appears to work.

WilliamJamieson: Why don't you just specify an `__all__`. automodapi generates documentation via a `*` import of the module. If no `__all__` is defined for the module, then anything not prefixed with `_` will be imported and automodapi will try to generate documentation for it. See #1577

braingram: Thanks! That explains why `:include:` did not do what I expected based on the documentation. Rather than commit to an __all__ for `asdf.asdf` (which seems more like it should be private) I moved `SerializationContext` and exposed it at `asdf.extension.SerializationContext` as it's public use appears to be exclusively for extension developers.

PR #1575, opened 2023-06-28
free tree on AsdfFile.close

### LINKED ISSUE #1558: closed `AsdfFile` instances hold private reference to tree
After `close` an `AsdfFile` instance will prevent tree access through the `tree` property:
https://github.com/asdf-format/asdf/blob/b0c9a50e8d3add337e1271369a3bf834925176a4/asdf/asdf.py#L552-L562

However, because `AsdfFile._tree` is not cleared, the closed instance will hold references to objects in the tree preventing them from being garbage collected as seen in the following example.

```python
import asdf
import numpy
import weakref

fn = 'test.asdf'
asdf.AsdfFile({'arr': numpy.arange(1000)}).write_to(fn)

with asdf.open(fn, copy_arrays=True, lazy_load=False) as af:
    array_ref = weakref.ref(af['arr'])

# we shouldn't be able to resolve this ref
try:
    print(af['arr'])
except OSError as e:
    print(f"Unable to access array through tree: {e}")

r = array_ref()
if r is not None:
    print(f"Weakref resolved to {r[:10]}...")
```
outputs:
```
Unable to access array through tree: Cannot access data from closed ASDF file
Weakref resolved to [0 1 2 3 4 5 6 7 8 9]...
```

braingram:
With these changes, when an AsdfFile object is closed the tree (which already could not be accessed because of the tree property gatekeeping) is set to a blank AsdfObject.

This allows the garbage collector to clean up objects referenced in the tree even though the AsdfFile object does not fall out of scope.

Fixes #1558

### DIFF
--- a/asdf/_tests/_issues/test_1558.py
+++ b/asdf/_tests/_issues/test_1558.py
@@ -1,0 +1,34 @@
+import weakref
+
+import numpy as np
+import pytest
+
+import asdf
+
+
+def test_1558(tmp_path):
+    """
+    closed AsdfFile instances hold private reference to tree
+
+    https://github.com/asdf-format/asdf/issues/1558
+    """
+
+    fn = tmp_path / "test.asdf"
+    asdf.AsdfFile({"a": np.arange(1000), "b": np.arange(42)}).write_to(fn)
+
+    with asdf.open(fn, copy_arrays=True, lazy_load=False) as af:
+        array_weakref = weakref.ref(af["a"])
+        array_ref = af["b"]
+
+    # we shouldn't be able to access the now closed file
+    with pytest.raises(OSError, match="Cannot access data"):
+        af["a"]
+    with pytest.raises(OSError, match="Cannot access data"):
+        af["b"]
+
+    # we also should not be able to resolve the weak reference
+    # meaning the tree should have been cleaned up
+    r = array_weakref()
+    assert r is None
+
+    assert array_ref is not None
--- a/asdf/asdf.py
+++ b/asdf/asdf.py
@@ -454,6 +454,9 @@ class AsdfFile:
             self._fd.close()
             self._fd = None
             self._closed = True
+            # as we're closing the file, also empty out the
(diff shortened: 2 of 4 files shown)


asdf/_tests/_issues/test_1558.py
```diff
@@ -0,0 +1,34 @@
+import weakref
+
+import numpy as np
+import pytest
+
+import asdf
+
+
+def test_1558(tmp_path):
```
WilliamJamieson, line 9: Use a descriptive name for the test.

WilliamJamieson (dismissed):

asdf/_tests/_issues/test_1558.py
braingram: This follows the convention set up in: https://github.com/asdf-format/asdf/pull/1537

asdf/asdf.py
```diff
@@ -449,6 +449,9 @@ def close(self):
             self._fd.close()
             self._fd = None
             self._closed = True
+            # as we're closing the file, also empty out the
+            # tree so that references to array data can be released
+            self._tree = AsdfObject()
```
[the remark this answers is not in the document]
braingram: A custom exception sounds like a great idea. Currently and `OSError` is raised so any custom exception we raise could inherit from that exception class for backwards compatibility.
https://github.com/asdf-format/asdf/blob/b43b2c5b8935c9bf99a7ead553b32b11ccbee823/asdf/asdf.py#L559-L561
It might make sense to raise the same error type when a block is accessed from a closed file (and any other places where access to a closed file is prevented):
https://github.com/asdf-format/asdf/blob/b43b2c5b8935c9bf99a7ead553b32b11ccbee823/asdf/block.py#L1290-L1292
Perhaps a `ClosedFileAccess` exception that inherits from `OSerror`? That seems reasonably backwards compatible to include in 3.0 (and would work towards cleaning up the rather generic exceptions raised by asdf).

PR #1576, opened 2023-07-03
Remove unnecessary dependency

WilliamJamieson:
`asdf-unit-schemas` was never necessary for `asdf`, it only existed so that asdf-format/asdf-standard#342 could be merged without issue.

Since that entire line of support has been abandoned at this point, this dependency should be dropped.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ The ASDF Standard is at v1.6.0
 - Remove deprecated legacy extension API [#1464]
 - Drop Python 3.8 support [#1556]
 - Drop NumPy 1.20, 1.21 support [#1568]
+- Drop asdf-unit-schemas dependency [#1576]
 
 2.15.0 (2023-03-28)
 -------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,6 @@ dynamic = [
 dependencies = [
   "asdf-standard>=1.0.1",
   "asdf-transform-schemas>=0.3",
-  "asdf-unit-schemas>=0.1",
   "importlib-metadata>=4.11.4",
   "jmespath>=0.6.2",
   "jsonschema<4.18,>=4.0.1", # jsonschema 4.18 contains incompatible changes: https://github.com/asdf-format/asdf/issues/1485


zacharyburnett:
as user API is not changed, and users will not need to update their environments, is a change log entry necessary? I guess it doesn't hurt to still have one

WilliamJamieson:
> as user API is not changed, and users will not need to update their environments, is a change log entry necessary? I guess it doesn't hurt to still have one

Its useful to have in the changelog simply because the dropping of the dependency does change the Conda configuration slightly.

WilliamJamieson:
> I'm all for cleaning up this if indeed the splitting up of the standard has been abandoned. However until @perrygreenfield @nden @eslavich and other stake holders can weight in I don't think we should undo the work that has been done.

Looking at my notes for ASDF 3.0, the question of keeping/removing this dependency is moot. It is my understanding that as part of ASDF 3.0 (which is the main development path for ASDF now), we would stop packaging schemas which are no longer part of the ASDF-standard.

Originally, the `asdf-transforms-schemas` where added as a dependency to ASDF for backwards compatiblity reasons. That is since they were totally removed from ASDF-standard, users would have to add them separately as a dependency in order to continue using them. At around the same time, ASDF-standard was formally removed from the ASDF package itself and repackaged into its own independent python package. Thus ASDF then required both `asdf-standard` and `asdf-transform-schemas` as dependencies to facilitate backwards compatiblity for ASDF 2.x versions. Part of the plan for ASDF 3.0 was to totally remove `asdf-transform-schemas` (`asdf-standard` is still required as the ASDF package is fundamentally tied to it) since we were breaking backwards compatiblity in several other ways.

The original proposal for breaking ASDF-standard apart into more sub-packages, including `asdf-unit-schemas`, was inspired by the process followed by `asdf-transform-schemas`. As such, the new packages were supposed to be added to ASDF as dependencies to preserve backwards compatiblity in the same way. However, only `asdf-unit-schemas` was ever actually made a dependency of ASDF because the split of ASDF-standard was forgotten about and/or abandoned. Thus just as `asdf-transform-schemas` ought to be removed, `asdf-unit-schemas` should be removed for ASDF 3.0.

I have not included the removal of `asdf-transform-schemas` in this PR as this topic clearly needs more discussion.

WilliamJamieson:
> However, only `asdf-unit-schemas` was ever actually made a dependency of ASDF because the split of ASDF-standard was forgotten about and/or abandoned.

By this I mean that ASDF-standard has made significant changes to its schemas which have not being properly propagated into the packages it was intended to be split into. As such, the work initially carried out to facilitate this, including asdf-format/asdf-standard#342, needs to almost entirely be re-done in order to ensure that no loss of changes will occur and that the git history can be properly followed, if the ASDF developers wish to continue with the proposed "split".

PR #1585, opened 2023-07-11
fix jsonschema docs links in asdf.schema

braingram:


WilliamJamieson (approved):

braingram:
Thanks for the review!

PR #1602, opened 2023-07-24
[benchmarks] add commit URL for showing commits when clicked

zacharyburnett:
Currently, the benchmarks at https://spacetelescope.github.io/bench/asdf/ do not show a commit when clicked; this configuration change should fix that.


braingram (approved):
Thanks! LGTM

Issue #1519, opened 2023-04-11
ASDF handling of numpy scalar values (and scalar precision)


braingram:
Based off discussions in recent PRs:
https://github.com/asdf-format/asdf/pull/1474#discussion_r1148473915
https://github.com/asdf-format/asdf/pull/1475#discussion_r1139089411

ASDF (and the ASDF standard) should likely clarify how to handle scalar values.

For a short example of what should be clarified. Suppose a user adds a scalar (non-array) 'numpy.uint8' to the tree, then saves, then loads the tree. Should the value be a 'numpy.uint8', int?

To summarize the discussion so far based off contributions by @perrygreenfield 

What should the standard require?
----
Perhaps support for "The greatest common types for integers, floats and complex values across standard platforms... int64, float64 and complex128". With larger types possibly supported with associated metadata.

What is required for round-tripping?
----
"Up to the library", so for python:
- require permission to write out larger than standard values
- raise a warning when the library doesn't support the size and return a special object
- require round-tripping the special objects
- raise an exception if the standard requirement is not met

How should numpy scalars that fit within the standard scalar?
----
The library will return the largest (standard) type which can violate round-tripping. Any user that expects a 'numpy.uint8' to stay a 'numpy.uint8' (and not be promoted to an int) should convert the value separately.

braingram:
Some guidance is already provided in the standard for integers:
https://asdf-standard.readthedocs.io/en/1.0.3/known_limits.html#literal-integer-values-in-the-tree

One thought, which I think is in partial agreement with what @perrygreenfield described would be to say that scalars in the tree can be:
- int64
- float64
- complex128
By default, ASDF (the python implementation) will convert any numpy scalar that fits into these types (and violate round-tripping, this means `numpy.uint8` will become int). For any value outside of this range, I'm thinking we should throw an error mentioning the limitation and directing the user to a page with more information and a description of how to save the scalar as a ndarray (which will round trip without the same scalar issues).
We could enable an option to disable the automatic conversion from numpy scalar to builtin type (and instead throw an error).

braingram:
Copying from: https://github.com/asdf-format/asdf/pull/1605#discussion_r1276449384
```
[...]

PR #1605, opened 2023-07-26
convert numpy scalars to python types before yaml encoding

braingram:
[nep51](https://numpy.org/neps/nep-0051-scalar-representation.html) changes numpy scalars modifying the repr.

this change conflicts with represent_float/int (from pyyaml, which uses repr). With numpy 2.0 (which implements nep51) scalars are being encoded as, for example, 'np.float64(3.14)' instead of '3.14'. To work around this change, convert the values to builtin python types before passing them to pyyaml for encoding.

This PR also disables pyyaml devdeps testing as it is incompatible with cython 3.0 and it's failure to install stops devdeps testing.

### DIFF
--- a/asdf/yamlutil.py
+++ b/asdf/yamlutil.py
@@ -117,11 +117,12 @@ AsdfDumper.add_representer(OrderedDict, represent_ordereddict)
 # ----------------------------------------------------------------------
 # Handle numpy scalars
 
+
 for scalar_type in util.iter_subclasses(np.floating):
-    AsdfDumper.add_representer(scalar_type, AsdfDumper.represent_float)
+    AsdfDumper.add_representer(scalar_type, lambda dumper, data: dumper.represent_float(float(data)))
 
 for scalar_type in util.iter_subclasses(np.integer):
-    AsdfDumper.add_representer(scalar_type, AsdfDumper.represent_int)
+    AsdfDumper.add_representer(scalar_type, lambda dumper, data: dumper.represent_int(int(data)))
 
 
 def represent_numpy_str(dumper, data):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ The ASDF Standard is at v1.6.0
 - Fix issue opening files that don't support ``fileno`` [#1557]
 - Allow Converters to defer conversion to other Converters
   by returning ``None`` in ``Converter.select_tag`` [#1561]
+- Convert numpy scalars to python types during yaml encoding
+  to handle NEP51 changes for numpy 2.0 [#1605]
 
 2.15.0 (2023-03-28)
 -------------------
--- a/asdf/_tests/test_yaml.py
+++ b/asdf/_tests/test_yaml.py
@@ -287,4 +287,10 @@ def test_numpy_scalar(numpy_value, expected_value):
     yamlutil.dump_tree(tree, buffer, ctx)
     buffer.seek(0)
 
-    assert yamlutil.load_tree(buffer)["value"] == expected_value
+    loaded_value = yamlutil.load_tree(buffer)["value"]
+    if isinstance(numpy_value, np.floating):
+        abs_diff = abs(expected_value - loaded_value)
+        eps = np.finfo(numpy_value.dtype).eps
+        assert abs_diff < eps, abs_diff
+    else:
+        assert loaded_value == expected_value
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,8 +6,9 @@ git+https://github.com/asdf-format/asdf-unit-schemas.git
 git+https://github.com/asdf-format/asdf-wcs-schemas
 git+https://github.com/astropy/astropy
 git+https://github.com/spacetelescope/gwcs
-git+https://github.com/yaml/pyyaml.git
+#git+https://github.com/yaml/pyyaml.git
 # jsonschema 4.18 contains incompatible changes: https://github.com/asdf-format/asdf/issues/1485
 #git+https://github.com/python-jsonschema/jsonschema
 
 numpy>=0.0.dev0
+scipy>=0.0.dev0


braingram:
jwst devdeps failure is because of an unreleased fix in stdatamodels:
https://github.com/spacetelescope/stdatamodels/pull/184
and could also be fixed by:
https://github.com/asdf-format/asdf/pull/1594

asdf/yamlutil.py
```diff
@@ -117,11 +117,12 @@ def represent_ordereddict(dumper, data):
 # ----------------------------------------------------------------------
 # Handle numpy scalars
 
+
 for scalar_type in util.iter_subclasses(np.floating):
-    AsdfDumper.add_representer(scalar_type, AsdfDumper.represent_float)
+    AsdfDumper.add_representer(scalar_type, lambda dumper, data: dumper.represent_float(float(data)))
```
WilliamJamieson, line 122: Will this fully preserve floating point precision?

Alternately you can just set the `numpy` representation version you want using [`np.set_printoptions`](https://numpy.org/devdocs/reference/generated/numpy.set_printoptions.html#numpy-set-printoptions)

Adding something like this snippet in the correct place (such as `conftest.py` to fix unit tests) should also fix the issue 

```python
from asdf.util import minversion
import numpy as np

if not minversion(np, "2.0.dev"):
    np.set_printoptions(legacy="1.25")
```

For example adding this to `romancal`'s top level [`conftest.py`](https://github.com/spacetelescope/romancal/blob/main/romancal/conftest.py) fixes the errors resulting from the proposed `numpy` 2.0 changes.

See astropy/astropy#15096

WilliamJamieson: Note that astropy/astropy#15065, in particular the astropy.io.yaml suggest a much clearer way to do this than my above suggestion.

braingram: Thanks for taking a look and for sharing the links to how astropy is handling this change. https://github.com/astropy/astropy/pull/15065 looks interesting (as it appears numpy calls `str` prior to `repr` where it [adds the dtype portion of repr](https://github.com/numpy/numpy/blob/main/numpy/core/src/multiarray/scalartypes.c.src#L1068)). However due to the way pyyaml deserializes floats I'm not sure we want to continue using `str` or `repr` on the numpy scalars (more on that below).

The short answer the precision question is that this PR should have no negative impact (and in some cases have a positive impact) on roundtripping floating point scalars.

The long answer is more complicated.

There has been some recent discussion on scalar handling and ASDF (see https://github.com/asdf-format/asdf/issues/1519). As discussed, the YAML standard is not overly prescriptive on float precision https://yaml.org/type/float.html. ASDF does not further define float handling and we should document that users should use arrays stored in ASDF blocks for accurate roundtripping and control of precision. I'm pinging @perrygreenfield @eslavich and @nden as they can hopefully fill in details from the discussion that I've forgotten and/or failed to note.

The simplest case is float128 (and anything else more than 64 bits). asdf currently deserializes floating point scalars in the tree as python native `float`. This means that there is already and continues to be loss of precision for (systems that support) floats with more than 64 bits when these values are written to the tree and not the ASDF blocks.

For 64 bit floats both numpy and python reprs (by default) select the shortest string that will roundtrip (see https://docs.python.org/3/tutorial/floatingpoint.html). So converting a `np.float64` to `float` prior to serialization has no impact on precision.

For floats with less than 64 bits the situation is the messiest and is where the changes in this PR will have a small impact on what is written to and roundtripped through an ASDF file. The difference comes from the numpy repr choosing the shortest string that roundtrips with the precision of the datatype (whereas asdf will always convert the value to 64 bits on read). It's probably helpful to use an example.
Prior to this PR, `np.float16(3.143)` did not roundtrip.
```
>>> v = np.float16(3.143)
>>> asdf.testing.helpers.roundtrip_object(v) == v
False
```
This can be boiled down to numpy repr choosing '3.143' to represent the value, which when loaded as a float produces a slightly different '3.143'.
```
>>> repr(v)
3.143
>>> fv = float(repr(v))  # repr also 3.143
>>> fv == v
False
>>> abs(fv - v)
0.00042187499999979394
```
However, as expected, casting the read value back to 16 bit allows the comparison to pass:
```
>>> np.float16(fv) == v
True
```

To summarize, prior to this PR, what was written for a <64 bit numpy floating point scalar was the shortest string that would round trip if the precision of the original float was used. However because pyyaml (and asdf) read the floats as 64 bits, the read values can fail a 64 bit comparison.

With this PR the values are converted to a float prior to conversion to a string for serialization. Using the above example, with this PR
```
>>> v = np.float16(3.143)
>>> asdf.testing.helpers.roundtrip_object(v) == v
np.True_  # different repr due to NEP51
```
Casting the value back to 16 bits also works without issue:
```
>>> np.float16(asdf.testing.helpers.roundtrip_object(v)) == v
np.True_
```
However it should be noted that to achieve this, the string written to the file is different due to the conversion to 64 bits prior to the call to repr (which will now select the shortest string that reconstructs the 64 instead of 16 bit float).
```
>>> v = np.float16(3.143)
>>> v
np.float16(3.143)
>>> float(v)
3.142578125
```
So for this example with this PR asdf will write `3.142578125` instead of `3.143` to the file.

WilliamJamieson: Thanks for the explanation. As I stated before, I wanted to make sure ASDF can "roundtrip" scalars correctly.

Now that light has been shined on the issue of numpy scalars, I think ASDF should work towards serializing numpy scalars differently than the built in python scalars. [NEP 41](https://numpy.org/neps/nep-0041-improved-dtype-support.html) is working towards having more complicated `dtypes` (which would apply to scalars too). One of the main motivations for this effort is to encode units into the `dtype` itself (for performance). Looking towards this indicates that ASDF might want to start considering how to encode the `dtype` for  scalars.

WilliamJamieson (approved):

braingram:
3.9 devdeps failure is because the scipy nightly for 3.9 failed to build due to an anaconda server 500:
https://github.com/scipy/scipy/actions/runs/5682181723/job/15400012240

PR #1606, opened 2023-07-26
BOT: Changelog check has moved

pllim:
Similar to https://github.com/spacetelescope/rad/pull/299

xref https://github.com/scientific-python/action-check-changelogfile/issues/4


braingram (approved):
Thanks!

PR #1610, opened 2023-08-01
remove label checker ci job

braingram:
The label checker CI job does not support the `Manual Backport` label and requires manual updates for every release (see #1514). It was originally introduced to try to help supporting possible long-term-support releases. However subsequent discussion resulted in the cancellation of any LTS releases and as such this CI job is less useful.

Rather than update it to support the `Manual Backport` label this PR removes the job which will allow us to manually add PRs to milestones.

Note that the branch protection rules will need to be updated to remove the requirement for this job and the [wiki](https://github.com/asdf-format/asdf/wiki/Release-Process) will need to be updated to reflect that every PR will not need to be assigned a 'development' or 'backport-x.x.x' label.


zacharyburnett (approved):
this LGTM, if it's not affecting current `meeseeks` operations

braingram:
> this LGTM, if it's not affecting current `meeseeks` operations

Thanks! `meeseeks` still works by applying the labels (`meeseeks` looks at the label descriptions) or by triggering it manually.

PR #1611, opened 2023-08-01
require and add min version to sphinx theme for rtd

braingram:
readthedocs is choosing to use an ancient version of the sphinx theme (0.4.3 from 2019). This PR modifies the rtd environment file to require a newer version of the theme.


zacharyburnett (approved):
LGTM, docs pass

braingram:
Thanks!

PR #1615, opened 2023-08-07
switch to simple changelog check

braingram:
The scientific-python changelog checker by default also checks milestones. Since https://github.com/asdf-format/asdf/pull/1610 removed the requirement that PRs have milestones this can result in PRs incorrectly failing the changelog check because they don't have a milestone.

The workflow is based largely off of one used for jwst: https://github.com/spacetelescope/jwst/blob/master/.github/workflows/changelog.yml and what was [used previously in asdf](https://github.com/asdf-format/asdf/blob/9e746c899482523b668b8737cd744c4c8fb0c45e/.github/workflows/changelog.yml).

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -17,8 +17,10 @@ jobs:
     name: Confirm changelog entry
     runs-on: ubuntu-latest
     steps:
-    - name: Check change log entry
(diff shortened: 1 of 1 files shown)


braingram:
Thanks!

PR #1623, opened 2023-08-10
remove deprecated  load_schema argument, resolve_local_refs

braingram:
This PR removes the deprecated `resolve_local_refs` argument to `asdf.schema.load_schema`.


braingram:
jwst failures are known and due to a requirement for unreleased stcal changes. See: https://github.com/spacetelescope/jwst/pull/7874

PR #1629, opened 2023-08-16
define python envs in github ci

braingram:
Updates tox.ini to define used environments as tox 4.9 does not allow 'on-the-fly' environment creation unless the environment is prefixed with `pyXX`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,17 @@
 [tox]
-env_list = py{39,310}{,-compatibility,-coverage}{,-parallel}
+env_list =
+    compatibility
+    coverage
+    py{39,310,311}{,-compatibility,-coverage,-jsonschema}{,-parallel}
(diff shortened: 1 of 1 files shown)


braingram:
The stdatamodels and jwst failures are unrelated.
Same stdatamodels failures are seen in that repo: https://github.com/spacetelescope/stdatamodels/actions/runs/5884253404/job/15958641295?pr=195
JWST failures have been occurring prior to this PR: https://github.com/asdf-format/asdf/actions/runs/5825732254/job/15797991911?pr=1623 (and are due to an stdatamodels bug fixed but not released: https://github.com/spacetelescope/stdatamodels/pull/184)

WilliamJamieson (approved):

braingram:
Thanks! @WilliamJamieson

Issue #1607, opened 2023-07-27
Re-add pyyaml to devdeps testing when it gets cython 3 compatibility


braingram:
#1605 disabled pyyaml devdeps testing as it consistently fails to build (due to cython 3 incompatibility)

When it builds it should be re-added to the devdeps:
https://github.com/asdf-format/asdf/pull/1605/files#diff-2b4945591edfeaa4cf4d3f155e66d4b43d1bda7a55d881d5cf3107f1b05abbbcR9

PR #1633, opened 2023-08-18
Clean up test dependencies, add python 3.12 testing

braingram:
~After https://github.com/asdf-format/asdf/pull/1594 is sorted (which at the moment removes the gwcs test dependency) this PR can be finished and brought out of draft.~

~EDIT: based off changes in https://github.com/asdf-format/asdf/pull/1650 which fixes 1 test for python 3.12.~

The changes in this PR clean up the `test` dependencies. Some likely targets are:
- gwcs: looks to be entirely unused
- asdf-astropy: used for a handful of outdated tests and for some tests of 'nested' objects which could be handled by custom objects
- astropy: mostly incorrect usage now that serialization has moved to asdf-astropy

This also adds 3.12 testing which will currently fail because of aiohttp (required by fsspec). See https://github.com/asdf-format/asdf/pull/1651 for results with fsspec removed as a test dependency.

This also re-adds pyyaml to devdeps testing as the cython issues appear to be addressed:
Fixes: https://github.com/asdf-format/asdf/issues/1607

### DIFF
--- a/asdf/_tests/test_api.py
+++ b/asdf/_tests/test_api.py
@@ -7,7 +7,6 @@ import sys
 
 import numpy as np
 import pytest
-from astropy.modeling import models
 from numpy.testing import assert_array_equal
 
 import asdf
@@ -115,23 +114,6 @@ def test_atomic_write(tmp_path, small_tree):
 
     with asdf.open(tmpfile, mode="r") as ff:
         ff.write_to(tmpfile)
-
-
-def test_overwrite(tmp_path):
-    """
-    This is intended to reproduce the following issue:
-    https://github.com/asdf-format/asdf/issues/100
-    """
-    tmpfile = str(tmp_path / "test.asdf")
-    aff = models.AffineTransformation2D(matrix=[[1, 2], [3, 4]])
-    f = asdf.AsdfFile()
-    f.tree["model"] = aff
-    f.write_to(tmpfile)
-    model = f.tree["model"]
-
-    ff = asdf.AsdfFile()
-    ff.tree["model"] = model
-    ff.write_to(tmpfile)
 
 
 def test_default_version():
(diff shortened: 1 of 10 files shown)


asdf/_tests/test_api.py
```diff
@@ -119,23 +118,6 @@ def test_atomic_write(tmp_path, small_tree):
         ff.write_to(tmpfile)
 
 
-def test_overwrite(tmp_path):
```
braingram, line 122: The issue mentioned for this test:
https://github.com/asdf-format/asdf/issues/100

describes that it was added to try and replicate an issue on windows. The test did not replicate the issue: https://github.com/asdf-format/asdf/pull/302#issuecomment-320997092
yet the test was left in place.

asdf/_tests/test_generic_io.py
```diff
@@ -394,7 +393,7 @@ def test_invalid_obj(tmp_path):
 
 
 def test_nonseekable_file(tmp_path):
-    base = io.IOBase
+    base = io.FileIO
```
braingram, line 396: I'm not sure why `IOBase` wasn't failing with 3.12.rc1 but it fails with 3.12.dev. I don't see any description of this change in the release notes or PRs. My suspicion is the use of `IOBase` here was working in part because of a bug that is fixed in 3.12 (as providing a filename to `IOBase.__init__` doesn't really make sense so I could see how this could be fixed as bug during related `io` cleanup).

braingram:
stdatamodels CI failing due to crds context update. This should (at least partially) be addressed in: https://github.com/spacetelescope/stdatamodels/pull/219
I also see dynamicdq warnings but these also appear in stdatamodels main:
https://github.com/spacetelescope/stdatamodels/actions/runs/6265797233/job/17015381989

braingram:
Now if only RTD would keep up with what appears to be a caffeinated Slavich.

Issue #752, opened 2020-02-14
Add warning when schema $ref is a tag


eslavich:
It looks to me like `$ref` values were intended to be URIs or URI fragments, and it's only a quirk of the load_schema function in this library that tag values are resolved at all.  Consider adding a check to the pytest plugin that fails when `$ref` is set to a tag, and a deprecation warning to load_schema so that we can encourage correct use of `$ref`.

jdavies-st:
And this is because the `tag_to_url_mapping and `url_mapping` are always chained together and operate on both tags and URLs?

PR #1624, opened 2023-08-10
Eslavich simpler integer type

braingram:
This PR includes changes that were in https://github.com/asdf-format/asdf/pull/1527 re-based to https://github.com/asdf-format/asdf/pull/1625

Closes #1621
Closes #1347

Issue #1598, opened 2023-07-19
asdf does not follow version conventions defined in standard


braingram:
The ASDF standard docs describe how a "library should gracefully handle version numbers that are greater than those understood by the library":
https://asdf-standard.readthedocs.io/en/latest/versioning.html#handling-version-mismatches

However the asdf library does not follow these conventions for tagged objects supported by extensions. For example, suppose we have a tag `asdf://example.com/tags/foo` and 4 versions, `1.0.0`, `1.1.0`, `1.1.1`, `2.0.0`.

We can test the 3 types of mismatches mentioned in the standard docs:

- major version mismatch: write with `2.0.0`, read with only `1.0.0` support
- minor version mismatch: write with `1.1.0`, read with only `1.0.0` support
- patch version mismatch: write with `1.1.1`, read with only `1.0.0` support

In all 3 cases, the object is returned `raw` and 2 warnings are issued (similar to the following in all cases):
```
AsdfConversionWarning: asdf://example.com/tags/foo-2.0.0 is not recognized, converting to raw Python data structure
AsdfWarning: File was created with extension URI 'asdf://example.com/extensions/foo-2.0.0', which is not currently installed
```

However the standard states that:
- major version mismatch: should be an error
- minor version mismatch: should be a warning and an attempt to convert
- patch version mismatch: no warning, attempt to convert

An example script that demonstrates the above is as follows:

```python
import contextlib
import io

import asdf


def dump(tree):
    bs = io.BytesIO()
    asdf.AsdfFile(tree).write_to(bs)
    bs.seek(0)
    return bs.read()


def load(bs):
    bs_io = io.BytesIO(bs)
    with asdf.open(bs_io) as af:
        return af.tree


class Foo:
    pass


class FooV1Converter(asdf.extension.Converter):
    tags = ["asdf://example.com/tags/foo-1.0.0"]
    types = [Foo]

    def to_yaml_tree(self, obj, tag, ctx):
        return {'version': '1'}

    def from_yaml_tree(self, node, tag, ctx):
        return Foo()


class FooV1p1Converter(asdf.extension.Converter):
    tags = ["asdf://example.com/tags/foo-1.1.0"]
    types = [Foo]

    def to_yaml_tree(self, obj, tag, ctx):
        return {'version': '1.1'}

    def from_yaml_tree(self, node, tag, ctx):
        return Foo()


class FooV1p1p1Converter(asdf.extension.Converter):
[...]

Issue #1555, opened 2023-06-01
Consider option for allowing more graceful loading of files with failed conversions


braingram:
For files that contain portions of invalid data (such as those generated with out-of-date extensions or older unsupported versions) it might be helpful to have an option to allow failed conversions (and pass through raw values for those failed conversions).

Attempting to open the following invalid example file:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 3.0.0.dev265+g0c44742c}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension._manifest.ManifestExtension
    extension_uri: asdf://asdf-format.org/core/extensions/core-1.5.0
    software: !core/software-1.0.0 {name: asdf-astropy, version: 0.3.0}
a: [1, 2, 3]
q: !unit/quantity-1.1.0 {datatype: float64, missing_unit: True}
...
```

results in ```KeyError: 'value'``` (as the quantity is invalid, note that does require disabling `validate_on_read` to skip the failure caused by the invalid yaml).

However a user might want to read `a` in the above example. For the above example reading the text would be one option but if `a` were more complicated (for example an array with the data contained in a block) reading the value from the text might not be possible.

This came up when @schlafly attempted to extract data from a ROMAN file generated with older versions of the ROMAN software.

braingram:
@sydduckworth what do you think about switching the default for this to `True` for the next major release? We may need to do some preparation of downstream test suites but I'm wondering if this option being enabled (by default) is a more helpful option for users.

braingram:
During conversation with @tapastro this came up and he judged the warning output to be more helpful for an example where a file produced in a new environment was loaded in an older one. For that file the error produced halted execution, left the user with a completely (rather than partially) non-functional AsdfFile instance and asdf did not reach the point where the extension versions are checked so no warnings of "you have an old package" were displayed.

PR #1637, opened 2023-08-29
Remove legacy extension api

### LINKED ISSUE #1622: Remove internal use of legacy extension API
Will have to wait for:
- https://github.com/asdf-format/asdf/issues/1621
- https://github.com/asdf-format/asdf/pull/1537

braingram:
Builds on changes from:
- #1625
- #1624

Some notes so far:
- with this PR the new-style core extensions now assume the `BuiltinExtension` legacy class name (so warnings should not be seen when a file is opened that was produced with the legacy extension api)
- this PR adds a warning for resolving `$ref` using a tag (not a tag uri) (see: https://github.com/asdf-format/asdf/issues/752) ~there are several uses in stdatamodels/jwst which should be updated~ EDIT: stdatamodels uses fixed in: https://github.com/spacetelescope/stdatamodels/pull/201
- the old type system was more lax on conversion (as noted in other places including: https://github.com/asdf-format/asdf/issues/1598). One issue that came up here was related to failed deserialization. See below.

Closes: https://github.com/asdf-format/asdf/issues/752
Closes #1622

Deserialization types vs converters
-----
The old type system caught TypeError raised on deserialization and turned it into a warning:
https://github.com/asdf-format/asdf/blob/53b87b4d628305db4e46eec21c63d2ada047afd3/asdf/yamlutil.py#L343-L349
The new converters do not catch TypeError:
https://github.com/asdf-format/asdf/blob/53b87b4d628305db4e46eec21c63d2ada047afd3/asdf/yamlutil.py#L311-L315
There were a few tests that relied on this behavior. However it appears that the `raw` value is only passed on when a `TypeError` is raised (all other exceptions bubble up and result in a failure to load the file). I see no documentation of this (looking at the 2.14.4 docs which have all descriptions of the old type system) and it seems like an inconsistent way to allow partial conversion failures during open. As we already have an open issue for discussing options for partial/graceful loading on failed conversions (https://github.com/asdf-format/asdf/issues/1555) I vote for not replicating this undocumented behavior of the type system and address the issue in a separate PR.


braingram:
dkist also contains at least 1 ref by tag:
https://github.com/DKISTDC/dkist/blob/c7e61155a2ac73d6c0bd0b389c5aea263b17076e/dkist/io/asdf/resources/schemas/varying_celestial_transform-1.0.0.yaml#L33C1-L33C57

PR #1642, opened 2023-09-11
Update 2.15.x branch

braingram:
Backports:
- #1620 
- #1629 
- #1627 
- #1639 
- #1641


braingram:
jwst downstream failures and known and unrelated. See https://github.com/spacetelescope/jwst/pull/7874

PR #1651, opened 2023-09-22
TEST: Py312

braingram:
Removes fsspec from our test dependencies to allow the python 3.12 tests to run.

Based off changes in: https://github.com/asdf-format/asdf/pull/1633

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,8 +2,7 @@
 env_list =
     compatibility
     coverage
-    devdep{,-parallel}
-    py{39,310,311,312}{,-compatibility,-coverage,-jsonschema}{,-parallel}
+    py{39,310,311,312}{,-compatibility,-coverage,-jsonschema}{,-devdeps}{,-parallel}
     asdf{-standard,-transform-schemas,-unit-schemas,-wcs-schemas,-coordinates-schemas,-astropy}
     gwcs
     jwst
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -2,7 +2,8 @@
 env_list =
     compatibility
     coverage
-    py{39,310,311}{,-compatibility,-coverage,-jsonschema}{,-parallel}
+    devdep{,-parallel}
```
zacharyburnett, line 5:
```suggestion
    devdeps{,-parallel}
```

tox.ini
```diff
@@ -2,7 +2,8 @@
...
+    py{39,310,311,312}{,-compatibility,-coverage,-jsonschema}{,-parallel}
```
zacharyburnett, line 6:
```suggestion
    py{39,310,311,312}{,-compatibility,-coverage,-jsonschema}{,-devdeps}{,-parallel}
```

braingram: Thanks!

braingram:
@pllim that appears to be an issue with `dateutil` fixed: https://github.com/dateutil/dateutil/pull/1285
It's likely not seen here as nothing requires the optional `isoduration` dependency (which is using `dateutil`).

pllim:
We don't either, but it got imported by asdf like this:

```
asdf/_jsonschema/_format.py:494: in <module>
    import isoduration
```

Weird.

PR #1665, opened 2023-10-24
use np.memmap instead of np.core.memmap

braingram:
`np.core` is deprecated in numpy 2.0 and causing failures in the devdeps tests:
https://github.com/asdf-format/asdf/actions/runs/6619133498/job/17979194872?pr=1664#step:10:607
This PR replaces all references to `np.core.memmap` with `np.memmap` to fix these errors.


braingram:
As these are only test and docs changes I'm merging this to clean up the devdeps.

Issue #884, opened 2020-11-05
segfault when loading array


lgarrison:
The following code causes a segfault on my machine:
```python
import numpy as np
import asdf

asdf.AsdfFile({'data':np.arange(100)}).write_to('tmp.asdf')
with asdf.open('tmp.asdf') as f:
  d = f.tree['data'][:]
print(d)
```

I'm on Python 3.8, Numpy 1.19.4, asdf 2.7.1 from pip (fresh conda env).  Also fails out of asdf master.  The segfault happens on the last line, but only when trying to force a non-lazy load with `[:]`.

I'm probably abusing asdf with this syntax!  I was experimenting to see if I could use `[:]` as shorthand for `lazy_load=False`, `copy_arrays=True`.  But I think calling `copy()` achieves the same thing (possibly at the minor expense of a temporary, extra copy?).  Or I should stop being lazy and just use the proper arguments!

eslavich:
I think we can do better than a segfault here, but `[:]` just gives you a view over the same memory mapped ndarray, so it also loses access to the data when the file is closed.  `copy()` is probably your best option -- I _think_ that will only result in one copy of the ndarray in process memory, since the original ndarray lives in the page cache.

Let's keep this issue open so that we remember to replace the segfault with a reasonable exception.

jdavies-st:
The case is handled when it's not a view.

```
In [10]: import numpy as np
    ...: import asdf
    ...: 
    ...: asdf.AsdfFile({'data':np.arange(100)}).write_to('tmp.asdf')
    ...: with asdf.open('tmp.asdf') as f:
    ...:   d = f.tree['data']
    ...: 

In [11]: print(d)
<array (unloaded) shape: [100] dtype: int64>

In [13]: d += 1
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-13-2237cfe673cf> in <module>
----> 1 d += 1

~/miniconda3/envs/jwst/lib/python3.8/site-packages/asdf/tags/core/ndarray.py in __operation__(self, *args)
    517 def _make_operation(name):
    518     def __operation__(self, *args):
--> 519         return getattr(self._make_array(), name)(*args)
    520     return __operation__
    521 

~/miniconda3/envs/jwst/lib/python3.8/site-packages/asdf/tags/core/ndarray.py in _make_array(self)
    266 
    267             self._array = np.ndarray(
--> 268                 shape, dtype, block.data,
[...]

PR #1668, opened 2023-10-26
Allow memmapped arrays to keep the backing mmap file open

### LINKED ISSUE #1334: Prevent segfaults and issue an informative error for array access on closed files for array containing objects.
```python
import asdf
import astropy.units
import numpy

tree = {'arr': astropy.units.Quantity(numpy.zeros(10), 'm')}
fn = 'test.asdf'
asdf.AsdfFile(tree).write_to(fn)

with asdf.open(fn) as af:
    a = af['arr']

print(a[:])
```
Results in `Segmentation fault: 11`.

Note that an error is already issued for stock ndarrays.
https://github.com/asdf-format/asdf/blob/5b6c232c3069fddda2b3376a56771ebc41fd01c7/asdf/block.py#L1221-L1223

### LINKED ISSUE #1530: Calling update with memmapped data can create invalid data in memmap views
If an `update` causes a memmapped block to be moved, any views taken of the array that uses the block will point to invalid data.

Running the following example code (that adds a large amount of text to the tree to cause the memmapped blocks to be moved):

```python
import asdf
import numpy


fn = 'test.asdf'
a = numpy.zeros(10, dtype='uint8')
b = numpy.ones(10, dtype='uint8')

af = asdf.AsdfFile({'a': a, 'b': b})
af.write_to(fn)

with asdf.open(fn, mode='rw') as af:
    va = af['a'][:3]
    print(f"A: {af['a'][:]}")
    print(f"b: {af['b'][:]}")
    print(f"va: {va}")
    af['c'] = 'a' * 10000
    af.update()
    print(f"A: {af['a'][:]}")
    print(f"b: {af['b'][:]}")
    print(f"va: {va}")
```

Produces the following output which shows that the view of 'a', here 'va', is incorrect following the update.

```bash
A: [0 0 0 0 0 0 0 0 0 0]
b: [1 1 1 1 1 1 1 1 1 1]
va: [0 0 0]
A: [0 0 0 0 0 0 0 0 0 0]
b: [1 1 1 1 1 1 1 1 1 1]
va: [97 97 97]
```

braingram:
asdf uses a mmap to memory map files to produce np.memmap arrays (when copy_arrays is True, the current default). np.memmap does not check if the backing mmap is closed before accessing array data and asdf closes the mmap when the asdf file is closed. This results in segfaults (see issue #1334).

This PR changes when happens to the mmap when the asdf file is closed. asdf will no longer force closure of the mmap. This will result in an open file pointer if any array still references the mmap (the mmap will likely be closed by the garbage collector when all arrays are collected which matches the behavior of [np.memmap](https://numpy.org/doc/stable/reference/generated/numpy.memmap.html#numpy-memmap) which does not have an API for closing the underlying mmap).

Windows does not allow a file to be trunacated (as can occur during `update`) if it is open more than once. The PR adds an os check and closes the mmap on windows which (with these changes) passes all tests:
https://github.com/asdf-format/asdf/actions/runs/6657397904/job/18092170804?pr=1668
including the regression test added with this PR that segfaults on linux and mac.

The failure of the `build` job is unrelated and addressed in: https://github.com/asdf-format/asdf/pull/1670
The python 3.12 failure is unrelated (due to `aiohttp` having no released version that supports 3.12).

~The stdatamodels failure should be addressed prior to this PR being merged. The failures are all related to the change in the number of open files (which stdatamodels has specific tests for). https://github.com/spacetelescope/stdatamodels/pull/232 addresses these tests.~ EDIT: https://github.com/spacetelescope/stdatamodels/pull/232 is merged

Fixes #1334 
Fixes #1530


zacharyburnett (approved):
LGTM (though I'm not familiar with these inner workings). Tests pass

braingram:
Thanks for the reviews! Marking this as draft until https://github.com/spacetelescope/stdatamodels/pull/232 can be merged then this can be brought out of draft and merged.

eslavich (approved):

jdavies-st:
I wish I'd been able to review this before this was merged.  I just ran into this issue that the `with` context manager no longer manages resources, i.e. the file handles to memmaps are not closed.  So the same surprising, buggy and unexpected behavior as astropy.io.fits.

We early developers of ASDF spent a lot of time and effort trying to make the `with` context manager work as expected, especially @drdavella.  See

#111 
#355
#407
https://github.com/astropy/astropy/issues/7404#issuecomment-481856255 

Clearly numpy views on the closed memmap as in

#884 
#1334 
#1530 

were not handled properly.  It's never good to get a segfault.  So this PR opted for a leak instead of a crash tradeoff.

But I can't help but wonder if we can return something wrapped for memmap objects, and make sure views are handled gracefully when the underlying memmap has been closed, as a `with` context manager implies?  I will work on this.

Happy to get feedback.

PR #1670, opened 2023-10-26
build with python 3.11 (not 3.12)

braingram:
The OpenAstronomy build workflow uses python version `3.x`. This is now defaulting to `3.12`. As `aiohttp` does not yet have a stable python 3.12 version (and this is a test dependency installed as part of the build testing run) this is causing build to fail:
https://github.com/asdf-format/asdf/actions/runs/6657397904/job/18092159951?pr=1668

This PR forces the build version to 3.11.


braingram:
@zacharyburnett it wouldn't surprise me if this impacts other packages that use `3.x` and/or the OpenAstronomy build workflow.

zacharyburnett (approved):

PR #1681, opened 2023-11-13
add pre-commit workflow to replace pre-commit-ci

braingram:
Since pre-commit-ci will no longer be auto-fixing PRs this PR also adds:
- a workflow that runs pre-commit (the branch protections will be updated when this PR is merged to make this a required job)
- a pull request template describing running pre-commit, pytest, etc

I also added an issue template to this PR since it seems like a good thing to provide.

Fixes https://github.com/asdf-format/asdf/issues/1679

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
+    - uses: pre-commit/action@v3.0.0
   core:
+    needs: [pre-commit]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
+    - uses: pre-commit/action@v3.0.0
```
WilliamJamieson, line 28: See https://github.com/asdf-format/asdf/issues/1679#issuecomment-1810737318 as an alternative.

zacharyburnett (approved):
@braingram I still do not think the fact that outside collaborators can run CI without approval warrants disabling `pre-commit.ci` entirely, but I'll leave it to your prerogative as maintainer

PR #1682, opened 2023-11-14
disable `pre-commit.ci` autofixing pull request branches

zacharyburnett:
applied suggestions from https://github.com/asdf-format/asdf/issues/1679#issuecomment-1810737318

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,7 +1,7 @@
 name: Changelog
 
 on:
-  pull_request:
+  pull_request_target:
     types: [labeled, unlabeled, opened, synchronize, reopened]
 
 # Only cancel in-progress jobs or runs for the current workflow
@@ -14,6 +14,7 @@ concurrency:
 
 jobs:
   changelog:
+    if: github.actor != 'pre-commit-ci[bot]'
     name: Confirm changelog entry
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
       - '*.x'
     tags:
       - '*'
-  pull_request:
+  pull_request_target:
 
 # Only cancel in-progress jobs or runs for the current workflow
 #   This cancels the already triggered workflows for a specific PR without canceling
@@ -21,6 +21,7 @@ concurrency:
 
 jobs:
   core:
+    if: github.actor != 'pre-commit-ci[bot]'
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
     with:
       submodules: false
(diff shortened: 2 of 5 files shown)


.github/workflows/changelog.yml
```diff
@@ -1,7 +1,7 @@
 name: Changelog
 
 on:
-  pull_request:
+  pull_request_target:
```
braingram, line 4: This unnecessarily [exposes secrets to these jobs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) and means that PRs updating these workflows cannot be evaluated until after they are merged making managing these files more difficult.

braingram (changes requested):
I am against re-enabling pre-commit-ci unless they fix the issue. The benefits it provides (autofixing prs and automatically updating pre-commit versions) do not seem worth the risk.

.github/workflows/changelog.yml
zacharyburnett: that's fair, it makes workflow development difficult

WilliamJamieson:
> I am against re-enabling pre-commit-ci unless they fix the issue. The benefits it provides (autofixing prs and automatically updating pre-commit versions) do not seem worth the risk.

I fail to see how this is a risk? ASDF uses GitHub's freely available runners and all the CI that can be triggered from a PR has read only permissions (or should have read-only permissions).

CagtayFabry:
> I am against re-enabling pre-commit-ci unless they fix the issue. The benefits it provides (autofixing prs and automatically updating pre-commit versions) do not seem worth the risk.

I am not sure if this it what you mean, but you can have at least one of the benefits (automatically updating pre-commit versions) without enabling the other (autofixing PRs) by changing the [pre-commit configuration ](https://pre-commit.ci/#configuration)in `.pre-commit-config.yaml`, e.g. by adding

```yaml
ci:
  autofix_prs: false
  autoupdate_schedule: weekly
```

braingram:
Thanks for the suggestion. Unfortunately even with `autofix_prs` disabled adding a `pre-commit.ci autofix` comment will trigger pre-commit-ci to fix the PR which will then allow the CI to run (this comment can be added by any account, even one brand new to GitHub). It would be nice if there were a way to disable autofixing or limit permissions for auto fixes to existing maintainers but the pre-commit-ci developers aren't interested in making that change.

.github/workflows/changelog.yml
braingram: Reading into this a bit more the use of `pull_request_target` also provides a read/write `GITHUB_TOKEN` to the job. The [documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) warns that jobs using `pull_request_target` should not check out untrusted code (which occurs in the similarly modified `ci.yml` job). I'm going to close this PR.

CagtayFabry:
Thank you for the insight @braingram , that is unfortunate to hear

PR #1687, opened 2023-11-15
Display value for all childless nodes during `info`

### LINKED ISSUE #1686: asdftool repr for Time objects
Using `asdftool info <file>` on an ASDF file containing astropy.time.Time objects produces the following (excerpted) output:
```
  │ │ ├─type (str): WFI_IMAGE
  │ │ ├─start_time (Time) # UTC exposure start time
  │ │ ├─mid_time (Time) # UTC exposure mid time
  │ │ ├─end_time (Time) # UTC exposure end time
```
Is it possible to add some kind of repr to this to display the time (maybe in ISOT or some other format?).

braingram:
For childless (leaf) nodes, show the `str` value (when `show_values` is `True`). This allows objects like `Time` to display their value.

For example, on main when running `asdftool info some_roman_file.asdf` the output will contain:
```
  │ │ ├─start_time (Time)
  │ │ ├─mid_time (Time)
  │ │ ├─end_time (Time)
```

With this PR the same file contains the output:
```
  │ │ ├─start_time (Time): 2021-01-01T00:00:00.000
  │ │ ├─mid_time (Time): 2021-01-01T00:01:16.020
  │ │ ├─end_time (Time): 2021-01-01T00:02:32.040
```

This PR also deprecates the now unused `asdf.util.is_primitive`.

Fixes #1686

### DIFF
--- a/asdf/_display.py
+++ b/asdf/_display.py
@@ -11,7 +11,6 @@ import numpy as np
 
 from ._node_info import create_tree
 from .tags.core.ndarray import NDArrayType
-from .util import is_primitive
 
 __all__ = [
     "DEFAULT_MAX_ROWS",
@@ -252,11 +251,12 @@ class _TreeRenderer:
 
     def _render_node_value(self, info):
         rendered_type = type(info.node).__name__
-        if is_primitive(info.node) and self._show_values:
-            return f"({rendered_type}): {info.node}"
 
         if isinstance(info.node, (NDArrayType, np.ndarray)):
             return f"({rendered_type}): shape={info.node.shape}, dtype={info.node.dtype.name}"
+
+        if not info.children and self._show_values:
+            return f"({rendered_type}): {info.node}"
 
         return f"({rendered_type})"
 
--- a/asdf/_tests/test_info.py
+++ b/asdf/_tests/test_info.py
@@ -633,6 +633,9 @@ class RecursiveObjectWithInfoSupport:
     def __asdf_traverse__(self):
         return {"the_meaning": self.the_meaning, "clown": self.clown, "recursive": self.recursive}
 
+    def __str__(self):
+        return "rec ref"
+
 
 def test_recursive_info_object_support(capsys, tmp_path):
     tempdir = pathlib.Path(tempfile.mkdtemp())
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ The ASDF Standard is at v1.6.0
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Fix bug in ``asdftool diff`` for arrays within a list [#1672]
+- For ``info`` and ``search`` show ``str`` representation of childless
+  (leaf) nodes if ``show_values`` is enabled  [#1687]
+- Deprecate ``asdf.util.is_primitive`` [#1687]
 
 3.0.0 (2023-10-16)
 ------------------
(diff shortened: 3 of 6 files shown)


asdf/_tests/test_info.py
```diff
@@ -633,6 +633,9 @@ def __init__(self):
     def __asdf_traverse__(self):
         return {"the_meaning": self.the_meaning, "clown": self.clown, "recursive": self.recursive}
 
+    def __str__(self):
+        return "rec ref"
```
braingram, line 637: This was added because with this PR (and without this change) the assert in `test_recursive_info_object_support`:
https://github.com/asdf-format/asdf/blob/ecb4d4f6308738c4de0da8eec5bd272e90f3954e/asdf/_tests/test_info.py#L650
would fail due to the output line being truncated due to this test object (`RecursiveObjectWithInfoSupport`) not having a `__str__`.

asdf/_display.py
```diff
@@ -236,7 +235,7 @@ def _render_node(self, info, active_depths, is_tail):
         )
 
         if info.info is not None:
-            line = line + format_faint(format_italic(" # " + info.info))
+            line = line + format_faint(format_italic(" # " + info.info.strip()))
```
braingram, line 238: This change is to deal with `info` that contains newlines like the [exposure](https://github.com/spacetelescope/rad/blob/186144b27f05bf654b5cab2331fb66d15f932cb1/src/rad/resources/schemas/exposure-1.0.0.yaml#L7) in RAD. See https://github.com/asdf-format/asdf/issues/1686#issuecomment-1814754907

braingram: tldr; yaml is messy, I think the schemas would be better off not using literal `|` blocks and the other multiline options seems to strip embedded newlines

I think this is already being done (in some places). For example:
```
  │ │ ├─obs_id (str): 0000101001001001001011010001 # Programmatic observation identifier. The format is 'PPPPPCCAAASSSOOOVVVggsaaeeee' where 'PPPPP' is the Program, 'CC' is the execution plan, 'AAA' is the pass, 'SSS' is the segment, 'OOO' is the Observation, 'VVV' is the Visit, 'gg' is the visit file group, 's' is the visit file sequence, 'aa' is the visit file activity, and 'eeee' is the exposure ID. The observation ID is the complete concatenation of visit_id + visit_file_statement (visit_file_group + visit_file_sequence + visit_file_activity) + exposure.
```
Is pulling in the info from this multi-line title here: https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/observation-1.0.0.yaml#L10-L16

I think the issue is the difference between:
```yaml
this: description
        with 2 lines
```
and
```yaml
this: |
    description
    with 2 lines
```
The first is read by pyyaml as a single string with no newlines:
```python
{'this': 'description with 2 lines'}
```
whereas the second preserves the embedded and trailing newlines
```python
{'this': 'description\nwith 2 lines\n'}
```

I think that the use of these literal blocks (with the `|`) might be considered bugs in the schema (these could be 'folded', see below, or multiline formatted like the first example).

To further complicate things the yaml "folded" style retains only the trailing newline:
```yaml
this: >
    description
    with 2 lines
```
```python
{'this': 'description with 2 lines\n'}
```
Although this can be controlled using a "Block Chomping Indicator"
```yaml
this: >-
    description
    with 2 lines
```
```python
{'this': 'description with 2 lines'}
```

braingram: I'm in favor of that change or something similar to truncate long info strings. I just checked and the "info" portion of the line is included in the "max_cols" check which truncates long lines.

I dropped the `strip` change from this PR to keep it focused on fixing the `Time` issue and opened a separate issue to track the line formatting: https://github.com/asdf-format/asdf/issues/1712

Issue #798, opened 2020-05-15
Consider replacing vendorized RangeHTTPServer


eslavich:
We've included a vendorized copy of RangeHTTPServer for testing with an http server, but it seems like there must be a more standard tool we can use for that purpose.  @jdavies-st suggests requests-mock as a possibility.

jdavies-st:
While we're at, we've also vendorized `extern.decorators` from Sunpy which we don't use either.  That can probably just be removed.

Issue #1522, opened 2023-04-19
asdf-standard submodule no longer checked out in CI


braingram:
It looks like this switched the CI to not checkout submodules.
https://github.com/asdf-format/asdf/commit/3a577dc7e0d329c0f3b764b94b9114d2e91965a1

I believe that this means that the asdf-standard submodule is no longer checked out (as seen in this action log):
https://github.com/asdf-format/asdf/actions/runs/4718032521/jobs/8367210716#step:2:5

If the submodule is not checked out reference files will not be available for tests in:
https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_reference_files.py

braingram:
Looks like the submodule was removed in this commit:
https://github.com/asdf-format/asdf/commit/f6559d73fe80c0e13d4ee65c037ef51591522ad4

It does not look like the reference files are packaged with the standard nor are there tests in asdf-standard that test these reference files.

The docs also mention this non-existent submodule: https://asdf.readthedocs.io/en/stable/asdf/developer_versioning.html#update-the-asdf-standard-submodule-commit-pointer

braingram:
The docs have been fixed, all that remains for this issue are the reference file tests.

PR #1703, opened 2023-12-13
Cleanup asdf._tests._helpers

braingram:
# Description

This PR cleans up `asdf._tests._helpers` removing all but `assert_tree_match`. This PR removes:

- `get_test_data_path` replaced with a test fixture using `importlib.resources` (data for commands tests were moved to `_tests.data` to simplify paths)
- `assert_roundtrip_tree` replaced with `asdf.testing.helpers.assert_roundtrip_object` and more explicit `assert` statements to compare the feature being tested in any given test
- `yaml_to_asdf` replaced with `asdf.testing.helpers.yaml_to_asdf`
- `get_file_sizes` replaced with uses of `pathlib.Path` and calls to `stat().st_size`
- `display_warnings` unused, removed
- `assert_no_warnings` all uses removed, tests already run with a filter that turns warnings into errors, a pytest configuration was added to `_tests/conftest` to make sure that this filter applies to all runs (even through through `pytest --pyargs asdf` as is run during the `Build` CI test)
- `_assert_extension_type_correctness` unused, removed

Although not removed `assert_tree_match` was simplified (removing unused arguments and other unused code).

`assert_roundtrip_tree` performed a number of tests for round tripping (that are not all performed by `testing.helpers.roundtrip_object`):
- read/write to `BytesIO`, tested with `assert_roundtrip_object`
- read/write to a real file, low-level file access modes (like real files and BytesIO) are tests in [test_generic_io.py](https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_generic_io.py)
- read from an http range server, http access is tested in [test_generic_io.py](https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_generic_io.py) and via `fsspec` in [test_asdf.py](https://github.com/asdf-format/asdf/blob/11f5ed5d5dac034934e9c9ec1eedd5f326655708/asdf/_tests/test_asdf.py#L376)
- read with lazy loading an no memmapping, low-level file access is tests in [test_generic_io.py](https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_generic_io.py) and memmapping is tested in [test_array_blocks.py](https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_array_blocks.py)
- read with memmapping memmapping is tested in [test_array_blocks.py](https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_array_blocks.py)

The simplification of `assert_tree_match` will further break the [test_reference_files.py](https://github.com/asdf-format/asdf/blob/main/asdf/_tests/test_reference_files.py) tests that have not been running since Feb 2022. When run manually (by checking out asdf-standard as if it were still a submodule) these fail without this PR and likely these should be updated and possibly moved to the asdf-standard test suite (which is run as part of the asdf CI). There is already an open issue for these tests: https://github.com/asdf-format/asdf/issues/1522

Removal of `assert_roundtrip_tree` allowed for the removal of the bundled `extern.RangeHTTPServer` (removed in this PR).

Fixes https://github.com/asdf-format/asdf/issues/798

# Checklist:

- [ ] tests ran successfully


zacharyburnett (approved):
this looks good to me and simplifies the test fixtures; can it be rebased?

braingram:
Thanks for the review! I will (attempt to) resolve the conflicts 😬

PR #1716, opened 2023-12-21
fix bug where a dict with an id causes contained refs to fail to resolve

### LINKED ISSUE #1715: Tree containing `id` causes reference resolution failure.
# Description of the problem

The existence of a dictionary with a key `id` anywhere "higher" in the tree than a reference causes the reference to fail to resolve.

# Example of the problem

```python
import asdf


other_tree = {
    'thing': 42,
}

tree = {
    'myref': {
        '$ref': 'external.asdf#/thing'
    }
}

asdf.AsdfFile(other_tree).write_to('external.asdf')

uri=asdf.util.filepath_to_url(os.path.abspath('main.asdf'))
af = asdf.AsdfFile(tree, uri=uri)
af.resolve_references()
assert af['myref'] == 42

tree['id'] = 'bogus'
af = asdf.AsdfFile(tree, uri=uri)
af.resolve_references()  # raises ValueError
```

# System information

asdf version: main
python version: 3.10
operating system: mac osx

braingram:
# Description

Remove use of `json_id` in `find_references` callback passed to `walk_and_modify`.

The `json_id` argument (which stores the most recently encountered 'id' value) is primarily (perhaps only) useful for schema parsing. However, `find_references` (called on the ASDF tree) uses the `json_id` value to construct the uri for the referred object. I see no reference to this behavior in:
https://datatracker.ietf.org/doc/html/rfc6901
https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03
https://datatracker.ietf.org/doc/html/rfc3986
and this can lead to reference resolution failure (see #1715).

Fixes #1715 

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@
   a file and ``AsdfPackageVersionWarning`` when installed extension/manifest
   package does not match that used to write the file [#1758]
 
+- Fix bug where a dictionary containing a key ``id`` caused
+  any contained references to fail to resolve [#1716]
+
 
 3.2.0 (2024-04-05)
 ------------------
--- a/asdf/_tests/_regtests/test_1715.py
+++ b/asdf/_tests/_regtests/test_1715.py
@@ -1,0 +1,28 @@
+import pytest
+
+import asdf
+
+
+def test_id_in_tree_breaks_ref(tmp_path):
+    """
+    a dict containing id will break contained References
+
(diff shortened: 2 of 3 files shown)


zacharyburnett (approved):
LGTM if tests pass

braingram:
The 2 downstream failures are unrelated to this PR and are addressed in: https://github.com/asdf-format/asdf/pull/1785

Issue #1723, opened 2024-01-03
asdf odd handling of `namedtuple` that prevents implementing a `Converter`


braingram:
Attempts to store a `namedtuple` instance in the tree result in behavior that doesn't match with other custom objects.

```python
import collections
import asdf

NT = collections.namedtuple("NT", ("a", "b"))
nt = NT(1, 2)
af = asdf.AsdfFile({"nt": nt})
print(type(af['nt']))
```
produces
```
/Users/bgraham/projects/240103_named_tuple/asdf/asdf/treeutil.py:360: AsdfWarning: Failed to serialize instance of <class '__main__.NT'>, converting to list instead
  warnings.warn(f"Failed to serialize instance of {type(node)}, converting to list instead", AsdfWarning)
<class 'list'>
```
The warning can be silenced by passing `ignore_implicit_conversion=True` when creating the `AsdfFile` instance.

The conversion to a list loses the by-name item access possible with `namedtuple`.

This prevents creation of any `Converter` to support a specific `namedtuple`.

If the following line is changed to use `type(node) == tuple`:
https://github.com/asdf-format/asdf/blob/334f6c387baebb1a9dd41c681d2b71806b84c3a8/asdf/treeutil.py#L368
it becomes possible to put `namedtuple` instances in the tree and to implement `Converter` instances for specific `namedtuple` classes. See the following:
```python
import collections
import asdf


NT = collections.namedtuple("NT", ("a", "b"))
nt = NT(1, 2)

nt_tag = "asdf://example.com/tags/nt-1.0.0"


class NTConverter:
    types = [NT]
    tags = [nt_tag]

    def to_yaml_tree(self, obj, tag, ctx):
        return obj._asdict()

    def from_yaml_tree(self, obj, tag, ctx):
        return NT(**obj)


class NTExtension:
    converters=[NTConverter()]
    tags=[nt_tag]
    extension_uri = "asdf://example.com/extensions/nt-1.0.0"


with asdf.config_context() as cfg:
    cfg.add_extension(NTExtension())

    af = asdf.AsdfFile({"nt": nt})
    assert type(af["nt"]) == NT
    af.write_to("test.asdf")

    with asdf.open("test.asdf") as af:
        assert type(af["nt"]) == NT
```

In summary, the current way that `namedtuple` is supported seems incomplete and hinders customization by users.

[...]

PR #1724, opened 2024-01-03
deprecate ignore_implicit_conversion and implicit_conversion

braingram:
# Description

This is work towards fixing #1723

This PR deprecates "implicit conversion" and the flag controlling warning messages when this behavior occurs (`ignore_implicit_conversion`). See the above issue for more details. In brief, this prevents extensions from adding support for `namedtuple` classes.

The removal of this feature (after the deprecation period, which this PR will start) will mean that attempting to add a `namedtuple` to the asdf tree (without implementing a converter for the `namedtuple`) will result in an error. The effectively drops the "support" we currently have for `namedtuple` (which does not roundtrip, loses all `namedtuple` specifics and saves these instances as lists). User code that wishes to continue to support `namedtuple` instances will need to:
- convert them all to lists prior to saving
- implement a converter for their `namedtuple` classes to allow roundtripping

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,8 @@
   ``AsdfFile.set_array_save_base`` and
   ``SerializationContext.set_array_save_base`` [#1753]
 
+- Deprecate ``ignore_implicit_conversion`` and "implicit conversion" [#1724]
+
 
 3.2.0 (2024-04-05)
 ------------------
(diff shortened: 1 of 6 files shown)


braingram:
Runs of romancal and jwst regtests will be run and this comment updated to check for any new `AsdfDeprecationWarnings` indicating a converted `named_tuple`.

perrygreenfield (approved):
LGTM

PR #1728, opened 2024-01-05
TEST: Lazy tree as default

braingram:
Downstream failures when turning `lazy_tree` on by default:

- weldx: Several comparisons in tests fail (see https://github.com/BAMWelDX/weldx/blob/ad4b8dc8d976c1306cbdfd409c84e73f288c5686/weldx/tests/asdf_tests/test_asdf_core.py#L348) due to an exact type comparison here: https://github.com/BAMWelDX/weldx/blob/ad4b8dc8d976c1306cbdfd409c84e73f288c5686/weldx/util/util.py#L205. These can all be fixed by adding another `type_equality`: https://github.com/BAMWelDX/weldx/blob/ad4b8dc8d976c1306cbdfd409c84e73f288c5686/weldx/util/util.py#L195
- asdf-astropy: one test fails due to the expectation that all ASDF file objects will be converted on load: https://github.com/astropy/asdf-astropy/blob/448542b99ff6a82b9aeb810e78fb253e2fbc155f/asdf_astropy/converters/table/tests/test_table.py#L144 this test can be updated to access the table (to trigger the conversion)
- dkist: `load_dataset` grabs a reference to part of an open asdf file (after opening it with `lazy_load=False`): https://github.com/DKISTDC/dkist/blob/e5cf0f9c598c56cd5bc1514d9149fac03cc9896f/dkist/dataset/loader.py#L159 so this should be updated to include `lazy_tree=False`. Similar issues exist for the `eit_dataset` and `small_visp_dataset` fixtures: https://github.com/DKISTDC/dkist/blob/e5cf0f9c598c56cd5bc1514d9149fac03cc9896f/dkist/conftest.py#L260
- jwst has issues similar to dkist. For example many `wfss` tests are failing due to handling of the reference file data in assign wcs where the file is closed after references to portions of the file are captured: https://github.com/spacetelescope/jwst/blob/b762c669639416d61b6e7d3facc5cbee2d6bc9b0/jwst/assign_wcs/nircam.py#L365C44-L365C51 Given the limited use of pure asdf files in jwst, stdatamodels could be updated to disable "lazy_tree" when opening an asdf file. A regression test run with a modified stdatamodels shows the usual random errors (msa_fail/nofail) and one seemingly unrelated unclosed file error: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1130/
- asdf-zarrr: the failures here are real as the new `AsdfDictNode` instance is not json serializable (more on this later)

# Checklist:

- [ ] tests ran successfully


asdf/_lazy_nodes.py
```diff
@@ -0,0 +1,162 @@
...
+        return obj
+    converter = extension_manager.get_converter_for_tag(tag)
+    data = value.data
+    if isinstance(data, dict):
+        data = AsdfDictNode(data, af_ref)
+    elif isinstance(data, list):
+        data = AsdfListNode(data, af_ref)
+    obj = converter.from_yaml_tree(data, tag, sctx)
+    if isinstance(obj, GeneratorType):
+        # TODO we can't quite do this for every instance
```
braingram, line 44: astropy transform converters return `GeneratorType` (transforms can refer to their own inverse). This works for all transform tests but does not work for `FractionWithInverse`.

asdf/_lazy_nodes.py
```diff
@@ -0,0 +1,162 @@
...
+                raise Exception("no ASDF for you!")
+            value_id = id(value)
+            if value_id in af._tagged_object_cache:
+                value = af._tagged_object_cache[value_id][1]
+            else:
+                obj = AsdfListNode(value, self._af_ref)
+                af._tagged_object_cache[value_id] = (value, obj)
+                value = obj
+            self[key] = value
+        elif type(value) in (dict, collections.OrderedDict):
```
braingram, line 100: gwcs uses `OrderedDict` and more formal support is needed (perhaps another `AsdfNode` subclass).

asdf/_lazy_nodes.py
```diff
@@ -0,0 +1,162 @@
...
+        return self.data
+
+
+class AsdfListNode(AsdfNode, collections.UserList, list):
+    def __init__(self, data=None, af_ref=None):
+        if data is None:
+            data = []
+        super().__init__(data, af_ref)
+
+    def __reduce__(self):
```
braingram, line 72: This is necessary due to the subclassing of `UserList` and `list`. This creates odd behavior where a deepcopy will duplicate items in the list (without this specific routing of `__reduce__`).

braingram: See https://github.com/asdf-format/asdf/pull/1728/commits/48262db4ec7c2ab998b0d89c8644169ae398c099

asdf/_lazy_nodes.py
braingram: See https://github.com/asdf-format/asdf/pull/1728/commits/48262db4ec7c2ab998b0d89c8644169ae398c099

asdf/_lazy_nodes.py
braingram: This was resolved in d10e15ae12dd921c22f30fba65fadbf9efc20b94

asdf/_lazy_nodes.py
braingram: This inheritance pattern is no longer used.

PR #1729, opened 2024-01-08
update test warning filter for pytest 8

braingram:
pytest 8 includes a change to how "pytest.warns" behaves (it now re-emits non-matched warnings). Running the asdf test suite against pytest 8 (rc1) there is one failing test test_asdf_asdf_SerializationContext_import_deprecation due to this change.

This PR updates the warning filters in test_asdf_asdf_SerializationContext_import_deprecation to work with pytest 8.
# Description

# Checklist:

- [ ] tests ran successfully


requirements-dev.txt
```diff
@@ -5,6 +5,8 @@ git+https://github.com/asdf-format/asdf-unit-schemas.git
 git+https://github.com/asdf-format/asdf-wcs-schemas
 git+https://github.com/yaml/pyyaml.git
 
+pytest>=0.0.dev0
```
zacharyburnett, line 8: should we run this in a different `dev` test?

zacharyburnett (approved):

requirements-dev.txt
braingram: Thanks for giving this a review.

That sounds like a great idea! It does seem bad to have all our devdeps fail just because pytest makes some change. I am planning to update https://github.com/asdf-format/asdf/pull/1731 and bring it out of draft after this PR is merged which would install pytest from dev (further increasing the risk of a change breaking all the tests).

Do you have a suggestion for how to break this out? Maybe update tox with a `-pytestdev` option and add it to the CI? Thanks in advance for your help :)

zacharyburnett: yeah I think a `-pytestdev` tox factor is a good idea, because then we could also theoretically mix and match it with `-devdeps` or `-stdevdeps`

braingram: Good point! I'll update this PR.

braingram:
Thanks @zacharyburnett I switched to the `-pytestdev` tox factor and added that to the CI. The log:
https://github.com/asdf-format/asdf/actions/runs/7616350863/job/20743185463?pr=1729#step:10:71
confirms it's running with `pytest-8.0.0rc2`.

zacharyburnett (approved):

PR #1739, opened 2024-01-23
fix __asdf_traverse__ for non-tagged object

### LINKED ISSUE #1738: Objects that implement `__asdf_traverse__` are expected to have a `_tag`
# Description of the problem

Adding `__asdf_traverse__` to an object will cause `info` (and possibly other commands) to fail if the object does not also contain a `_tag` attribute.

# Example of the problem

```python
import asdf

class MyContainer:
    def __init__(self, data):
        self.data = data

    def __asdf_traverse__(self):
        return self.data

c = MyContainer([1, 2, 3])
af = asdf.AsdfFile()
af['c'] = c
af.info()
```

The above example produces the following error:
```
  File "/Users/bgraham/projects/src/asdf/traverse_error.py", line 15, in <module>
    af.info()
  File "/Users/bgraham/projects/src/asdf/asdf/_asdf.py", line 1426, in info
    lines = display.render_tree(
  File "/Users/bgraham/projects/src/asdf/asdf/_display.py", line 45, in render_tree
    info = create_tree(
  File "/Users/bgraham/projects/src/asdf/asdf/_node_info.py", line 44, in create_tree
    schema_info = NodeSchemaInfo.from_root_node(
  File "/Users/bgraham/projects/src/asdf/asdf/_node_info.py", line 285, in from_root_node
    info.set_schema_from_node(node, extension_manager)
  File "/Users/bgraham/projects/src/asdf/asdf/_node_info.py", line 242, in set_schema_from_node
    tag_def = extension_manager.get_tag_definition(node._tag)
AttributeError: 'MyContainer' object has no attribute '_tag'
```

# System information

asdf version: main
python version: 3.10
operating system: mac osx

braingram:
# Description

This PR adds a check for a `_tag` attribute before attempting to fetch schema for a node during `NodeSchemaInfo` tree traversal.

The current code assumes that any object implementing `__asdf_traverse__` also has a `_tag` attribute which contains a valid tag.

"duck typing" was used here (instead of checking for `isinstance(Tagged)` as `roman_datamodels` defines non-Tagged objects that contain valid `_tag` attributes.

Fixes #1738

The weldx failure is unrelated (caused by a new pandas version) the roman_datamodels failure is addressed in: https://github.com/asdf-format/asdf/pull/1748

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/asdf/_node_info.py
+++ b/asdf/_node_info.py
@@ -287,7 +287,9 @@ class NodeSchemaInfo:
                                 info.set_schema_from_node(node, extension_manager)
                             except KeyError:
                                 # if _tag is not a valid tag, no schema will be found
-                                # and a KeyError will be raised
+                                # and a KeyError will be raised. We don't raise the KeyError
+                                # (or another exception) here as the node (object) might
+                                # be using _tag for a non-ASDF purpose.
                                 pass
 
                     else:
--- a/asdf/_node_info.py
+++ b/asdf/_node_info.py
@@ -282,7 +282,13 @@ class NodeSchemaInfo:
 
                     if cls.traversable(node):
                         t_node = node.__asdf_traverse__()
-                        info.set_schema_from_node(node, extension_manager)
+                        if hasattr(node, "_tag") and isinstance(node._tag, str):
+                            try:
+                                info.set_schema_from_node(node, extension_manager)
+                            except KeyError:
+                                # if _tag is not a valid tag, no schema will be found
+                                # and a KeyError will be raised
+                                pass
 
                     else:
                         t_node = node
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+3.3.0 (unreleased)
+------------------
+
+- Fix ``__asdf_traverse__`` for non-tagged objects [#1739]
+
 3.2.0 (2024-04-05)
 ------------------
 
--- a/asdf/_tests/_regtests/test_1738.py
+++ b/asdf/_tests/_regtests/test_1738.py
@@ -29,6 +29,8 @@ def test_info_on_non_tagged_asdf_traverse_object():
 
     # this should work even if _tag exists (and is not a tag)
     c._tag = {}
+    af.info()
     assert af.search(type_=int).paths == ["root['c'][0]", "root['c'][1]", "root['c'][2]"]
     c._tag = "a"
+    af.info()
     assert af.search(type_=int).paths == ["root['c'][0]", "root['c'][1]", "root['c'][2]"]
--- a/asdf/_tests/_regtests/test_1738.py
+++ b/asdf/_tests/_regtests/test_1738.py
@@ -1,0 +1,34 @@
+import asdf
+
+
+def test_info_on_non_tagged_asdf_traverse_object():
+    """
+    Calling info with a tree containing an object that implements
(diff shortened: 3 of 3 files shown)


braingram:
A few unrelated errors are seen in downstream:
- the [weldx errors](https://github.com/asdf-format/asdf/actions/runs/7629799922/job/20784056067?pr=1739#step:10:932) are due to a new pandas version
- the [specutils error] is due to a failure to fetch iers data
- the [stdatamodels error] is due to a crds update

asdf/_node_info.py
```diff
@@ -282,7 +282,13 @@ def from_root_node(cls, key, root_identifier, root_node, schema=None, refresh_ex
...
                     if cls.traversable(node):
                         t_node = node.__asdf_traverse__()
-                        info.set_schema_from_node(node, extension_manager)
+                        if hasattr(node, "_tag") and isinstance(node._tag, str):
+                            try:
+                                info.set_schema_from_node(node, extension_manager)
+                            except KeyError:
+                                # if _tag is not a valid tag, no schema will be found
+                                # and a KeyError will be raised
+                                pass
```
zacharyburnett, line 291: should this be raised instead, if `_tag` in invalid, to inform the user?
```suggestion
                                raise
```

braingram: Thanks for taking a look and asking the question.

`pass` here allows `info` to succeed if the tree contains an object with a `_tag` attribute that is not a valid ASDF tag (for example if the object uses `_tag` for non-ASDF reasons). With the following example:
```python
import asdf
class Foo:
    def __asdf_traverse__(self):
        return ["bar"]
af = asdf.AsdfFile()
af['foo'] = Foo()
af.info()
```
Running the example on main fails (see #1738)
```
AttributeError: 'Foo' object has no attribute '_tag'
```
I don't believe there should be an expectation that an object that implements `__asdf_traverse__` also have a valid asdf tag in `_tag` since the [docs](https://asdf.readthedocs.io/en/latest/asdf/extending/extensions.html#making-converted-object-s-contents-visible-to-info-and-search) don't mention the need for `_tag`. With this PR the above example succeeds producing an info output:
```
root (AsdfObject)
└─foo (Foo)
  └─0 (str): bar
```
However, if we modify the example and add a `_tag` attribute to `Foo` (let's assume `Foo` uses `_tag` for some non-ASDF use) the situation depends on the `pass` vs `raise`.
```python
import asdf
class Foo:
    def __asdf_traverse__(self):
        return ["bar"]
    _tag = "non_asdf_tag"
af = asdf.AsdfFile()
af['foo'] = Foo()
af.info()
```
With the `pass` use, the `info` output succeeds and is the same as above. However, switching to a `raise` causes the `info` to fail with the raised exception. That would mean that any object implementing `__asdf_traverse__` must also have a valid ASDF tag in `_tag` if it uses the `_tag` attribute. If we want to follow that expectation the docs should be updated to describe that requirement.

zacharyburnett: ok, that makes sense, so a `_tag` could be some other implementation, and that is valid for another use case. In that case, I also think `pass` is correct

zacharyburnett (approved):

braingram:
Thanks again! I made a minor addition to the comment before the `pass` to include some details from the discussion here. It will hopefully serve as a reminder for future me (or whomever) encounters the `pass` in the future.

PR #1744, opened 2024-01-30
Make asdf standard 1.6.0 stable (default)

braingram:
# Description

I milestoned this PR for asdf 4.0 because it's a rather major change.

Making 1.6.0 stable will require:
Phase 0 (at any time):
- [x] https://github.com/spacetelescope/rad/pull/370
- [x] https://github.com/asdf-format/asdf-unit-schemas/pull/34

Phase 1:
- [x] https://github.com/asdf-format/asdf/pull/1749

Phase 2:
- [x] https://github.com/asdf-format/asdf-standard/pull/421

Phase 3:
- [x] https://github.com/asdf-format/asdf-standard/pull/422
- [x] release asdf-standard with above PR
- [x] https://github.com/asdf-format/asdf-coordinates-schemas/pull/55
- [x] https://github.com/asdf-format/asdf-transform-schemas/pull/110
- [x] release asdf-transform-schemas asdf-coordinates-schemas with above PRs
- [x] update https://github.com/asdf-format/asdf-wcs-schemas/pull/59 to add pins for asdf-transform-schemas and asdf-coordinates-schemas
- [x] release asdf-wcs-schemas

Unfortunately at this point most of the schema repo CIs will be broken (as will asdf-astropy CI) due to the intertwined nature of these packages.

Phase 4:
- [x] update below asdf-astropy PR with new asdf-standard, asdf-coordinates-schemas, asdf-transform-schemas
- [x] https://github.com/astropy/asdf-astropy/pull/219
- [x] release asf-astropy
- [x] https://github.com/spacetelescope/stdatamodels/pull/258
- [x] https://github.com/sunpy/sunpy/pull/7432
- [x] https://github.com/DKISTDC/dkist/pull/324
- [ ] https://github.com/BAMWelDX/weldx/pull/910
- [x] update lower pin for asdf-wcs-schemas in gwcs: https://github.com/spacetelescope/gwcs/pull/491

Phase 5:
- [ ] this PR

Much of the above is due to updates to the ndarray schema:
- https://github.com/asdf-format/asdf-standard/pull/350
- https://github.com/asdf-format/asdf-standard/pull/411

All schemas that `$ref` ndarray (like [quantity](https://github.com/asdf-format/asdf-standard/blob/c6dd92d3006f566259caec40e3ac8dfd7ef42ced/resources/schemas/stsci.edu/asdf/unit/quantity-1.2.0.yaml#L63)) then need a version bump and so on...

This is further complicated by `quantity-1.1.0` currently existing in 2 released packages:
- [asdf-standard](https://github.com/asdf-format/asdf-standard/blob/main/resources/schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml)
- [asdf-unit-schemas](https://github.com/asdf-format/asdf-unit-schemas/blob/main/resources/stsci.edu/schemas/quantity-1.1.0.yaml)

(because of an incomplete effort to split `unit` `fits` `table` and `time` out of the core).
The approach taken here is to [decomission asdf-unit-schemas](https://github.com/asdf-format/asdf-unit-schemas/pull/34). asdf-standard will continue to provide updates to the `unit` (and other non-`core`) schemas. This seems sensible as these schemas are highly interdependent. More details can be found in: https://github.com/asdf-format/asdf-standard/pull/422

One question that occupied a lot of my thought was "should we change some of the `$ref`s to `tag`s?" On one hand this could make migrations like this easier (if every ndarray `$ref` was instead a wildcard `tag` (`ndarray-1.*`) most of these schemas would not need to be updated. However, this links the tag to the schema which has a few downsides:
- as tags depend on the extension version, we have effectively added a second version to the schema (a tag might only appear say in asdf-standard 1.6.0 which means that a schema that contains a `tag` link to that tag will only be valid for asdf-standard 1.6.0).
- `tag` links make "duck typing" with a schema very difficult if not impossible.

For example `unit/quantity` contains a `$ref` link to `unit/unit`. This means that `asdf-astropy` can use a differently tagged unit (`astropy/unint` for non-vo units) and still produce a valid `unit/quantity` (see [the wfi schema in rad as an example](https://github.com/spacetelescope/rad/blob/64cf407b88c1a369ca11d704f5af51347f7d9439/src/rad/resources/schemas/wfi_image-1.0.0.yaml#L29-L39)). If instead `unit/quantity` used a `tag` link to `unit/unit`, this same "duck typing" would not work with a differently tagged unit. `asdf-astropy` would have to instead:
- use a different schema for the `unit/unit` tag (this would break the tag-schema mapping defined in the standard `unit` manifest making any file produced with this approach likely to be incompatible with any other implementation or with other libraries that might choose to implement units)
- define a new `astropy/quantity` for non-vo units (and so on up the tree of schema references...)
- add the `astropy/unit` tag to the `quantity` schema (further linking tags and schemas and requiring addition standard schema updates for "downstream" libraries).

At the moment I am of the mind that keeping the schemas as separate from the tags as possible is the better option (so `$ref` instead of `tag`). This allows the schemas to function even if they are treated as normal "jsonschema"s. Additionally the `tag` validator behavior seems loosely defined in the standard where [it states](https://github.com/asdf-format/asdf-standard/blob/c6dd92d3006f566259caec40e3ac8dfd7ef42ced/resources/schemas/stsci.edu/yaml-schema/draft-01.yaml#L20-L21) "Implementation of this validator is optional and depends on details of the YAML parser." For similar reasons the above `asdf-transform-schemas` PR  did not rely on the feature in asdf to use multiple schemas in a tag definition (to allow the many transforms that `$ref` the transform schema to instead include them in the manifest).

Both `stdatamodels` (datamodels) and `rad` use metaschemas based off of `asdf-schema-1.0.0` (which is updated to `1.1.0` in https://github.com/asdf-format/asdf-standard/pull/422). As neither of these packages version schemas updating the metaschema version will force these packages to use exclusively the `1.6.0` standard (which will almost certainly cause issues if old versions of asdf/asdf-standard are used). Instead, I suggest we not update the metaschemas (and keep them using the old `asdf-schema-1.0.0` metaschema). The only downside is the lack of `float16` support in the `datatype` keyword validator. This seems like an acceptable limitation for the time being. Once asdf standard 1.6.0 is stable and the asdf version that sets it as the default agrees with the minimum required version for each of those packages the metaschemas can be updated.

Closes: https://github.com/asdf-format/asdf/issues/1866

# Checklist:

- [x] tests ran successfully

### DIFF
--- a/asdf/_tests/core/_converters/test_tree.py
+++ b/asdf/_tests/core/_converters/test_tree.py
@@ -77,6 +77,7 @@ def test_history_entry(tmp_path):
 def test_subclass_metadata():
     subclass_metadata = SubclassMetadata(name="SomeCoolSubclass")
 
-    result = helpers.roundtrip_object(subclass_metadata)
+    # support for this was dropped in 1.6.0
+    result = helpers.roundtrip_object(subclass_metadata, version="1.5.0")
 
     assert result == subclass_metadata
--- a/asdf/_tests/tags/core/tests/test_ndarray.py
+++ b/asdf/_tests/tags/core/tests/test_ndarray.py
@@ -75,7 +75,7 @@ def with_custom_extension():
             {
                 "http://nowhere.org/schemas/custom/datatype-1.0.0": """%YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.0.0"
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://nowhere.org/schemas/custom/datatype-1.0.0"
 type: object
 properties:
@@ -102,7 +102,7 @@ properties:
     exact_datatype: true""",
                 "http://nowhere.org/schemas/custom/ndim-1.0.0": """%YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.0.0"
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://nowhere.org/schemas/custom/ndim-1.0.0"
 type: object
 properties:
--- a/asdf/_tests/tags/core/tests/test_ndarray.py
+++ b/asdf/_tests/tags/core/tests/test_ndarray.py
@@ -282,7 +282,7 @@ def test_array_inline_threshold_recursive(tmp_path):
 
 def test_copy_inline():
     yaml = """
-x0: !core/ndarray-1.0.0
+x0: !core/ndarray-1.1.0
   data: [-1.0, 1.0]
     """
 
@@ -371,7 +371,7 @@ def test_inline():
 
 
 def test_inline_bare():
-    content = "arr: !core/ndarray-1.0.0 [[1, 2, 3, 4], [5, 6, 7, 8]]"
+    content = "arr: !core/ndarray-1.1.0 [[1, 2, 3, 4], [5, 6, 7, 8]]"
     buff = helpers.yaml_to_asdf(content)
 
     with asdf.open(buff) as ff:
@@ -413,7 +413,7 @@ def test_len_roundtrip(tmp_path):
 
 def test_mask_arbitrary():
     content = """
-arr: !core/ndarray-1.0.0
+arr: !core/ndarray-1.1.0
   data: [[1, 2, 3, 1234], [5, 6, 7, 8]]
   mask: 1234
     """
@@ -425,7 +425,7 @@ arr: !core/ndarray-1.0.0
 
 def test_mask_nan():
     content = """
-arr: !core/ndarray-1.0.0
+arr: !core/ndarray-1.1.0
   data: [[1, 2, 3, .NaN], [5, 6, 7, 8]]
   mask: .NaN
     """
@@ -455,7 +455,7 @@ def test_string_table(tmp_path):
 
 
 def test_inline_string():
-    content = "arr: !core/ndarray-1.0.0 ['a', 'b', 'c']"
+    content = "arr: !core/ndarray-1.1.0 ['a', 'b', 'c']"
     buff = helpers.yaml_to_asdf(content)
 
     with asdf.open(buff) as ff:
@@ -464,7 +464,7 @@ def test_inline_string():
 
 def test_inline_structured():
     content = """
-arr: !core/ndarray-1.0.0
+arr: !core/ndarray-1.1.0
     datatype: [['ascii', 4], uint16, uint16, ['ascii', 4]]
     data: [[M110, 110, 205, And],
            [ M31,  31, 224, And],
@@ -622,10 +622,10 @@ def test_operations_on_ndarray_proxies(tmp_path):
 
 def test_mask_datatype(tmp_path):
     content = """
-arr: !core/ndarray-1.0.0
+arr: !core/ndarray-1.1.0
     data: [1, 2, 3]
     dtype: int32
-    mask: !core/ndarray-1.0.0
+    mask: !core/ndarray-1.1.0
         data: [true, true, false]
     """
     buff = helpers.yaml_to_asdf(content)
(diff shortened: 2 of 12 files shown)


braingram:
Reserving a comment for regression tests.

Roman:
https://github.com/spacetelescope/RegressionTests/actions/runs/10221961819

braingram:
The weldx downstream failure is real. I rebased the open PR and added a comment that the development branch will soon be moving to asdf 4.0. We will need to keep in mind that the asdf downstream test will fail after this PR is merged.

braingram:
https://github.com/asdf-format/asdf/issues/1866 to track adding this change to "what's new"

perrygreenfield:
> Both `stdatamodels` (datamodels) and `rad` use metaschemas based off of `asdf-schema-1.0.0` (which is updated to `1.1.0` in [asdf-format/asdf-standard#422](https://github.com/asdf-format/asdf-standard/pull/422)). As neither of these packages version schemas updating the metaschema version will force these packages to use exclusively the `1.6.0` standard (which will almost certainly cause issues if old versions of asdf/asdf-standard are used). Instead, I suggest we not update the metaschemas (and keep them using the old `asdf-schema-1.0.0` metaschema). The only downside is the lack of `float16` support in the `datatype` keyword validator. This seems like an acceptable limitation for the time being. Once asdf standard 1.6.0 is stable and the asdf version that sets it as the default agrees with the minimum required version for each of those packages the metaschemas can be updated.

Does this mean that use of `float16` in datatype won't pass validation until  1.6.0 becomes default? I want to make sure I understand the consequence (not that I disagree with it).

braingram:
> > Both `stdatamodels` (datamodels) and `rad` use metaschemas based off of `asdf-schema-1.0.0` (which is updated to `1.1.0` in [asdf-format/asdf-standard#422](https://github.com/asdf-format/asdf-standard/pull/422)). As neither of these packages version schemas updating the metaschema version will force these packages to use exclusively the `1.6.0` standard (which will almost certainly cause issues if old versions of asdf/asdf-standard are used). Instead, I suggest we not update the metaschemas (and keep them using the old `asdf-schema-1.0.0` metaschema). The only downside is the lack of `float16` support in the `datatype` keyword validator. This seems like an acceptable limitation for the time being. Once asdf standard 1.6.0 is stable and the asdf version that sets it as the default agrees with the minimum required version for each of those packages the metaschemas can be updated.
> 
> Does this mean that use of `float16` in datatype won't pass validation until 1.6.0 becomes default? I want to make sure I understand the consequence (not that I disagree with it).

`float16` support is available in 1.6.0. We could start using it in (for example) romancal once 1.6.0 is "stable" (which I'm considering the same as making it the default). The metaschema issues won't prevent using `float16` arrays in (for example) roman datamodels (schemas aren't checked against metaschemas as part of validation).

However the metaschema issue will mean that adding a datatype validator to the `float16` array would cause the call to `check_schema` that `pytest-asdf` runs to fail.

To provide an example. Let's say we want to change `ImageModel.data` to use `float16`. We could change this line:
https://github.com/spacetelescope/rad/blob/f2d7af769ec0fd77283cdcbe020cca96f026a5a4/src/rad/resources/schemas/wfi_image-1.0.0.yaml#L45
to:
```
datatype: float16
```
and save files without issue (no validation errors as long as we assign a `float16` array to `data`). However the following would fail (and is equivalent to a check that `pytest-asdf` runs on schema files):
```
>> asdf.schema.check_schema(asdf.schema.load_schema("asdf://stsci.edu/datamodels/roman/schemas/wfi_image-1.0.0"))
ValidationError: 'float16' is not valid under any of the given schemas
```
due to the `wfi_image-1.0.0` using the `asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0` metaschema which refs `http://stsci.edu/schemas/asdf/asdf-schema-1.0.0` which refs `http://stsci.edu/schemas/asdf/core/ndarray-1.0.0#/definitions/datatype` which doesn't define `float16` as a valid datatype.

Note that this only prevents adding something like:
```
datatype: float16
```
to the schemas that extend `asdf-schema-1.0.0`. Files can contain `float16` arrays (as long as they use ASDF 1.6.0).

perrygreenfield:
So this only affects validation of schemas against the metaschema I take it. OK, that isn't so serious.

perrygreenfield (approved):
LGTM

PR #1745, opened 2024-01-30
Deprecate `AsdfFile.version_map`

braingram:
# Description

This PR:
- deprecates `AsdfFile.version_map`
- removes asdf uses of `versioning.get_version_map` (which this PR also renames to `_get_version_map` as it is undocumented, not listed in `versioning.__all__` and unused by all known downstream packages)
- no longer relies on the [version maps in asdf-standard](https://github.com/asdf-format/asdf-standard/blob/main/resources/schemas/stsci.edu/asdf/version_map-1.6.0.yaml) to map asdf standard versions to file format and yaml versions

Note that at the moment ASDF only supports file format 1.0.0 and yaml 1.1. This PR adds those constants to `asdf.versioning`.

When this PR is merged a new issue can be opened to track the removal of `AsdfFile.version_map` in asdf 4.0. The PR that removes `version_map` can also remove `versioning._get_version_map`. Once that PR is merged, the version map files can be removed from asdf standard.

sunpy downstream is failing due to pytest 8
weldx is failing due to new pandas deprecations

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/asdf/_asdf.py
+++ b/asdf/_asdf.py
@@ -258,7 +258,7 @@ class AsdfFile:
 
     @property
     def version_map(self):
-        return versioning.get_version_map(self.version_string)
+        return versioning._get_version_map(self.version_string)
 
     @property
     def extensions(self):
--- a/asdf/_asdf.py
+++ b/asdf/_asdf.py
@@ -750,6 +750,10 @@ class AsdfFile:
             msg = f"Unparsable version in ASDF file: {parts[1]}"
             raise ValueError(msg) from err
 
+        if version != versioning._FILE_FORMAT_VERSION:
+            msg = f"Unsupported ASDF file format version {version}"
+            raise ValueError(msg)
+
         return version
 
     @classmethod
@@ -825,13 +829,15 @@ class AsdfFile:
                 msg = "Does not appear to be a ASDF file."
                 raise ValueError(msg) from e
             self._file_format_version = cls._parse_header_line(header_line)
-            self.version = self.file_format_version
 
             self._comments = cls._read_comment_section(fd)
 
             version = cls._find_asdf_version_in_comments(self._comments)
             if version is not None:
                 self.version = version
+            else:
+                # If no ASDF_STANDARD comment is found...
+                self.version = versioning.AsdfVersion("1.0.0")
 
             # Now that version is set for good, we can add any additional
             # extensions, which may have narrow ASDF Standard version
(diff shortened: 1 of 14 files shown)


asdf/_asdf.py
```diff
@@ -807,7 +809,7 @@ def _open_asdf(
                 msg = "Does not appear to be a ASDF file."
                 raise ValueError(msg) from e
             self._file_format_version = cls._parse_header_line(header_line)
-            self.version = self._file_format_version
+            self.version = self.file_format_version
```
braingram, line 812: This seems incorrect. `self.version` is the asdf standard version yet here it is being set to the `file_format_version` (which is always `1.0.0`). `self.version` is overwritten for probably every file a few lines down (from the undocumented `ASDF_STANDARD` comment). As the ASDF standard does not define the comment or what to do in this case I'm inclined to leave in this behavior for this PR (as it's only tangentially related to deprecating `version_map`).

It is likely an improvement to change this to default to the `default_version` if the file doesn't define a version. However this would be a major change (switching from asdf standard 1.0.0 to 1.5.0) if there is any code that relies on this.

asdf/_tests/_helpers.py
```diff
@@ -408,62 +402,3 @@ def assert_no_warnings(warning_class=None):
         assert not any(isinstance(w.message, warning_class) for w in recorded_warnings), display_warnings(
             recorded_warnings,
         )
-
-
-def _assert_extension_type_correctness(extension, extension_type, resolver):
```
braingram, line 413: This was unused so it was removed instead of updating it to not use `get_version_map`.

asdf/_asdf.py
braingram: Good guess!

I changed this to just default to `1.0.0` if no `#ASDF_STANDARD` comment is found. That way it should hopefully not break for `2.0.0`.

44fe69f9ee1abb3ebf2dc58133e45484a3d1c4ef

There were 2 tests that failed following this change. One was filling in the standard version for `#ASDF` (so should have been producing an error but was not), the other was expecting an unsupported standard version warning when the `#ASDF` version was incorrect. Both of these were updated. I wouldn't be surprised if there are more places where these two things are conflated :-/

asdf/_tests/_helpers.py
```diff
@@ -286,57 +275,6 @@ def _assert_roundtrip_tree(
             asdf_check_func(ff)
 
 
-def yaml_to_asdf(yaml_content, yaml_headers=True, standard_version=None):
```
braingram, line 289: This was replaced with `asdf.testing.helpers.yaml_to_asdf` instead of updating it to remove the use of `version_map`.

PR #1759, opened 2024-02-20
Remove the use of tmpdir and related

WilliamJamieson:
# Description

[`pytest` recommends creating temporary directories with the `tmp_path` fixture instead of the legacy `tmpdir`](https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures). To check for use of the legacy method one can run with
```
pytest -p no:legacypath
```
This PR adds this to he `addopts` for `pytest` and then fixes all the issues. In particular the `pytest_asdf` pluggin blocks all downstream packages from running their tests with this option to detect the legacy temporary directory creations.

# Checklist:

- [ ] tests ran successfully


braingram:
Thanks for making these improvements.

The `compatibility` tests are failing and will need to be updated if we want to use the `no:legacypath` option. Would you apply similar fixes to the `tmpdir` usage there?

braingram (approved):
Thanks for the update! Looks good to me.

PR #1765, opened 2024-03-04
drop doctestplus test dependency

braingram:
# Description

pytest-doctestplus is listed as a test dependency. However, none of the "plus" features are used and this PR drops it as a dependency and replaces it with the builtin doctest.

With this PR the 3.11 tests pick up 1756 tests
https://github.com/asdf-format/asdf/actions/runs/8143804626/job/22256481482?pr=1765#step:10:247

this is the same as the 3.11 tests for main:
https://github.com/asdf-format/asdf/actions/runs/8096624540/job/22125799613#step:10:248

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/.github/workflows/s390x.yml
+++ b/.github/workflows/s390x.yml
@@ -69,7 +69,7 @@ jobs:
           run: |
             python3 -m venv --system-site-packages tests
             source tests/bin/activate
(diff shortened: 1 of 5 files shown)


zacharyburnett (approved):
nice catch!

PR #1769, opened 2024-03-10
fix masked array class path for numpy 2.0

braingram:
Numpy 2.0 moved the internal path for `np.ma.MaskedArray`. See: https://github.com/numpy/numpy/commit/7b9dede257171bdfafdd4a6fa43ac4df806b1103

This causes masked arrays to be saved as standard numpy arrays with a warning. See:
https://github.com/astropy/asdf-astropy/actions/runs/8222457553/job/22484029609?pr=219#step:10:291

This PR updates the converter to be aware of the new path.

I believe https://github.com/asdf-format/asdf/pull/1654 would have prevented this failure (by allowing for the public class path).

# Description

# Checklist:

- [ ] pre-commit checks ran successfully
- [ ] tests ran successfully
- [ ] for a public change, a changelog entry was added
- [ ] for a public change, documentation was updated
- [ ] for any new features, unit tests were added

PR #1770, opened 2024-03-10
Fix docs and modify CI job run requirements

braingram:
# Description

Fix a few docs links (due to changes in the asdf-standard docs) and modify the CI jobs to not prevent jobs from running if the `asdf-schemas` job fails (as it does now).

# Checklist:

- [ ] tests ran successfully


braingram:
Docs built, CI now is back to testing numpy 2.0 to see issue addressed in: https://github.com/asdf-format/asdf/pull/1769

PR #1771, opened 2024-03-10
restrict dev ASDF standard tests for older version in compatibility tests

braingram:
For asdf version 2.13 and earlier only ndarray version 1.0.0 was supported:
https://github.com/asdf-format/asdf/blob/8d535c61444733d57e2680eceea85612ca76c2ea/asdf/tags/core/ndarray.py#L213-L216

Yet these versions claimed support for 1.6.0
https://github.com/asdf-format/asdf/blob/efafac8e6efec81370bb6fbdf06c82e958185d7c/asdf/versioning.py#L156

This causes issues for the compatibility tests as the old version will write out a file claiming to be 1.6.0 but using the ndarray 1.0.0 tag. Resulting in an error:
```
FAILED compatibility_tests/test_file_compatibility.py::test_file_compatibility[asdf==2.13.0] - asdf.exceptions.AsdfConversionWarning: tag:stsci.edu:asdf/core/ndarray-1.0....
```
See: https://github.com/asdf-format/asdf/actions/runs/8222842083/job/22484881421#step:10:770
resulting in

This PR updates the compatibility tests to remove the comparison of 1.6.0 (the "development" version of the standard) for asdf versions older than 2.14.0 (to exclude 2.13).

# Checklist:

- [ ] tests ran successfully


braingram:
devdeps tests addressed in: https://github.com/asdf-format/asdf/pull/1769

zacharyburnett (approved):
I don't understand this, but if the dev tests pass then I assume it works

braingram:
I converted this back to draft as a reminder to look into the failures a bit more. Especially since dkist is still using 2.11.

Issue #1773, opened 2024-03-18
deprecate `AsdfSpec` and `format_tag`


braingram:
`AsdfSpec` (based off of `SimpleSpec` from semantic_version) is effectively unused.

It is referenced once in `format_tag`:
https://github.com/asdf-format/asdf/blob/aedf1563eeeb64eec61c860073e20e88ddb49d5c/asdf/testing/helpers.py#L20-L21
`format_tag` has only a single use in `asdf`:
https://github.com/asdf-format/asdf/blob/aedf1563eeeb64eec61c860073e20e88ddb49d5c/asdf/_tests/test_schema.py#L841
where a spec is not provided. 

The `format_tag` use can be replaced by an f-string.

One motivation for deprecating `AsdfSpec` is that it provides unexpected version comparisons: see #1772

PR #1774, opened 2024-03-18
Deprecate `asdf.testing.helpers.format_tag` and `asdf.versioning.AsdfSpec`

braingram:
# Description

This PR deprecates `asdf.testing.helpers.format_tag` (and replaces it's one use with a string) and `asdf.versioning.AsdfSpec` (which is unused).

Fixes: #1773

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@
 
 - Fix ``__asdf_traverse__`` for non-tagged objects [#1739]
 
+- Deprecate ``asdf.testing.helpers.format_tag`` [#1774]
+
+- Deprecate ``asdf.versioning.AsdfSpec`` [#1774]
+
+
 3.2.0 (2024-04-05)
 ------------------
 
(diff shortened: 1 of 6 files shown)


braingram:
The downstream dkist error is unrelated and looks to be caused by a `gwcs` version comparison in one of their tests.

zacharyburnett (approved):
LGTM; is this in the docs anywhere? I see an auto-generated reference to `format_tag` but not explicitly creating a tag from a string

braingram:
Thanks!

The only docs I also see are the auto-generated ones:
https://asdf.readthedocs.io/en/latest/search.html?q=format_tag&check_keywords=yes&area=default

I don't quite know the whole history of `format_tag` but looking through various `git blames`:
- was moved from `asdf.types` (when that was deprecated)
- was moved from `asdf.asdftypes` (when that was deprecated)
- was added in the first "skeleton" of the project

but never appears to be documented beyond the docstring and auto-generated docs. It seems to always have been a way to construct a uri string from several string (or `AsdfVersion`, when that was added) components.

Issue #1711, opened 2023-12-18
Remove `copy_arrays` and switch `memmap` default to `False`


braingram:
See https://github.com/asdf-format/asdf/pull/1667

For the 4.0.0 release remove `copy_arrays` and switch the default value of `memmap` to `False`.

PR #1797, opened 2024-07-12
deprecate `copy_arrays` with warning

### LINKED ISSUE #1710: Deprecate `copy_arrays`
As described in https://github.com/asdf-format/asdf/pull/1667

After asdf 3.1 is released deprecate `copy_arrays` (mentioning new `memmap` option).

zacharyburnett:
resolves #1710 #1711 

# Checklist:

- [x] tests ran successfully
- [x] for a public change, documentation was updated
- [x] for any new features, unit tests were added

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -185,6 +185,13 @@ The `open` function also works as a context handler:
     with asdf.open("example.asdf") as af:
         ...
 
+.. warning::
+    The ``copy_arrays`` argument of `asdf.open()` and `AsdfFile` is deprecated,
+    and will be removed in ASDF 4.0. It is replaced by the more-explicitly-named
+    ``memmap`` argument as the inverse of ``copy_arrays``
+    (``memmap == not copy_arrays``). Using ``copy_arrays`` will work for now, but
+    give an ``AsdfWarning``.
+
 To get a quick overview of the data stored in the file, use the top-level
 `AsdfFile.info()` method:
 
--- a/README.rst
+++ b/README.rst
@@ -250,7 +250,7 @@ the ``memmap`` option of `open` when reading:
 
 .. code:: python
 
-    af = asdf.open("example.asdf", memmap=False)
+    af = asdf.open("example.asdf", memmap=True)
 
 .. _end-read-file-text:
 
--- a/README.rst
+++ b/README.rst
@@ -244,13 +244,13 @@ Array data remains unloaded until it is explicitly accessed:
     >>> np.equal(af["powers"]["squares"], expected).all()
     True
 
-By default, uncompressed data blocks are not memory mapped.
-Memory mapping can be enabled for efficient access by using
-the ``memmap`` option of `open` when reading:
+By default, uncompressed data blocks are not memory mapped for efficient
+access. Memory mapping can be disabled by using the ``memmap``
+option of `open` when reading:
 
 .. code:: python
 
-    af = asdf.open("example.asdf", memmap=True)
+    af = asdf.open("example.asdf", memmap=False)
 
 .. _end-read-file-text:
 
--- a/README.rst
+++ b/README.rst
@@ -189,7 +189,7 @@ The `open` function also works as a context handler:
     The ``copy_arrays`` argument of `asdf.open()` and `AsdfFile` is deprecated,
     and will be removed in ASDF 4.0. It is replaced by ``memmap``, which
     is the opposite of ``copy_arrays`` (``memmap == not copy_arrays``).
-    In ASDF 4.0, ``memmap`` will default to ``False``, which means arrays 
+    In ASDF 4.0, ``memmap`` will default to ``False``, which means arrays
     will no longer be memory-mapped by default.
 
 To get a quick overview of the data stored in the file, use the top-level
--- a/README.rst
+++ b/README.rst
@@ -244,13 +244,13 @@ Array data remains unloaded until it is explicitly accessed:
     >>> np.equal(af["powers"]["squares"], expected).all()
     True
 
-By default, uncompressed data blocks are memory mapped for efficient
-access. Memory mapping can be disabled by using the ``copy_arrays``
-option of `open` when reading:
+By default, uncompressed data blocks are not memory mapped.
+Memory mapping can be enabled for efficient access by using
+the ``memmap`` option of `open` when reading:
 
 .. code:: python
 
-    af = asdf.open("example.asdf", copy_arrays=True)
+    af = asdf.open("example.asdf", memmap=False)
 
 .. _end-read-file-text:
 
--- a/README.rst
+++ b/README.rst
@@ -187,10 +187,10 @@ The `open` function also works as a context handler:
 
 .. warning::
     The ``copy_arrays`` argument of `asdf.open()` and `AsdfFile` is deprecated,
-    and will be removed in ASDF 4.0. It is replaced by the more-explicitly-named
-    ``memmap`` argument as the inverse of ``copy_arrays``
-    (``memmap == not copy_arrays``). Using ``copy_arrays`` will work for now, but
-    give an ``AsdfWarning``.
+    and will be removed in ASDF 4.0. It is replaced by ``memmap``, which
+    is the opposite of ``copy_arrays`` (``memmap == not copy_arrays``).
+    In ASDF 4.0, ``memmap`` will default to ``False``, which means arrays 
+    will no longer be memory-mapped by default.
 
 To get a quick overview of the data stored in the file, use the top-level
 `AsdfFile.info()` method:
--- a/README.rst
+++ b/README.rst
@@ -244,7 +244,7 @@ Array data remains unloaded until it is explicitly accessed:
     >>> np.equal(af["powers"]["squares"], expected).all()
     True
 
-By default, uncompressed data blocks are not memory mapped for efficient
+By default, uncompressed data blocks are memory mapped for efficient
 access. Memory mapping can be disabled by using the ``memmap``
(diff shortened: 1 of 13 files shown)


README.rst
```diff
@@ -248,13 +248,13 @@ Array data remains unloaded until it is explicitly accessed:
     >>> np.equal(af["powers"]["squares"], expected).all()
     True
 
-By default, uncompressed data blocks are memory mapped for efficient
-access. Memory mapping can be disabled by using the ``copy_arrays``
-option of `open` when reading:
+By default, uncompressed data blocks are not memory mapped.
+Memory mapping can be enabled for efficient access by using
+the ``memmap`` option of `open` when reading:
```
braingram, line 253:
```suggestion
By default, uncompressed data blocks are not memory mapped for efficient
access. Memory mapping can be disabled by using the ``memmap``
option of `open` when reading:
```

README.rst
```diff
@@ -248,13 +248,13 @@ Array data remains unloaded until it is explicitly accessed:
...
 
 .. code:: python
 
-    af = asdf.open("example.asdf", copy_arrays=True)
+    af = asdf.open("example.asdf", memmap=True)
```
braingram, line 257:
```suggestion
    af = asdf.open("example.asdf", memmap=False)
```

asdf/_asdf.py
```diff
@@ -75,8 +75,8 @@ def __init__(
         ignore_version_mismatch=True,
         ignore_unrecognized_tag=False,
         ignore_implicit_conversion=NotSet,
-        copy_arrays=False,
         memmap=NotSet,
+        copy_arrays=NotSet,
```
braingram, line 79: Since we're not using a `*` here (to make these keyword only arguments) the order of the kwargs matters on the (very) off chance someone is calling this will all positioned arguments. Would you move `copy_arrays` back before `memmap`?

asdf/_asdf.py
```diff
@@ -182,9 +179,13 @@ def __init__(
         self._closed = False
         self._external_asdf_by_uri = {}
         # if memmap is set, it overrides copy_arrays
-        if memmap is not NotSet:
-            copy_arrays = not memmap
-        self._blocks = BlockManager(uri=uri, lazy_load=lazy_load, memmap=not copy_arrays)
+        if copy_arrays is not NotSet:
+            warnings.warn("copy_arrays is deprecated; use memmap instead", AsdfDeprecationWarning)
```
braingram, line 183: Let's use a non-`DeprecationWarning` here (maybe just `AsdfWarning`). Python by default will hide deprecation warnings not originating in `__main__` and we want all users to see this (even those using jupyter etc).

Also let's expand the message to include something about the upcoming `memmap=False` default change. This is pretty much the only opportunity we have to communicate this change to users. Maybe something like:
```
copy_arrays is deprecated; use memmap instead. Note that memmap will default to False in asdf 4.0.
```

asdf/_tests/test_deprecated.py
```diff
@@ -146,3 +146,15 @@ def test_AsdfFile_ignore_implicit_conversion_deprecation(value):
...
     with pytest.warns(AsdfDeprecationWarning, match="ignore_implicit_conversion is deprecated"):
         asdf.treeutil.walk_and_modify({}, lambda obj: obj, ignore_implicit_conversion=value)
+
+
+@pytest.mark.parametrize("memmap", [True, False, asdf._asdf.NotSet])
+def test_copy_arrays_deprecation(memmap, tmp_path):
+    fn = tmp_path / "test.asdf"
+    af = asdf.AsdfFile()
+    af["a"] = 1
+    af["b"] = {"$ref": "#"}
```
braingram, line 156: Is "b" needed here?

asdf/_tests/test_deprecated.py
```diff
@@ -146,3 +146,15 @@ def test_AsdfFile_ignore_implicit_conversion_deprecation(value):
 def test_walk_and_modify_ignore_implicit_conversion_deprecation(value):
     with pytest.warns(AsdfDeprecationWarning, match="ignore_implicit_conversion is deprecated"):
         asdf.treeutil.walk_and_modify({}, lambda obj: obj, ignore_implicit_conversion=value)
+
+
+@pytest.mark.parametrize("memmap", [True, False, asdf._asdf.NotSet])
```
braingram, line 151: I think parametrizing "copy_arrays" would be more useful here, making sure the warning is issued for both `True` and `False` and not issued for the default (not provided).

braingram (changes requested):
I left a few suggestions. One major one being the warning message. Also, if you can think of other ways we can inform users of the upcoming memmap default change that would be great! It's a rather major change and this switch from "copy_arrays" to "memmap" was somewhat intended to provide a warning to the users.

braingram:
I also added the `Downstream CI` label. Some of those packages don't turn warnings into errors (like jwst). Would you check which downstream packages are going to fail (or produce new warnings) with these changes?

zacharyburnett:
> I also added the `Downstream CI` label. Some of those packages don't turn warnings into errors (like jwst). Would you check which downstream packages are going to fail (or produce new warnings) with these changes?

Sure!

zacharyburnett:
> Also, if you can think of other ways we can inform users of the upcoming memmap default change that would be great! It's a rather major change and this switch from "copy_arrays" to "memmap" was somewhat intended to provide a warning to the users.

I tried adding this warning box but GitHub doesn't render the RST correctly:
image

asdf/_asdf.py
```diff
@@ -117,16 +117,15 @@ def __init__(
...
 
         memmap : bool, optional
             When `True`, when reading files, attempt to memmap underlying data
             arrays when possible. When set, this argument will override
-            ``copy_arrays``.  When not set, the ``copy_arrays`` will determine
-            if arrays are memory mapped or copied. ``copy_arrays`` will be
-            deprecated and the default will change in an upcoming asdf version
-            which by default will not memory map arrays.
+            ``copy_arrays``. The default will change to ``False`` in an upcoming
+            ASDF version.
```
braingram, line 128:
```suggestion
            ASDF version. At the moment the default is ``True``.
```
I think since both copy_arrays and memmap are `NotSet` in the arguments some description of the default state here would be helpful. This is only temporary since in asdf 4.0 we can remove the `NotSet` and switch it to the default.

asdf/_asdf.py
```diff
@@ -1661,16 +1665,15 @@ def open_asdf(
...
 
     memmap : bool, optional
         When `True`, when reading files, attempt to memmap underlying data
         arrays when possible. When set, this argument will override
-        ``copy_arrays``.  When not set, the ``copy_arrays`` will determine
-        if arrays are memory mapped or copied. ``copy_arrays`` will be
-        deprecated and the default will change in an upcoming asdf version
-        which by default will not memory map arrays.
+        ``copy_arrays``. The default will change to ``False`` in an upcoming
+        ASDF version.
```
braingram, line 1676:
```suggestion
        ASDF version. At the moment the default is ``True``.
```
Similar to the other comment.

braingram (changes requested):
Code changes look good to me!

I added a few more suggestions to docs.

The remaining issue is the downstream usage. Of the failed packages most we can update (and perhaps bump asdf to at least 3.1.0 (to use memmap). dkist follows [SPEC-0000](https://scientific-python.org/specs/spec-0000/) and considers asdf a "core" package so we can open a PR which accommodates the older versions (only providing memmap for 3.1.0 and after).

Are there any warnings about this change in the non-failing jobs?

README.rst
```diff
@@ -189,6 +189,13 @@ The `open` function also works as a context handler:
     with asdf.open("example.asdf") as af:
         ...
 
+.. warning::
+    The ``copy_arrays`` argument of `asdf.open()` and `AsdfFile` is deprecated,
+    and will be removed in ASDF 4.0. It is replaced by the more-explicitly-named
+    ``memmap`` argument as the inverse of ``copy_arrays``
+    (``memmap == not copy_arrays``). Using ``copy_arrays`` will work for now, but
+    give an ``AsdfWarning``.
```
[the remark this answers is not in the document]
zacharyburnett:
```suggestion
    and will be removed in ASDF 4.0. It is replaced by ``memmap``, which
    is the opposite of ``copy_arrays`` (``memmap == not copy_arrays``).
    In ASDF 4.0, ``memmap`` will default to ``False``, which means arrays 
    will no longer be memory-mapped by default.
```

zacharyburnett:
a few warnings in `asdf-astropy` and `sunpy`; I can work on PRs for `stdatamodels` and `roman_datamodels`

braingram:
Thanks! 

asdf-astropy is maintained by "us" so that could also use a PR (and gwcs). I also try to make PRs for our other downstream users for stuff like this and am happy to help with PRs. Would you like to divide these up? I think so far we have:
- asdf-zarr: https://github.com/search?q=repo%3Aasdf-format%2Fasdf-zarr%20copy_arrays&type=code
- stdatamodels
- roman_datamodels
- gwcs https://github.com/search?q=repo%3Aspacetelescope%2Fgwcs%20copy_arrays&type=code
- asdf-astropy
- one use in jwst: https://github.com/search?q=repo%3Aspacetelescope%2Fjwst%20copy_arrays&type=code
- a few in romancal: https://github.com/search?q=repo%3Aspacetelescope%2Fromancal+copy_arrays&type=code
- weldx
- sunpy
- dkist

What a kettle of fish! :)

braingram:
Also one of my suggestions is causing a pre-commit failure.

PR #1800, opened 2024-07-15
remove `copy_arrays` parameter

zacharyburnett:
# Description

deprecate `copy_arrays`

# Checklist:

- [x] pre-commit checks ran successfully
- [ ] tests ran successfully
- [ ] for a public change, a changelog entry was added
- [ ] for a public change, documentation was updated
- [ ] for any new features, unit tests were added

PR #1801, opened 2024-07-15
default `memmap` parameter to `False`

zacharyburnett:
based on (and blocked by) https://github.com/asdf-format/asdf/pull/1800

# Description

default `memmap=False`

# Checklist:

- [ ] tests ran successfully

### DIFF
--- a/asdf/_asdf.py
+++ b/asdf/_asdf.py
@@ -57,7 +57,6 @@ class AsdfFile:
         extensions=None,
         version=None,
         ignore_unrecognized_tag=False,
-        ignore_implicit_conversion=NotSet,
         memmap=False,
         lazy_load=True,
         custom_schema=None,
--- a/asdf/_asdf.py
+++ b/asdf/_asdf.py
@@ -57,7 +57,8 @@ class AsdfFile:
         extensions=None,
         version=None,
         ignore_unrecognized_tag=False,
-        memmap=True,
+        ignore_implicit_conversion=NotSet,
+        memmap=False,
         lazy_load=True,
         custom_schema=None,
     ):
@@ -88,7 +89,7 @@ class AsdfFile:
 
         memmap : bool, optional
             When `True`, when reading files, attempt to memmap underlying data
-            arrays when possible. Defaults to ``True``.
+            arrays when possible. Defaults to ``False``.
 
         lazy_load : bool, optional
             When `True` and the underlying file handle is seekable, data
@@ -1510,7 +1511,7 @@ def open_asdf(
     extensions=None,
     ignore_unrecognized_tag=False,
     _force_raw_types=False,
-    memmap=True,
+    memmap=False,
     lazy_tree=NotSet,
     lazy_load=True,
     custom_schema=None,
@@ -1549,7 +1550,7 @@ def open_asdf(
 
     memmap : bool, optional
         When `True`, when reading files, attempt to memmap underlying data
-        arrays when possible. Defaults to ``True``.
+        arrays when possible. Defaults to ``False``.
 
     lazy_load : bool, optional
         When `True` and the underlying file handle is seekable, data
--- a/asdf/_tests/tags/core/tests/test_ndarray.py
+++ b/asdf/_tests/tags/core/tests/test_ndarray.py
@@ -976,7 +976,7 @@ def test_memmap_write(tmp_path):
     tmpfile = str(tmp_path / "data.asdf")
     tree = {"data": np.zeros(100)}
 
-    with asdf.AsdfFile(tree) as af:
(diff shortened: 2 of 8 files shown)


braingram:
Unit test failures are real. It looks like there are more tests that need updated for the new default.

braingram:
https://github.com/asdf-format/asdf/pull/1852 was merged and introduced the "what's new" page for asdf 4.0. Would you update this PR and add an entry describing the default change:
https://asdf.readthedocs.io/en/latest/asdf/whats_new.html#new-defaults
I'd say let's put this change first in that section given the scope of this change.

asdf/_asdf.py
```diff
@@ -57,7 +57,8 @@ def __init__(
         extensions=None,
         version=None,
         ignore_unrecognized_tag=False,
-        memmap=True,
+        ignore_implicit_conversion=NotSet,
```
zacharyburnett, line 60: remove detritus from rebasing
```suggestion
```

asdf/_tests/test_reference.py
```diff
@@ -78,7 +78,9 @@ def do_asserts(ff):
 
         assert_array_equal(ff.tree["internal"], exttree["cool_stuff"]["a"])
 
-    with asdf.AsdfFile({}, uri=(tmp_path / "main.asdf").as_uri()) as ff:
+    with asdf.AsdfFile({}, uri=(tmp_path / "main.asdf").as_uri(), memmap=True) as ff:
+        # avoid passing tree to AsdfFile to avoid the deprecation warning, this can be updated
+        # when automatic find_references on AsdfFile.__init__ is removed
```
braingram, line 83: My suggestion is to:
```suggestion
    with asdf.AsdfFile({}, uri=(tmp_path / "main.asdf").as_uri()) as ff:
```
and delete these lines:
https://github.com/asdf-format/asdf/blob/2ed29f0ad1a2452a90cff536fb3a0a0575ff3111/asdf/_tests/test_reference.py#L59-L61

The test on main is checking that an external reference to an array is memmapped. This is true when the default is `memmap=True` but is false with the default `memmap=False`. On main, when opening an `open(..., memmap=False)` the `memmap` argument is never passed to `open_external` (so it will always use the default).

I opened an issue for the pre-existing bug: https://github.com/asdf-format/asdf/issues/1867

For this PR I'd say we delete that portion of the test (as suggested above) since it doesn't check for anything specific to references (and failed to catch the above bug).

docs/asdf/whats_new.rst
```diff
@@ -46,6 +46,13 @@ Removed API
 New Defaults
 ------------
 
+File I/O
+^^^^^^^^
+
+- calls to ``asdf.open()`` and ``AsdfFile()`` do NOT memory-map the
+  underlying arrays by default; the relevant parameter defaults to
+  ``memmap=False``
+
```
braingram, line 55:
```suggestion
.. _whats_new_4.0.0_memmap

Memory mapping disabled by default
^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Calls to ``asdf.open() and ``AsdfFile()`` will now default to ``memmap=False``, disabling memory mapping of arrays by default.

```

asdf/_tests/tags/core/tests/test_ndarray.py
```diff
@@ -976,7 +976,7 @@ def test_memmap_write(tmp_path):
     tmpfile = str(tmp_path / "data.asdf")
     tree = {"data": np.zeros(100)}
 
-    with asdf.AsdfFile(tree) as af:
+    with asdf.AsdfFile(tree, memmap=False) as af:
```
braingram, line 979: Is this needed?
```suggestion
    with asdf.AsdfFile(tree) as af:
```

changes/1801.general.rst
```diff
@@ -0,0 +1 @@
+set ``memmap=False`` by default
```
braingram, line 1:
```suggestion
Set ``memmap=False`` to default for ``asdf.open`` and ``AsdfFile.__init__``.
```

braingram:
I think this PR should be good to go once the lines mentioned in https://github.com/asdf-format/asdf/pull/1801#discussion_r1834830106 are removed.

docs/asdf/whats_new.rst
```diff
@@ -46,6 +46,13 @@ Removed API
 New Defaults
 ------------
 
+.. _whats_new_4.0.0_memmap
```
braingram, line 49:
```suggestion
.. _whats_new_4.0.0_memmap:
```

I always forget the colon 🤦 

This should fix the docs.

braingram (approved):
LGTM. Thanks again for taking this on!

PR #1803, opened 2024-07-17
fix issue where roundtripping a masked array strips the mask if no values are masked

### LINKED ISSUE #1768: masked arrays do not roundtrip with all false masks
# Description of the problem

Attempting to save a masked array with an all false mask results in no errors. Yet when loaded the array is no longer masked.

See:
https://github.com/asdf-format/asdf/blob/4a0064aa48d6360d27ab588b2bb85255016e3876/asdf/_core/_converters/ndarray.py#L121-L125

I believe that removing `np.any(data.mask)` would fix the issue.

# Example of the problem

# System information

asdf version: main
python version: python 3.10
operating system: mac

zacharyburnett:
# Description

resolves #1768 

# Checklist:

- [x] tests ran successfully
- [x] ~for a public change, a changelog entry was added~
- [x] ~for a public change, documentation was updated~
- [x] for any new features, unit tests were added


braingram:
Would you add a changelog for this that documents the fixed bug?

Also, would you either move the test near:
https://github.com/asdf-format/asdf/blob/main/asdf/_tests/tags/core/tests/test_ndarray.py#L381
or update that test to the nicely parametrized one added in this PR?

braingram:
There's one more issue and I'm a little scared now (not from this PR but from what follows):
```python
import numpy as np
arr = np.arange(5)
m0 = np.ma.masked_array(arr, True)
m1 = np.ma.masked_array(arr, False)
np.testing.assert_array_equal(m0, m11)
```
The last assert does not fail :-/ so apparently assert_array_equal does not consider the mask. The same is true for `assert_equal`. Additionally (and this part is asdf's fault), `assert_tree_match` adds an extra oddity:
https://github.com/asdf-format/asdf/blob/33e7095df0c58705043de8f75497dfd64d04bbfb/asdf/_tests/_helpers.py#L55
which drops the mask.

Would you update the test to not use `assert_tree_match` and apparently (maybe any) of the numpy testing routines. I'll open an issue to check our test suite for other places where the mask is being ignored.

asdf/_tests/tags/core/tests/test_ndarray.py
```diff
@@ -378,17 +378,26 @@ def test_inline_bare():
...
     with roundtrip(tree) as af:
-        tree = af.tree
-
-        m = tree["masked_array"]
-
-        assert np.all(m.mask[6:])
+        # assert_array_equal ignores the mask, so use equality here
+        assert (tree["masked"] == af["masked"]).all()
+        # ensure tree validity
+        assert af["unmasked"] == af["masked"].data
```
braingram, line 400:
```suggestion
        assert (af["unmasked"] == af["unmasked"]).all()
```

braingram:
To add to the confusion, the test is failing because:
```python
>> (m0 == m1).all()
masked
```
when m0 and m1 are both fully masked and matching values. So I think we need to compare `m0.mask` and `m0.data` separately.

asdf/_tests/tags/core/tests/test_ndarray.py
```diff
@@ -378,17 +378,27 @@ def test_inline_bare():
...
-        tree = af.tree
-
-        m = tree["masked_array"]
-
-        assert np.all(m.mask[6:])
+        # assert_array_equal ignores the mask, so use equality here
+        assert (tree["masked"].data == af["masked"].data).all()
+        assert (tree["masked"].mask == af["masked"].mask).all()
+        # ensure tree validity
+        assert (af["unmasked"] == af["masked"].data).all()
```
braingram, line 401: I see now, my previous suggestion was wrong (it compared an array to itself). What's happening here is you're comparing the saved unmasked array to saved masked data to make sure they're equal (which they should be based on how the tree is set up). Makes sense now!

braingram (approved):
Thanks! One more bug squished.

asdf/_tests/tags/core/tests/test_ndarray.py
zacharyburnett: yeah sorry! I was confused but initially thought you might just be doing some asdf wizardry that I didn't know about 😅

PR #1809, opened 2024-07-22
improve error when an object fails to serialize

braingram:
# Description

asdf currently produces an unhelpful message for some objects that fail to serialize during `asdf.write_to`:
```python
import asdf, collections

class MyDict(collections.UserDict):
    pass

asdf.AsdfFile({"d": MyDict({"a": 1})}).write_to("foo.asdf")
```
Produces:
```
...
yaml.representer.RepresenterError: ('cannot represent an object', {'a': 1})
```

This PR adds a new exception `AsdfSerializationError` that includes the object type in the error message when yaml emits a `yaml.representer.RepresenterError` (for backwards compatibility the new exception inherits from `yaml.representer.RepresenterError`).

Running the above example with this PR produces:
```
asdf.exceptions.AsdfSerializationError: ("Object of type[<class '__main__.MyDict'>] is not serializable by asdf. Please convert the object to a supported type or implement a Converter for this type to allow the tree to be serialized.", {'a': 1})
```

Here is a related issue where the above message would have been helpful: https://github.com/spacetelescope/roman_datamodels/issues/361

# Checklist:

- [ ] tests ran successfully
- [x] for a public change, documentation was updated

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,12 @@
 
 - Fix issue where roundtripping a masked array with no masked values removes the mask [#1803]
 
+- Use a custom exception ``AsdfSerializationError`` to indicate when an object in the
+  tree fails to be serialized by asdf (and by yaml). This exception currently inherits
+  from ``yaml.representer.RepresenterError`` to provide backwards compatibility. However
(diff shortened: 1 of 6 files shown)


zacharyburnett (approved):
self-documenting errors!

PR #1810, opened 2024-07-24
DEP: drop importlib_metadata as a dependency on Python 3.12 and newer

neutrinoceros:
This is a follow up to #1260: according to [the comment linked](https://github.com/python/importlib_metadata/issues/410#issuecomment-1304258228) in that PR's commit message, standard library `importlib.metadata` isn't problematic starting from Python 3.12

# Checklist:

- [x] tests ran successfully

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,13 +6,18 @@ if sys.version_info < (3, 11):
 else:
     import tomllib
 
+from sphinx_asdf.conf import *  # noqa: F403
+
 # The standard library importlib.metadata returns duplicate entrypoints
 # for all python versions up to and including 3.11
 # https://github.com/python/importlib_metadata/issues/410#issuecomment-1304258228
 # see PR https://github.com/asdf-format/asdf/pull/1260
 # see issue https://github.com/asdf-format/asdf/issues/1254
-from importlib_metadata import distribution
-from sphinx_asdf.conf import *  # noqa: F403
+if sys.version_info >= (3, 12):
+    from importlib.metadata import distribution
+else:
+    from importlib_metadata import distribution
+
 
 # Get configuration information from `pyproject.toml`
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -6,13 +6,18 @@
...
 # for all python versions up to and including 3.11
 # https://github.com/python/importlib_metadata/issues/410#issuecomment-1304258228
 # see PR https://github.com/asdf-format/asdf/pull/1260
 # see issue https://github.com/asdf-format/asdf/issues/1254
-from importlib_metadata import distribution
-from sphinx_asdf.conf import *  # noqa: F403
+if sys.version_info >= (3, 12):
+    from importlib.metadata import entry_points
+else:
+    from importlib_metadata import entry_points
```
braingram, line 19:
```suggestion
if sys.version_info >= (3, 12):
    from importlib.metadata import distribution
else:
    from importlib_metadata import distribution
```

braingram:
Thanks for the PR. I'll try to give this a test with weldx sometime soon.

Unfortunately the downstream testing in the CI will be insufficient to test this change (the possibly affected package, weldx, is being run with python 3.10).

docs/conf.py
braingram: This should hopefully fix the docs failing to build.

neutrinoceros: whoops, thanks for catching this. Done !

braingram (approved):
I set up a local python 3.12 environment, installed the source branch for this PR and weldx (I had to remove the upper pin they have for the python version). Running the script in the previous issue succeeded and I confirmed that in python 3.12 we no longer see duplicate entry points for the standard library `importlib`:
```python
(EntryPoint(name='weldx', value='weldx.asdf.extension:get_extensions', group='asdf.extensions'),
 EntryPoint(name='asdf', value='asdf._core._integration:get_extensions', group='asdf.extensions'))
```

Thanks!

PR #1815, opened 2024-08-04
MNT: handle a deprecation warning from `attrs` v24.1.0

neutrinoceros:
# Description

`attrs` 24.1.0 (released yesterday) deprecated the `hash` argument in favor of `unsafe_hash`
For simplicity of implementation, I also bumped the minimal requirement to 22.2.0, which added the `unsafe_hash` argument.

xref https://github.com/python-attrs/attrs/pull/1323

For context, I noticed this because it breaks astropy's CI, so please let me know if this patch can be released promptly or not. We'll be able to adapt in any case.

# Checklist:

- [x] tests ran successfully

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 
 - Drop ``importlib_metadata`` as a dependency on Python 3.12 and newer [#1810]
 
+- Bumped minimal requirement on ``attrs`` from ``20.1.0`` to ``22.2.0`` [#1815]
+
 3.3.0 (2024-07-12)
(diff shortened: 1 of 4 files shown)


braingram (approved):
Thanks!

I'll try to make a release with this today.

braingram:
@neutrinoceros asdf 3.4.0 (with these changes) is up on pypi. Let me know if you run into more issues and thanks again for the fix.

neutrinoceros:
Awesome, thank you very much !

PR #1830, opened 2024-08-19
change asdf standard changelog entries to notes

braingram:
# Description

Since towncrier renders sub-sections with `-` and we include out `CHANGES.rst` in the docs we run into issues due to previous sub-sections with `^` marking the then-current ASDF-standard version.

This PR changes the `^` sub-sections to `.. note::` to avoid the conflict with towncrier.

# Checklist:

- [ ] tests ran successfully
- [ ] for a public change, added news fragment (`changes/<PR#>.<changetype>.rst`)
- [ ] for a public change, updated documentation


braingram:
I mistakenly thought `misc` fragments would be shown. This isn't the case (they default to `showcontent=False`). I'll update this PR to use `doc` instead.

I'd also like to see if we can make `misc` show content (without having to redefine all the default sections). It seems a little silly to write a fragment that will later be thrown out.

zacharyburnett:
> I mistakenly thought `misc` fragments would be shown. This isn't the case (they default to `showcontent=False`). I'll update this PR to use `doc` instead.
> 
> I'd also like to see if we can make `misc` show content (without having to redefine all the default sections). It seems a little silly to write a fragment that will later be thrown out.

that's a good point; we will unfortunately have to redefine all the non-default fragment types, but that shouldn't be too difficult

zacharyburnett (approved):

PR #1875, opened 2024-11-27
improve schema info collection combiner handling

braingram:
## Description

`schema_info` currently fails to find both subschemas and attributes of subschemas for schemas that use certain "combiners" (like `allOf`) and/or references.

This PR updates `schema_info` to support more schema configurations (including `$ref` and `allOf` usage).

~Requires https://github.com/spacetelescope/rad/pull/525 as the rad schemas contain invalid extra `$ref` items.~ EDIT: rad PR was merged see comment below about [compatibility](https://github.com/asdf-format/asdf/pull/1875#issuecomment-2546691245) (tldr; we'll want to set the asdf min in rad to >4.0.x when the changes in this PR are released)

## Tasks

</details>

### DIFF
--- a/asdf/_node_info.py
+++ b/asdf/_node_info.py
@@ -46,7 +46,7 @@ def _get_subschema_for_property(schema, key):
 
     # next handle schema combiners
     if "not" in schema:
-        # since we're only concerned here with if the schema applies
+        # Since we're only concerned here with if the schema applies
         # it doesn't matter if the schema is nested in a not
         subschema = _get_subschema_for_property(schema["not"], key)
         if subschema is not None:
@@ -75,10 +75,8 @@ def _get_schema_key(schema, key):
     applicable = []
     if key in schema:
         applicable.append(schema[key])
-    if "not" in schema:
-        possible = _get_schema_key(schema["not"], key)
-        if possible is not None:
-            applicable.append(possible)
+    # Here we don't consider any subschema under "not" to avoid
+    # false positives for keys like "type" etc.
     for combiner in ("allOf", "oneOf", "anyOf"):
         for combined_schema in schema.get(combiner, []):
             possible = _get_schema_key(combined_schema, key)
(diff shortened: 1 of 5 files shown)


braingram:
Specutils error comes from a bug in gwcs with astropy 7 that is fixed in the development version but not yet released.

The weldx failure is from weldx pinning asdf to <4.

One thing to consider with this PR is that the changes will (correctly) cause `schema_info` to not pick up `archive_catalog` info for old rad schemas (non-development versions, see https://github.com/spacetelescope/rad/pull/525). The latest released version of rad has an [asdf upper pin](https://github.com/spacetelescope/rad/blob/6fe18551a5c40bb1bdd024b10639048d41ae318e/pyproject.toml#L15) for <4. So as long as we merge and release this before the next rad release and add a rad pin for >= 4.0.1 (if 4.0.1 is the asdf release version) we should avoid issues.

asdf/_node_info.py
```diff
@@ -20,6 +20,67 @@ def _filter_tree(info, filters):
...
+            if subschema is not None:
+                applicable.append(subschema)
+
+    # only return the subschema if we found exactly 1 applicable
+    if len(applicable) == 1:
+        return applicable[0]
+    return None
+
+
+def _get_schema_key(schema, key):
```
perrygreenfield, line 66: I think it would be helpful to add explanations of what distinguishes `_get_schema_key` from `_get_subschema_for_property`.  In both cases key is an argument, in the first the key presumably is expected to be one of the properties, and in the second, `key` is not a property. Examples of the different usage contexts to distinguish the two might be useful.

perrygreenfield (commented):
Mainly asking for more helpful code documentation.

asdf/_node_info.py
braingram: Good idea! How does e39d1e18b83e3c2b3099cac7bff7a4e85f34ebc6 look?

perrygreenfield: LGTM

perrygreenfield (approved):
LGTM

PR #1894, opened 2025-01-12
pyproject.toml: version_file requires setuptools_scm>=8

wdconinc:
## Description

When the switch was made from `write_to` to `version_file`, the version requirement on setuptools_scm was not increased. `version_file` is only available as of v8, https://github.com/pypa/setuptools-scm/blob/v8.0.4/CHANGELOG.md?plain=1#L62.

## Tasks

Most tasks below are not applicable.


braingram (approved):
Thanks for finding the mistake!

PR #1906, opened 2025-03-10
Use fsspec for all non-local urls

braingram:
## Description

This PR uses fsspec for all non-local urls (including `http`). As fsspec is an optional dependency this PR:
- adds a deprecation warning for `http` urls stating that in a future version fsspec will be required and then the existing http handling code is used
- adds a new optional dependency group `http` to `pyproject.toml` that installs fsspec with http support (and adds this dependency group to `all`)

The above change allows for native handling of `s3` urls (instead of requiring opening the url first in `s3fs`) allowing the following to work:
```python
asdf.open("s3://bucket/file.asdf")
```

Adding a test for this behavior was complicated by:
- the lack of a suitable small public asdf file on s3 to use for testing (so we need to mock s3)
- the mock s3 (moto) is picky about it's dependencies, often conflicting with `boto` (the aws client), adding these as test dependencies for all runs seemed like it would make the test runs brittle
- the s3 mocking needs to happen before any fsspec import. This could be done as part of the unit tests but that seemed overly complicated (as most tests don't require the mock s3). Given that the overall utility of the s3 tests is minimal (if we assume fsspec is doing what it advertises by abstracting out the backend, testing http should be sufficient) I chose to isolate the s3 tests (more on that below).

The s3 tests were added to `integration_tests/s3` to:
- keep them out of the packaged wheel
- keep them out of normal pytest runs (so downstream package maintainers don't have to run these tests as part of their packaging workflows)
- allow the s3 mocking to occur for the entire session (without interfering with other tests)

Since the `compatibility_tests` are sort of `integration_tests` I moved them into the same subdirectory (so they're now at `integration_tests/compatibility`.

I'll call out a couple other changes inline.

## Tasks

</details>


asdf/generic_io.py
```diff
@@ -402,7 +403,8 @@ def close(self):
         """
         if self._close:
             self._fd.close()
-            self._fix_permissions()
+            if hasattr(self, "_fix_permissions"):
```
braingram, line 406: This is a minor bugfix that I found when allowing fsspec to natively handle http urls. The `_fix_permissions` method is only present in the RealFile subclass.

pyproject.toml
```diff
@@ -33,8 +33,15 @@ dependencies = [
 ]
 [project.optional-dependencies]
 all = [
+  "asdf[lz4]",
+  "asdf[http]",
+]
+lz4 = [
```
braingram, line 39: Although not required for this PR since I was changing `optional-dependencies` I figured I'd separate out the lz4 one so that downstream users (like roman_datamodels) could use `asdf[lz4]`.

pyproject.toml
```diff
@@ -43,12 +50,14 @@ docs = [
...
 ]
 tests = [
-  "fsspec[http]>=2022.8.2",
-  "lz4>=0.10",
+  "asdf[all]",
   "psutil",
   "pytest>=8",
   "pytest-remotedata",
 ]
+test = [
```
braingram, line 58: I snuck this in because I often type `pip install asdf[test]` then it complains that `tests` was expected. I figured we could just make both work.

zacharyburnett: I do this too 😅

tox.ini
```diff
@@ -1,6 +1,7 @@
 [tox]
 env_list =
     compatibility
+    s3
```
zacharyburnett, line 4: I would rename this `mocks3` or something to be more explicit about why it is separated

zacharyburnett (approved):
LGTM, good isolation solution

braingram:
Testing on the roman research nexus (which has some ASDF files on s3) with this PR and roman_datamodels (post https://github.com/spacetelescope/roman_datamodels/pull/453) I was able to run:
```python
m = rdm.open("s3://...")
```
successfully.

Issue #1948, opened 2025-07-25
Failure to convert nested object raises exception instead of warning


emolter:
Attempting to open an asdf file that requires a converter to deserialize some portion of the file raises an exception when that converter is not found, e.g. if the parent package decided to drop support for it.  It might be more helpful if this raised a warning and loaded the un-convert-able parts as TaggedDict.

braingram:
To expand on this a bit. I propose that we consider.
1) Add a `warn_on_failed_conversion` option to `AsdfConfig`
2) Set the default option of above to `False` (to retain backwards compatibility)
3) when above is enabled, catch converter errors and turn them into warnings

This would allow a user to:
```python
import asdf
asdf.config.get_config().warn_on_failed_conversion = False
tree_with_broken_data = asdf.load("broken_file")
# inspection of the tree is possible
```

emolter:
As we discussed in person, that plan sounds great to me.  Let me know if you need any support from me with that.

PR #1983, opened 2025-10-20
allow failed conversion

braingram:
## Description

At the moment a test PR to run downstream tests with this feature enabled. Some asdf unit tests fail when it's enabled so some downstream failures would not be unexpected.

So far there are:
- asdf unit test failures
- asdf-astropy unit test failure

Taking one of the failing asdf unit tests as an example:
https://github.com/asdf-format/asdf/blob/52149a99812f3dfaaaaf57f948e1727d32ac044d/asdf/_tests/_regtests/test_1541.py#L31-L34
This test enables block checksum checking. The test (as is) passes if "lazy_load=True" but fails if "lazy_load=False". When "lazy_load=False" the checksum is checked within "load_yaml_tree" which with this PR (as is) the error gets converted to a warning. I think this is a good example of why we don't want to do a blanket convert all converter errors to warnings and instead make this "opt-in". Users could then enable the "allow failed conversion" feature if they fail to open a file.

Closes: https://github.com/asdf-format/asdf/issues/1948

## Tasks

</details>

### DIFF
--- a/asdf/_tests/_regtests/test_1948.py
+++ b/asdf/_tests/_regtests/test_1948.py
@@ -1,124 +1,133 @@
+"""
+Test that warn_on_failed_conversion works as expected for:
+    - lazy and non-lazy trees
+    - nested objects
+    - generator producing (yielding) converters
+
+Broken out into a regtest since there is significant
+test setup.
+
+https://github.com/asdf-format/asdf/issues/1948
+"""
+
 import pytest
 
 import asdf
 
-foo_tag_uri = "asdf://somewhere.org/tags/foo-1.0.0"
-bar_tag_uri = "asdf://somewhere.org/tags/bar-1.0.0"
-bam_tag_uri = "asdf://somewhere.org/tags/bam-1.0.0"
+test_dict_tag_uri = "asdf://somewhere.org/tags/test_dict-1.0.0"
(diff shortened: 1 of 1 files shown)


asdf/lazy_nodes.py
```diff
@@ -109,7 +110,11 @@ def _to_lazy_node(node, af_ref):
     If the object does not have a corresponding subclass
     it will be returned unchanged.
     """
-    if isinstance(node, list):
+    if isinstance(node, tagged.TaggedList):
```
braingram: I'm going to investigate this further but I think there may be a related bug where a unknown tag can cause more distal tree portions to go unconverted for a lazy tree. I may roll that fix into this PR (since it was revealed by the added tests). More investigation to follow.

braingram: See https://github.com/asdf-format/asdf/pull/1983/commits/c7b1c18803cfc0a1ae9cd9409c58f4aa2f242044 for an updated unit test. In brief, when an unknown tag was encountered for a lazy tree it resulted in:
- loss of the tag
- return of the non-tagged dict/list

With this PR the result is a `Tagged` object that:
- keeps the tag
- for `TaggedDict` and `TaggedList` the backing dict/list is lazy

Issue #1683, opened 2023-11-14
asdf expects that block compression label bytes are valid ascii


braingram:
Here is one example:
https://github.com/asdf-format/asdf/blob/ecb4d4f6308738c4de0da8eec5bd272e90f3954e/asdf/compression.py#L183

The standard doesn't mention that these 4 bytes should be limited to the ascii range:
https://asdf-standard.readthedocs.io/en/1.0.3/file_layout.html#block-header

Modifying the [LzmaCompressor](https://github.com/asdf-format/asdf/blob/ecb4d4f6308738c4de0da8eec5bd272e90f3954e/asdf/_tests/test_compression.py#L196) used in `test_compression_with_extension` to use label `b"lzm\255" results in a failure at:
https://github.com/asdf-format/asdf/blob/ecb4d4f6308738c4de0da8eec5bd272e90f3954e/asdf/compression.py#L201
```python
UnicodeDecodeError: 'ascii' codec can't decode byte 0xad in position 3: ordinal not in range(128)
```

I can see the benefit of having these be just ascii (they are human readable when the file is opened with a text editor). If this restriction is ok than the standard should be updated to reflect that these bytes must be ascii.

PR #2014, opened 2026-04-01
`asdf info` block display

### LINKED ISSUE #1923: add block display to asdf info
libasdf has an option to display block contents: https://github.com/asdf-format/libasdf/issues/2

Add a similar option to `asdf info`. This certainly makes sense for the command line tool (where a filename is the input) but I'm not sure (and doubtful) that it would be informative for `AsdfFile.info` (where the concept of blocks is more removed from the info tree).

sydduckworth:
## Description

* Added block table output which is optionally printed after the tree
* Added `show_blocks` parameter to `AsdfFile.info` method which defaults to `False`
* Added `--show-blocks` argument to `info` CLI command
* Format matches the equivalent `libasdf` command

Closes #1923

## Example Output

```console
❯ asdftool info --show-blocks basic.asdf 
root (AsdfObject)
├─asdf_library (Software)
│ ├─author (str): The ASDF Developers
│ ├─homepage (str): http://github.com/asdf-format/asdf
│ ├─name (str): asdf
│ └─version (str): 4.1.0
├─history (dict)
│ └─extensions (list)
│   └─[0] (ExtensionMetadata)
│     ├─extension_class (str): asdf.extension._manifest.ManifestExtension
│     ├─extension_uri (str): asdf://asdf-format.org/core/extensions/core-1.6.0
│     ├─manifest_software (Software)
│     │ ├─name (str): asdf_standard
│     │ └─version (str): 1.1.1
│     └─software (Software)
│       ├─name (str): asdf
│       └─version (str): 4.1.0
└─data (NDArrayType)
  ├─shape (tuple)
  │ └─[0] (int): 8
  └─dtype (Int64DType): int64
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃                     Block #0                      ┃
┣━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ flags          │            0x00000000            ┃
┃ compression    │               None               ┃
┃ allocated_size │                64                ┃
┃ used_size      │                64                ┃
┃ data_size      │                64                ┃
┃ checksum       │ 35594cae5fb11be3ea419c26bc4cfbee ┃
┗━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛

```


asdf/_asdf.py
```diff
@@ -1164,9 +1164,10 @@ def schema_info(self, key="description", path=None, preserve_list=True):
 
     def info(
         self,
-        max_rows=display.DEFAULT_MAX_ROWS,
-        max_cols=display.DEFAULT_MAX_COLS,
-        show_values=display.DEFAULT_SHOW_VALUES,
+        max_rows: int | tuple[int, ...] | None = display.DEFAULT_MAX_ROWS,
+        max_cols: int | None = display.DEFAULT_MAX_COLS,
+        show_values: bool = display.DEFAULT_SHOW_VALUES,
```
braingram:
```suggestion
        max_rows: int = display.DEFAULT_MAX_ROWS,
        max_cols: int = display.DEFAULT_MAX_COLS,
```
Before we introduce type hints let's add mypy to the pre-commit hooks.

asdf/_display.py
```diff
@@ -25,6 +26,30 @@
 DEFAULT_SHOW_VALUES = True
 
 
+def render_table(
+    title: str,
+    rows: list[tuple[str, str]],
+) -> list[str]:
```
braingram:
```suggestion
    title
    rows
)
```
Same as other comment about mypy.

changes/2014.feature.rst
```diff
@@ -0,0 +1 @@
+Added optional block info table to ``asdftool info`` and ``AsdfFile.info``.
```
braingram, line 1: Would you also update `asdf.info`?
https://github.com/asdf-format/asdf/blob/f97c7a6f1c66fe2847cae00523d17cbf4ec1a900/asdf/_convenience.py#L15

asdf/_asdf.py
```diff
@@ -1201,6 +1206,24 @@ def info(
         )
         print("\n".join(lines))
 
+        if show_blocks:
+            for i, block in enumerate(self._blocks.blocks):
+                if block.header["compression"] == b"\0\0\0\0":
+                    compression = "none"
+                else:
+                    compression = block.header["compression"].decode()
```
braingram: There is no restriction in the specification that these be valid ascii:
https://github.com/asdf-format/asdf/issues/1683
How do you think we should handle non-ascii values?

asdf/_display.py
```diff
@@ -25,6 +26,30 @@
 DEFAULT_SHOW_VALUES = True
 
 
+def render_table(
```
braingram, line 29: Testing with a roman file (where the compression code "lz4" is <4) the table formatting breaks:
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃                     Block #26                     ┃
┣━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ flags          │            0x00000000            ┃
┃ compression    │               lz4               ┃
┃ allocated_size │              131222              ┃
┃ used_size      │              131222              ┃
┃ data_size      │             33423488             ┃
┃ checksum       │ 89f82a03128c2237929ecbe3c0fff2e9 ┃
┗━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```
What's needed to fix that?

asdf/_asdf.py
```diff
@@ -1201,6 +1206,24 @@ def info(
         )
         print("\n".join(lines))
 
+        if show_blocks:
+            for i, block in enumerate(self._blocks.blocks):
+                if block.header["compression"] == b"\0\0\0\0":
+                    compression = "none"
```
braingram: libasdf displays no compression as `""`. For `set_array_compression` either an empty string or `None` signifies "no compression". To keep this consistent I'd say we show one of the following:
- whatever decision we use for non-ascii bytes
- `""` (consistent with libasdf and `set_array_compression`)

What do you think?

asdf/_tests/test_info.py
```diff
@@ -832,3 +832,33 @@ def test_info_no_infinite_loop(capsys):
...
+
+
+def test_info_blocks_show(capsys, test_data_path):
+    """Verify blocks are printed when `show_blocks=True`."""
+
+    file_path = test_data_path / "ndarray0.asdf"
+    with asdf.open(file_path) as af:
+        af.info(show_blocks=True)
+        captured = capsys.readouterr()
+        assert "Block #0" in captured.out
```
braingram: Would you expand these tests (or other or new tests) to check for:
- expected table formatting for files with 1 block and files with multiple (>=10) blocks
- expected table contents: flags (streamed, not streamed), checksum (0 or some value), sizes (where these differ due to compression), compression (no compression, some compression), indices

braingram (commented):
Thanks for working on this. I left a few comments. Overall this is an improvement to the inspectability of files.

As noted I'm not against adding type hints. To check those are correct we'd benefit from first adding a type checker. If you're interested in adding type hints would you make a separate PR that sets up mypy for this project as a pre-commit hook (and addresses any configuration and other issues that might uncover)?

Once the comments are addressed would you update this PR with a comment showing some example block info output?

asdf/_asdf.py
sydduckworth: My thinking was that `\0\0\0\0` is a constant defined by the ASDF standard, so `none` or `None` conveys semantic meaning to a user whereas an empty string or an unencoded byte array requires you to have knowledge of the ASDF format to interpret.

asdf/_asdf.py
sydduckworth: Can we assume trailing `0x00` bytes can be ignored?

asdf/_tests/test_info.py
sydduckworth: What's the policy on adding dev dependencies? This kind of testing could be expressed a lot more compactly using [syrupy](https://github.com/syrupy-project/syrupy) to do a snapshot test of the command output.

asdf/_display.py
sydduckworth: Fixed. Trailing `\0` bytes are now stripped before printing.

asdf/_asdf.py
sydduckworth: Okay, the command now formats any non-ascii characters using slash escape sequences.

changes/2014.feature.rst
sydduckworth: Done!

asdf/_tests/test_info.py
braingram: Are you thinking adding that as a test dependency? It looks like a great match to the needs. Am I correct that the "amber" files here would be essentially plain text? If so sounds good to me.

sydduckworth: Yep, test dependency. And yeah, the amber files have some structure to them but they are human-readable and Git can handle diffing them and whatnot.

asdf/_asdf.py
braingram: My hesitation for `none` or `None` is that no compression is indistinguishable from a compression with code "none" or "None" and a user might think:
```python
af.set_array_compression(arr, "none")
```
should work. `None` is a little less misleading (in that `af.set_array_compression(arr, None)` does work as expected but `af.set_array_compression(arr, "None")` does not).

`""` works as expected but as you note doesn't convey that no compression was used. Maybe instead we make the line say something like:
```
compressed using "lz4"
```
or
```
not compressed
```
That would deviate from what libasdf currently shows. Do you think that would be more helpful for someone less knowledgeable about the ASDF format?

sydduckworth: We could have it match the `set_array_compression` by having no compression as the literal `None` and the string compression names in quotes, e.g. `"lz4"`. That way a user could copy-paste the info output into `set_array_compression` and it would work.

braingram: Good idea! Does that seem like a workable solution to you?

asdf/_tests/test_info.py
sydduckworth: Okay, tests have now been updated to full snapshot tests, and I also added a new test case with multiple blocks and different types of compression.

.pre-commit-config.yaml
```diff
@@ -15,9 +15,9 @@ repos:
     - id: debug-statements
     - id: detect-private-key
     - id: end-of-file-fixer
-      exclude: ".*fits"
+      exclude: "(.*fits)|(.*\\.ambr)"
```
braingram: Do we ever want pre-commit to touch these `.ambr` files? If not, would you add a global ignore?

pyproject.toml
braingram, line 1: Would you undo the unrelated formatting changes to this file?

I don't have any objection to formatting this file but let's split that into a different PR since it's unrelated.

asdf/_asdf.py
braingram: I unresolved this as the output is still showing `none` instead of `None` and the string compression names in quotes (`"lz4"`).

asdf/_tests/test_info.py
braingram: The added ambr files look great and make reviewing these changes much easier. I suspect there are other tests that could be improved by switching to this pattern. Would you open a follow-up issue to look for and update those tests?

pyproject.toml
sydduckworth: Done

.pre-commit-config.yaml
sydduckworth: Done

asdf/_asdf.py
sydduckworth: Whoops for some reason I thought I had already done this! Should now be updated.

asdf/_tests/test_info.py
sydduckworth: #2023

asdf/_tests/data/ndarray2.asdf
braingram, line 1: Would you regenerate this file with `pad_blocks=True`? That way the ambr files will have different values for `used_size` and `allocated_size` (to test that those don't get mixed up in future changes).

braingram (commented):
One minor request to update the test data with padded blocks so that `used_size` != `allocated_size`.

asdf/_tests/data/ndarray2.asdf
sydduckworth: Done

braingram (approved):
Thanks! This looks great and funny enough I was handed an ASDF file today that I already wanted to use this new feature on :)

PR #2016, opened 2026-04-08
TEST: break the checksum

braingram:
Test PR breaking the checksum to see how that impacts downstream.


braingram:
No failures in downstream unit tests if we generate incorrect checksums.

jwst/romancal regression tests weren't run.

PR #2018, opened 2026-04-10
Support not writing checksums

### LINKED ISSUE #2015: Block checksum
The asdf specification [section describing the block header](https://github.com/spacetelescope/romancal/actions/runs/23652437489/job/68900650432#step:11:151) contains the following description of the checksum:
> checksum (16-byte string): An optional MD5 checksum of the used data in the block. The special value of all zeros indicates that no checksum verification should be performed.

The mention of "used data" presumably means the compressed bytes (for compressed data). However this library is storing the checksum for the uncompressed bytes.

```python
import hashlib
import asdf, numpy as np

arr = np.arange(100)
uncompressed_checksum = hashlib.md5(arr).digest()
print(f"Uncompressed checksum: {uncompressed_checksum}")

fn = "foo.asdf"

asdf.dump({"arr": arr}, fn, all_array_compression="lz4")

af = asdf.open(fn)
blk = af._blocks.blocks[0]
assert blk.header["checksum"] == uncompressed_checksum

f = open(fn, "rb")
f.seek(blk.data_offset)
bs = f.read(blk.header["used_size"])

compressed_checksum = hashlib.md5(bs).digest()
print(f"Compressed checksum: {compressed_checksum}")
assert blk.header["checksum"] != compressed_checksum
```

This was uncovered by ASDF.jl developers:
https://juliaastro.org/ASDF.jl/previews/PR19/examples/roman/#Note-cbaa8683607656d3

sydduckworth:
## Description
* Added support for disabling block checksums in file write methods. Writing checksums is on by default.
* Added `write_checksums` parameter to `AsdfFile.write_to`, `AsdfFile.update`, `asdf.dump`, and `asdf.dumps` with default value `True`

Closes #2015

## Tasks

</details>


asdf/config.py
```diff
@@ -27,6 +27,7 @@
 DEFAULT_DEFAULT_ARRAY_SAVE_BASE = True
 DEFAULT_LAZY_TREE = False
 DEFAULT_WARN_ON_FAILED_CONVERSION = False
+DEFAULT_WRITE_CHECKSUMS = False
```
braingram:
```suggestion
```

The values here are defaults for `AsdfConfig`. Do you think it would be easier/an improvement to implement `write_checksums` as an `AsdfConfig` property? If not, let's find a new place to define this default.

asdf/_tests/commands/test_validate.py
```diff
@@ -8,7 +8,7 @@
 @pytest.fixture()
 def valid_file_path(tmp_path):
     path = tmp_path / "valid_file.asdf"
-    asdf.dump({"foo": 42, "arr": np.arange(42)}, path)
+    asdf.dump({"foo": 42, "arr": np.arange(42)}, path, write_checksums=True)
```
braingram, line 11: Which test(s) required checksums for this fixture result?

asdf/_tests/_block/test_reader.py
```diff
@@ -51,7 +61,15 @@ def test_read(tmp_path, lazy_load, memmap, with_index, validate_checksums, paddi
...
     size = 10
-    with gen_blocks(fn=fn, n=n, size=size, padding=padding, with_index=with_index, streamed=streamed) as (fd, check):
+    with gen_blocks(
+        fn=fn,
+        n=n,
+        size=size,
+        padding=padding,
+        with_index=with_index,
+        streamed=streamed,
+        write_checksums=validate_checksums,
```
braingram: Would you help me understand what requires this? I'm confused why `write_checksums` is set to `validate_checksums`.

asdf/_tests/_block/test_io.py
```diff
@@ -18,12 +20,32 @@ def test_checksum(tmp_path):
     # check that when written, a block generates the correct checksum
     path = tmp_path / "test"
     with generic_io.get_file(path, mode="w") as fd:
-        bio.write_block(fd, my_array.view(dtype="uint8"))
+        # check that no warnings are raised when writing checksums without compression
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            bio.write_block(fd, my_array.view(dtype="uint8"), write_checksum=True)
```
braingram:
```suggestion
        bio.write_block(fd, my_array.view(dtype="uint8"), write_checksum=True)
```
Uncaught warnings are turned into errors:
https://github.com/asdf-format/asdf/blob/f97c7a6f1c66fe2847cae00523d17cbf4ec1a900/pyproject.toml#L109
and this is enforced in the conftest:
https://github.com/asdf-format/asdf/blob/f97c7a6f1c66fe2847cae00523d17cbf4ec1a900/asdf/conftest.py#L13

asdf/_tests/_block/test_io.py
```diff
@@ -18,12 +20,32 @@ def test_checksum(tmp_path):
...
 
+def test_checksum_compression_warning(tmp_path):
+    my_array = np.arange(0, 64, dtype="<i8")
+
+    path = tmp_path / "test"
+    with generic_io.get_file(path, mode="w") as fd:
+        # check that writing checksums with compression enabled raises a warning
+        with pytest.warns(AsdfWarning):
+            warnings.simplefilter("always")
+
```
braingram:
```suggestion
```
Is this filter change needed? It was my understanding that pytest handled this for us:
https://github.com/pytest-dev/pytest/blob/2d3d16907540149a9d8bd5b58d5a767e513718d7/src/_pytest/recwarn.py#L244

asdf/_tests/_block/test_io.py
```diff
@@ -18,12 +20,32 @@ def test_checksum(tmp_path):
...
     assert header["checksum"] == target_checksum
 
 
+def test_checksum_compression_warning(tmp_path):
+    my_array = np.arange(0, 64, dtype="<i8")
+
+    path = tmp_path / "test"
+    with generic_io.get_file(path, mode="w") as fd:
+        # check that writing checksums with compression enabled raises a warning
+        with pytest.warns(AsdfWarning):
```
braingram: Would you add a "match" here?

asdf/_block/io.py
```diff
@@ -298,14 +306,28 @@ def generate_write_header(data, stream=False, compression_kwargs=None, padding=F
...
         used_size = header_kwargs["data_size"]
         buff = None
     else:
+        # Raise a warning if checksums are enabled but only if we're actually writing a compressed block
+        if write_checksum:
+            warnings.warn(
+                (
+                    "Writing compressed block with block checksums enabled. "
+                    "Checksums generated by this library may be unreadable by other ASDF implementations "
+                    "when block compression is also enabled."
```
braingram:
```suggestion
                    "Writing checksums with compressed blocks produces files that may be difficult to open in other libraries or other versions of asdf. See: <some docs link> for more information"
```
There is some nuance here that may warrant a dedicated docs page. We'll also want to update this when this library is updated to write out the checksum of the compressed bytes (to match the specification).

Would you add that docs page to this PR? Some things to highlight are:
- a description of the inconsistency between the python implementation and the specification and what that means for end users and developers
- which versions of asdf wrote out checksums for compressed blocks based on the uncompressed contents
- a description of the stop-gap measure we're putting in place for this PR (writing all blank checksums)
- which versions of asdf wrote out checksums for compressed blocks based on the compressed contents (TBD for now, should we make that a goal for this next release?)
- how to handle these files in old/new versions
- how to handle this in other libraries (which should ideally provide a way to skip checksum validation)

asdf/_asdf.py
```diff
@@ -839,6 +840,11 @@ def update(
         version : str, optional
             Update the ASDF core schemas version of this AsdfFile before
             writing.
+
+        write_checksums: bool or None, optional
+            Compute and write block checksums to the file.
+            If `None` then checksums will be written if the existing blocks
+            in the file contain checksums. Otherwise defaults to `False`.
```
braingram: How might a user benefit from `None` here? It seems easier to understand what will happen (and implement this) if we restrict this to a bool. (FWIW `update` is rarely used).

braingram (commented):
Thanks for putting this together. I left a few comments.

asdf/config.py
sydduckworth: I don't think it should go in `AsdfConfig` but there didn't seem to be a public module where it would make more sense to put it. Do you have a suggestion?

asdf/_tests/commands/test_validate.py
sydduckworth: `asdf.open` fails if checksums are missing and `validate_checksums` is set, and the validate command has `validate_checksums` on by default unless `--skip-block-validation` is passed.

asdf/_tests/_block/test_reader.py
sydduckworth: Originally I had added another parameterization to the test function for whether to write checksums, but as mentioned in another comment reading blocks without checksums fails if `validate_checksums` is set.
So this is an attempt to ensure test coverage of with/without checksums in combination with the other write options.

asdf/_asdf.py
sydduckworth: It seems like the least surprising behavior. If I am updating a file that contains existing blocks with checksums then as a user I wouldn't expect the checksums to be silently stripped from the file.

asdf/_tests/_block/test_io.py
sydduckworth: I've added a specific warning subclass instead.
I'd prefer not to add a `match` parameter because in my opinion error/warning message text should not be treated as part of a public API contract. 
Let me know if you don't think that change is sufficient and I can add the `match` regex.

asdf/_block/io.py
sydduckworth: Okay, where do you think that page should go? Making it a top-level page feels like it would be adding too much noise since it wouldn't be relevant to most users.
I could add it as a section in the block compressors page?

asdf/_tests/commands/test_validate.py
braingram: Well... that changes things. So that means writing a file with all 0 checksums will fail to load if validate_checksums is enabled?

sydduckworth: Apparently so. I can also change it so that `validate_checksums` ignores zero checksums? That seems like its more in line with the spec anyway.

sydduckworth: If we wanted to support the old behavior we could also change `validate_checksums` to accept either a bool or an enum like:

```py
class ValidateChecksums(Enum):
    ON = auto()
    OFF = auto()
    STRICT = auto()
```

Where `ValidateChecksums.STRICT` replicates the old behavior and `ValidateChecksums.ON` only validates them if they're nonzero.

braingram: Thanks. We should rethink the strategy in https://github.com/asdf-format/asdf/issues/2015 since an all 0 checksum doesn't work with old clients if they validate checksums. How about we move the discussion back to that issue, put this PR back in draft (until we reassess the plan) and also discuss this offline.

changes/2018.feature.rst
braingram, line 1: Would you rename this to `feature`?

asdf/_block/io.py
```diff
@@ -275,7 +275,9 @@ def callback():
     return offset, header, data_offset, data
 
 
-def generate_write_header(data, stream=False, compression_kwargs=None, padding=False, fs_block_size=1, **header_kwargs):
+def generate_write_header(
+    data, stream=False, compression_kwargs=None, padding=False, fs_block_size=1, write_checksum=False, **header_kwargs
```
braingram:
```suggestion
    data, stream=False, compression_kwargs=None, padding=False, fs_block_size=1, write_checksum=True, **header_kwargs
```

asdf/_block/io.py
```diff
@@ -370,7 +376,9 @@ def generate_write_header(data, stream=False, compression_kwargs=None, padding=F
     return header_kwargs, buff, padding_bytes
 
 
-def write_block(fd, data, offset=None, stream=False, compression_kwargs=None, padding=False, **header_kwargs):
+def write_block(
+    fd, data, offset=None, stream=False, compression_kwargs=None, padding=False, write_checksum=False, **header_kwargs
```
braingram:
```suggestion
    fd, data, offset=None, stream=False, compression_kwargs=None, padding=False, write_checksum=True, **header_kwargs
```

asdf/_block/io.py
```diff
@@ -307,6 +309,10 @@ def generate_write_header(data, stream=False, compression_kwargs=None, padding=F
         The filesystem block size. See the `asdf.util.calculate_padding`
         ``block_size`` argument for more details.
 
+    write_checksum: bool, optional, default False
```
braingram:
```suggestion
    write_checksum: bool, optional
```
Removing the default here (since it's in the function signature).

asdf/_block/io.py
```diff
@@ -392,6 +400,10 @@ def write_block(fd, data, offset=None, stream=False, compression_kwargs=None, pa
     padding : bool, optional, default False
         Optionally pad the block data. See `generate_write_header`.
 
+    write_checksum: bool, optional, default False
```
braingram:
```suggestion
    write_checksum: bool, optional
```

asdf/_block/manager.py
```diff
@@ -596,11 +600,15 @@ def update(self, new_tree_size, pad_blocks, include_block_index):
             a number of padding bytes based off a ratio of the data
             size.
 
-        include_block_index : bool
+        include_block_index : bool or None
```
braingram, line 603:
```suggestion
        include_block_index : bool
```
Is there code that calls this with `None`?

asdf/_block/manager.py
```diff
@@ -596,11 +600,15 @@ def update(self, new_tree_size, pad_blocks, include_block_index):
...
             If True, include a block index at the end of the file.
             If a streamed_block is provided (or the file is not
             seekable) no block index will be written.
 
+        write_checksums: bool or None, optional
+            Compute and write block checksums to the file.
+            If `None` then checksums will be written if the existing blocks
+            in the file contain checksums. Otherwise defaults to `False`.
```
braingram:
```suggestion
        write_checksums: bool, optional
            Compute and write block checksums to the file.
```
From local testing `None` is interpreted as `False` here. If I open a file with checksums and run `update(write_checksums=None)` the checksums are blank.

I'm in favor of supporting only `True/False` (as the code appears to currently do) and updating this docstring to remove mention of `None`. This does mean a user will need to explicitly disable or enable checksum writing but given the little usage of `update` I think requiring the user to provide `True/False` is easier to understand.

asdf/_block/writer.py
```diff
@@ -29,7 +29,7 @@ def data_bytes(self):
         return np.ndarray(0, np.uint8)
 
 
-def write_blocks(fd, blocks, padding=False, streamed_block=None, write_index=True):
+def write_blocks(fd, blocks, padding=False, streamed_block=None, write_index=True, write_checksums=False):
```
braingram:
```suggestion
def write_blocks(fd, blocks, padding=False, streamed_block=None, write_index=True, write_checksums=True):
```

asdf/_tests/_block/test_io.py
```diff
@@ -139,7 +140,8 @@ def test_write_block(tmp_path):
         + b"\0\0\0\0\0\0\0\0"  # allocated size
         + b"\0\0\0\0\0\0\0\0"  # used size
         + b"\0\0\0\0\0\0\0\0"  # data size
-        + b"\xd4\x1d\x8c\xd9\x8f\x00\xb2\x04\xe9\x80\t\x98\xec\xf8B~"  # checksum
+        + b"\0" * 16  # checksum is disabled
+        # + b"\xd4\x1d\x8c\xd9\x8f\x00\xb2\x04\xe9\x80\t\x98\xec\xf8B~"  # checksum
```
braingram:
```suggestion
        + b"\xd4\x1d\x8c\xd9\x8f\x00\xb2\x04\xe9\x80\t\x98\xec\xf8B~"  # checksum
```
This assumes the other suggestions setting the default `write_checksum=True` are accepted.

asdf/_tests/_block/test_manager.py
```diff
@@ -76,7 +76,7 @@ def test_update_outside_context(tmp_path):
     with asdf.open(fn) as af:
         m = af._blocks
         with pytest.raises(OSError, match=r"update called outside of valid write_context"):
-            m.update(0, False, False)
+            m.update(0, pad_blocks=False, include_block_index=False, write_checksums=None)
```
braingram:
```suggestion
            m.update(0, pad_blocks=False, include_block_index=False, write_checksums=False)
```

asdf/_tests/_block/test_reader.py
braingram: Would you split this out into a separate parametrization now that the https://github.com/asdf-format/asdf/pull/2027 is merged?

braingram (commented):
Thanks. Overall looks good to me. I can see this option as being useful in generating reference files without checksums. I left some comments about:
- switching some internal defaults to writing out checksums
- removing `None` as an option for update (this was interpreted as `False`)
- and some other minor comments

sydduckworth:
@braingram all changes made! I previously updated some of the function signatures to default `True` and get rid of `None`, can't believe I missed so many places!

braingram (approved):
Thanks! This looks great.

PR #2024, opened 2026-04-15
Fixed empty checksum validation

### LINKED ISSUE #2019: asdf considers all blank checksums as invalid
The specification contains:
```
checksum (16-byte string): An optional MD5 checksum of the used data in the block. The special value of all zeros indicates that no checksum verification should be performed.
```

However asdf does not allow all 0 checksums for non-streamed blocks.

Current relevant code is here:
https://github.com/asdf-format/asdf/blob/f97c7a6f1c66fe2847cae00523d17cbf4ec1a900/asdf/_block/reader.py#L79-L85

sydduckworth:
## Description

* Fixed bug where missing/all-zero block checksums were treated as invalid when `validate_checksums` is enabled

Fixes #2019

### DIFF
--- a/asdf/_block/reader.py
+++ b/asdf/_block/reader.py
@@ -76,11 +76,14 @@ class ReadBlock:
             data = self._data()
         else:
             data = self._data
+
         if self.validate_checksum:
-            checksum = bio.calculate_block_checksum(data)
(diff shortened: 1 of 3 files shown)


braingram (approved):
Thanks for the quick fix. Looks good to me.

PR #2025, opened 2026-04-15
TOML formatting and validation

### LINKED ISSUE #2022: Add TOML formatting to Precommit
Currently `pyproject.toml` doesn't follow any particular formatting convention. This creates a problem for the many developers (such as myself) who have format-on-save enabled in their editor, since it means that changing a TOML file introduces many other unrelated formatting changes.  
We would benefit from adding a TOML formatting step to the precommit hooks (and CI/CD by extension) and updating the formatting for existing TOML files.

For the specific formatter [Taplo](https://taplo.tamasfe.dev/) seems like the best choice. It has first-party VSCode support and can easily be configured for use in other editors via its CLI.
It features both formatting and validation, so we could also replace the existing pyproject validation hook.

Taplo doesn't have first-party precommit support. Someone did create a 3rd-party hook [here](https://github.com/ComPWA/taplo-pre-commit) but it might be better to just set it up as a custom hook since it only requires a few lines.

sydduckworth:
## Description

* Added Taplo format and validate steps to precommit config
* Formatted `pyproject.toml`
* Updated `pyproject.toml` to conform to Pyproject schema
* Removed `validate-pyproject` precommit hook

Closes #2022

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,11 +58,6 @@ repos:
   hooks:
     - id: blacken-docs
 
-- repo: https://github.com/abravalheri/validate-pyproject
-  rev: "v0.24.1"
-  hooks:
-    - id: validate-pyproject
-
 - repo: local
   hooks:
     - id: taplo-format
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,23 +58,8 @@ repos:
   hooks:
     - id: blacken-docs
 
-- repo: local
+- repo: https://github.com/tombi-toml/tombi-pre-commit
+  rev: v0.9.17
   hooks:
-    - id: taplo-format
-      name: taplo-format
-      description: Format TOML documents
-      entry: taplo format
-      language: python
-      types: [toml]
-      args: []
-      require_serial: false
-      additional_dependencies: ["taplo==0.9.3"]
-    - id: taplo-lint
-      name: taplo-lint
-      description: Lint TOML documents
-      entry: taplo lint
-      language: python
-      types: [toml]
-      args: [--default-schema-catalogs]
-      require_serial: false
-      additional_dependencies: ["taplo==0.9.3"]
+    - id: tombi-format
+    - id: tombi-lint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,3 +62,24 @@ repos:
   rev: "v0.24.1"
   hooks:
     - id: validate-pyproject
+
+- repo: local
+  hooks:
+    - id: taplo-format
+      name: taplo-format
+      description: Format TOML documents
+      entry: taplo format
+      language: python
+      types: [toml]
+      args: []
+      require_serial: false
+      additional_dependencies: ["taplo==0.9.3"]
+    - id: taplo-lint
+      name: taplo-lint
+      description: Lint TOML documents
+      entry: taplo lint
+      language: python
+      types: [toml]
+      args: [--default-schema-catalogs]
+      require_serial: false
+      additional_dependencies: ["taplo==0.9.3"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,9 +10,7 @@ classifiers = [
   'Programming Language :: Python',
   'Programming Language :: Python :: 3',
 ]
-dynamic = [
-  'version',
-]
+dynamic = ['version']
 dependencies = [
   "asdf-standard>=1.1.0",
   "importlib-metadata>=4.11.4 ; python_version<='3.11'",
@@ -26,16 +24,9 @@ dependencies = [
   # end of vendorized jsonschema deps
 ]
 [project.optional-dependencies]
-all = [
-  "asdf[lz4]",
-  "asdf[http]",
-]
-lz4 = [
-  "lz4>=0.10",
-]
-http = [
-  "fsspec[http]>=2022.8.2",
-]
+all = ["asdf[lz4]", "asdf[http]"]
+lz4 = ["lz4>=0.10"]
+http = ["fsspec[http]>=2022.8.2"]
 docs = [
   "sphinx-asdf>=0.2.2",
   "graphviz",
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -58,7 +58,23 @@ repos:
   hooks:
     - id: blacken-docs
 
-- repo: https://github.com/abravalheri/validate-pyproject
```
braingram, line 61: Why remove this hook? Are the various PEP compliance checks being handled elsewhere? I assume it may be `default-schema-catalogs` but trying to look that up doesn't instill confidence.

.pre-commit-config.yaml
```diff
@@ -58,7 +58,23 @@ repos:
...
-    - id: validate-pyproject
+    - id: taplo-format
+      name: taplo-format
+      description: Format TOML documents
+      entry: taplo format
+      language: python
+      types: [toml]
+      args: []
+      require_serial: false
+      additional_dependencies: ["taplo==0.9.3"]
```
braingram: Testing this by setting the pin back to `0.9.0` and running `pre-commit autoupdate` pre-commit doesn't update this pin. The third-party hook should be compatible so I'd say if we go for taplo we go that route so we don't have to manually update this pin.

braingram (commented):
Thanks. What is the state of taplo? I ask because while testing if `pre-commit autoupdate` would work I noticed the latest release on pypi 0.9.3 is from Aug 2024. It appears the original maintainer is no longer maintaining it https://github.com/tamasfe/taplo/issues/715 and projects are moving away from using it. Also while looking up `default-schema-catalogs` the docs point to a dead link: https://www.schemastore.org/json/ so I'm not sure what "validate" is doing here.

One of the projects that moved away from taplo mentions https://github.com/tombi-toml/tombi is that a suitable alternative?

sydduckworth:
Taplo is getting occasional updates (at least to the VSCode plugin) and seems more stable than Tombi (e.g. 4M downloads for the Tombi VSCode extension vs 10k for Taplo). 

Testing Tombi it seems to produce the same results in terms of both formatting and linting as Taplo, but it also has a first-party precommit hook so that seems like a good option.

.pre-commit-config.yaml
sydduckworth: Taplo does handle validation via JSON schema. The docs link is dead but it is actually using the correct link internally (plus I will be switching it over to Tombi anyway)

.pre-commit-config.yaml
sydduckworth: Done. Switched to Tombi and their first-party hook.

pyproject.toml
```diff
@@ -1,81 +1,64 @@
...
-  "asdf[tests]",
-]
-benchmark = [
-  "asdf[tests]",
-  "pytest-benchmark",
-]
+http = ["fsspec[http]>=2022.8.2"]
+lz4 = ["lz4>=0.10"]
+test = ["asdf[tests]"]
+tests = ["asdf[all]", "psutil", "pytest>=8"]
```
braingram:
```suggestion
tests = ["asdf[all]", "psutil", "pytest>=8"]
test = ["asdf[tests]"]
```
Since `test` is essentially an alias of `tests` I think this ordering makes more sense. How do we stop this from being reordered?

pyproject.toml
```diff
@@ -1,81 +1,64 @@
...
-dynamic = [
-  'version',
+  "Development Status :: 5 - Production/Stable",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3",
 ]
 dependencies = [
   "asdf-standard>=1.1.0",
+  # for vendorized jsonschema
+  "attrs>=22.2.0",
```
braingram: How do we prevent this from being reordered? It now no longer aligns with the `# end` comment.

braingram (commented):
It looks like tombi is reordering things. Can we globally disable this? It does support inline comments to disable sorting but so far I've only been able to get them to apply to one item so we'd have to either:
- add auto-sorting disable to each item we want to disable
- define a new custom schema

neither seems very helpful here.

.pre-commit-config.yaml
braingram: It looks like validate-pyproject uses the same schema and adds some checks that aren't easy to (or possible to) put in the schema. Are some of these checks not valid or is there some other issue with validate-pyproject that's leading us to turn it off?

sydduckworth: I thought it made sense to remove so that we avoid accumulating several tools that all do mostly the same thing, which will be harder to maintain. I also noticed that tombi/taplo caught several issues with the pyproject schema that `validate-pyproject` didn't catch.

I'm fine with keeping `validate-pyproject` if we want for the extra validation its doing.

Can I remove the `check-toml` hook? I have to imagine its not doing anything the other 2 hooks aren't already doing.

pyproject.toml
sydduckworth: Yeah it seems like there's no way to prevent reordering other than on a case-by-case basis ([relevant issue](https://github.com/tombi-toml/tombi/discussions/1704)).
Lack of configurability seems to be a common complaint about Tombi. In the Taplo github issues several people cite it as the reason they haven't switched over. 

We can just switch back to Taplo? It doesn't seem to be getting new features but there are at least maintenance updates happening and the actual schema validation references [SchemaStore](https://www.schemastore.org/) which is updated independently.

sydduckworth: Actually, looks like Tombi does respect blank lines so that could work.

I went ahead and updated `pyproject.toml` to restore the ordering for the two tables you mentioned. Let me know if that works for you as a solution.

pyproject.toml
```diff
@@ -21,61 +18,49 @@ dependencies = [
...
+'documentation' = "https://asdf.readthedocs.io/en/stable/"
+'repository' = "https://github.com/asdf-format/asdf"
+'tracker' = "https://github.com/asdf-format/asdf/issues"
+
 [project.entry-points]
-'asdf.extensions' = {asdf = 'asdf._core._integration:get_extensions'}
-'asdf.resource_mappings' = {asdf = 'asdf._core._integration:get_json_schema_resource_mappings'}
-asdf_extensions = {builtin = 'asdf.extension._legacy:BuiltinExtension'}
-console_scripts = {asdftool = 'asdf._commands.main:main'}
+asdf_extensions = { builtin = "asdf.extension._legacy:BuiltinExtension" }
```
braingram, line 50: Just a note that I think we can remove this (in a different PR).

braingram (approved):
Thanks for working through this. I can't say I'm entirely a fan of the new formatting but it's certainly an improvement to have something enforced.

PR #2027, opened 2026-04-16
Fixed checksum calculation for compressed blocks

sydduckworth:
## Description

* Updated block write behavior so that the checksum in the block header is computed after the block buffer is compressed instead of before.
* Updated block read behavior so that for compressed blocks the checksum is computed before the block is decompressed. This adds a small amount of overhead for compressed blocks when `validate_checksums` is enabled because the raw block is read into a temporary buffer before being decompressed. No additional overhead if `validate_checksums` is off, blocks are not compressed, or blocks do not contain checksums.

Addresses part of #2015


braingram:
Thanks! Overall this looks great. I added the `downstream` label to run some extra tests in the CI.

Would you also queue up jwst and romancal regression tests and link them here: https://github.com/spacetelescope/RegressionTests/blob/main/docs/running_regression_tests.md

braingram:
Would you add a "warning" block to the docs under the "Compression" section:
https://www.asdf-format.org/projects/asdf/en/latest/asdf/arrays.html#compression

that describes that writing out compressed blocks will fail checksum validation in older (<=5.2.0) versions of asdf. Please include that asdf doesn't validate checksums by default and that disabling checksum validation will allow opening the new files in older versions of asdf and whatever else you think is important.

sydduckworth:
Regression tests:
* [romancal](https://github.com/spacetelescope/RegressionTests/actions/runs/24572423684/job/71848839649)
* [jwst](https://github.com/spacetelescope/RegressionTests/actions/runs/24573080576/job/71851171321)

sydduckworth:
@braingram I added a warning to the docs!

[Link](https://asdf-website--2027.org.readthedocs.build/projects/asdf/en/2027/asdf/arrays.html#compression)

docs/asdf/arrays.rst
```diff
@@ -240,10 +240,22 @@ Compression
 
 Individual blocks in an ASDF file may be compressed.
 
+.. warning::
+
+    Files created by ``asdf`` versions prior to ``v5.2.1`` used an incorrect method of computing
+    checksums for compressed blocks. This bug was fixed in ``v5.2.1``.
+    As a result files created by ``asdf<=v5.2.0`` are not fully compatible with ``asdf>=v5.2.1`` and vice-versa.
```
braingram:
```suggestion
    Files created by ``asdf`` versions prior to ``5.2.1`` used an incorrect method of computing
    checksums for compressed blocks. This bug was fixed in ``5.2.1``.
    As a result files created by ``asdf<=5.2.0`` will fail block checksum validation with ``asdf>=5.2.1`` and vice-versa.
```
Thanks for adding this section. This suggestion removes the `vX.X.X` version formatting in favor of the `X.X.X` format we use elsewhere in the docs and clarifies the incompatibility.

asdf/_block/io.py
```diff
@@ -11,6 +11,7 @@
 
 import yaml
 
+import asdf.generic_io as gio
```
braingram:
```suggestion
from asdf.generic_io import get_file
```
Let's either import what is needed or import this as `generic_io`. Using `gio` makes operations like grepping for `generic_io` harder and misleading.

asdf/_block/io.py
```diff
@@ -317,6 +354,13 @@ def generate_write_header(data, stream=False, compression_kwargs=None, padding=F
         padding = util.calculate_padding(used_size, padding, fs_block_size)
         header_kwargs["allocated_size"] = header_kwargs.get("allocated_size", used_size + padding)
 
+    if stream:
+        header_kwargs["checksum"] = b"\0" * 16
+    elif buff is not None:
+        header_kwargs["checksum"] = calculate_block_checksum(buff.getvalue())
```
braingram:
```suggestion
        header_kwargs["checksum"] = calculate_block_checksum(buff.getbuffer())
```
Thanks for running the regression tests. The romancal regtests show some tests are consuming more memory with this PR (showing ~5 to 25% increases in memory). I think this is due to `getvalue` returning a copy. Can we use `getbuffer` here? It's compatible with the buffer API (so it should work with hashlib).

braingram (commented):
Thanks for making this fix. I left a few comments and suggestions. If the `getvalue` -> `getbuffer` change doesn't undo the memory regression for romancal we may need to consider other options for addressing that.

asdf/_block/io.py
sydduckworth: Sure, changed. I assumed that was a standard practice for IO modules since `asdf._block.io` is imported as `bio` in a number of places.

docs/asdf/arrays.rst
sydduckworth: Updated

asdf/_block/io.py
```diff
@@ -317,6 +353,13 @@ def generate_write_header(data, stream=False, compression_kwargs=None, padding=F
...
```
braingram:
```suggestion
        header_kwargs["checksum"] = calculate_block_checksum(buff.getbuffer())
```
Can we use `getbuffer` here as well?

braingram (commented):
One more question about getvalue/getbuffer.

asdf/_block/io.py
sydduckworth: Sure, changed. Even without that change the increase in memory usage in the regression tests seems to have been resolved:

* [romancal](https://github.com/spacetelescope/RegressionTests/actions/runs/24688799008)
* [jwst](https://github.com/spacetelescope/RegressionTests/actions/runs/24688819445)

braingram (approved):
Thanks. Let's make a release once this is merged.

PR #2028, opened 2026-04-22
Add type checking using Pyrefly

sydduckworth:
## Description

* Added Pyrefly configuration to `pyproject.toml`
* Added Pyrefly [baseline file](https://pyrefly.org/en/docs/error-suppressions/#baseline-files-experimental) as `.pyrefly-baseline.json` to suppress existing typing errors
* Updated Ruff config to enable `TC` type-checking lints and `future-annotations` option
* Added type-checking job to Github actions


braingram:
Can we add this as a pre-commit hook?

zacharyburnett:
> Can we add this as a pre-commit hook?

_should_ this be a pre-commit hook? Having it as a github CI is great, and the user can also install pyrefly and have the language server running with their editor if they choose, but maybe it's too heavy for pre-commit?

.github/workflows/ci.yml
```diff
@@ -33,6 +33,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v6
+      with:
+        # Run type checking against the oldest supported version
+        python-version: "3.10"
+    - uses: astral-sh/setup-uv@v7
+      with:
+        enable-cache: true
```
zacharyburnett: why install both? I don't think `setup-python` is needed here because `uv run` will handle the python install right? Or does uv pick up the system python?

sydduckworth:
My opinion is that type-checking should be treated like the test suite and shouldn't go into precommit.  
There are also some practical problems with adding it to precommit:

Precommit runs hooks in an isolated environment, but type checkers need to import project dependencies. 
You can configure the type checker to ignore unresolved dependencies, but then the result you get from precommit no longer matches the result from CI.  

Alternatively you can have precommit use the system version of the type-checker but then you're requiring users to not only have the tool installed but also have it available in the shell where they're running `git commit`. This would break VSCode's integrated Git support and probably most editors.  
This also breaks my workflow because I use `uv` for project management so I don't have my virtualenv sourced in my shell, I just use `uv run`, which precommit obviously can't detect.

.github/workflows/ci.yml
sydduckworth: Not strictly needed but as mentioned in the [uv docs](https://docs.astral.sh/uv/guides/integration/github/#setting-up-python) doing it this way is faster because it allows Github to cache the Python version.

asdf/_compression.py
```diff
@@ -115,7 +116,11 @@ def decompress(self, blocks, out, **kwargs):
                         )  # use numpy instead of bytearray so we can avoid zero initialization
                         _pos = 0
                     newbytes = min(_size - _pos, len(blk))  # don't fill past the buffer len!
-                    _buffer[_pos : _pos + newbytes] = np.frombuffer(blk[:newbytes], dtype=np.byte)
+
+                    # Older versions of python seem to treat memoryview specializations weirdly
+                    # So cast memoryview[bytes] to memoryview[any] so numpy doesn't complain
+                    blk_buf: memoryview[Any] = blk[:newbytes]
+                    _buffer[_pos : _pos + newbytes] = np.frombuffer(blk_buf, dtype=np.byte)
```
braingram:
```suggestion
                    _buffer[_pos : _pos + newbytes] = np.frombuffer(blk[:newbytes], dtype=np.byte)  # type: ignore
```
Changing code to handle weird type hints seems backwards to me.

asdf/_tests/test_array_blocks.py
```diff
@@ -926,7 +929,7 @@ def test_remove_blocks(tmp_path, lazy_load, memmap):
     fn1 = tmp_path / "test.asdf"
     fn2 = tmp_path / "test2.asdf"
 
-    tree = {"a": np.zeros(3), "b": np.ones(10)}
+    tree: dict[str, Any] = {"a": np.zeros(3), "b": np.ones(10)}
```
braingram: What benefit does this type hint provide?

.github/workflows/ci.yml
```diff
@@ -33,6 +33,19 @@ jobs:
...
     - uses: pre-commit/action@v3.0.1
+  type-checking:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v6
+      with:
+        # Run type checking against the oldest supported version
+        python-version: "3.10"
+    - uses: astral-sh/setup-uv@v7
```
braingram: This is a dead-end version that will never be updated. Would you update this to remove `uv`?

asdf/_compression.py
sydduckworth: Would you prefer `typing.cast`?

braingram: Slightly? Since there is nothing wrong with the code here both solutions make it:
- less readable
- more complex

for no benefit (please let me know if I'm missing something here). I'd prefer `# type: ignore`.

asdf/_tests/test_array_blocks.py
sydduckworth: It widens the value type to allow the assignment on the next line to succeed.  
In newer versions of Python/numpy its able to infer that slicing into a 1D array produces a 1D array but in older versions it seems like it assumes that slicing into an array produces an array with arbitrary dimensions.

I could have made the value type more specific but wading into the depths of the numpy type system didn't seem worthwhile for something only used in a test case.

asdf/_compression.py
sydduckworth: Sure changed.

asdf/_tests/test_array_blocks.py
braingram: Thanks. I should have been more specific. I understand what this does for the type checker but how does it provide benefit to us or our users? Since this is a local variable inside a test function it doesn't provide any benefit to users. For developers I also don't see how it provides any benefit.

I'm also surprised that pyrefly is checking this line since the function is untyped but I guess that's because by default it will check untyped functions due to [check-unannoted-defs](https://pyrefly.org/en/docs/configuration/#check-unannotated-defs) being enabled. It seems preferable at this point (given that there are no type hints) to start with that disabled. To test this out locally I:
- disabled check-unannotated-defs
- removed the baseline file
- ran pyrefly check

The errors are of 2 types:
- invalid-inheritance, inconsistent-inheritance for a number of classes: TaggedDict, TaggedList, TaggedStr, AsdfObject, the various loaders and dumpers in yamlutil
- 2 errors for the module patching in asdf.util

We could address these by:
- adding ignore comments for the inheritance errors (and opening a follow-up issue on addressing them)
- the errors for asdf.util are for not handling possible `None` if `urllib.parse` isn't found. That's a pretty terminal error if a module from the standard library is missing so I'd be ok with either ignoring the errors, updating the code to handle `None` (and adding a test where `urllib.parse` is missing) or some other solution.

What do you think about that approach instead of using the baseline file?

asdf/_tests/test_array_blocks.py
```diff
@@ -938,7 +941,7 @@ def test_remove_blocks(tmp_path, lazy_load, memmap):
...
```
braingram:
```suggestion
    tree: dict = {"a": np.zeros(3), "b": np.ones(10)}
```

braingram (approved):
LGTM. One nitpick on the type hint in the test. We should update the branch protections once this is merged.

PR #2030, opened 2026-04-23
Prepare release 5.3.0

sydduckworth:


braingram:
Can we squeeze https://github.com/asdf-format/asdf/pull/2029 in?

CHANGES.rst
```diff
@@ -1,3 +1,28 @@
+5.2.1.dev62+g416ac124e.d20260423 (2026-04-23)
+=============================================
```
braingram:
```suggestion
5.3.0 (2026-04-23)
===============
```
towncrier needs an explicit version when building the changelog. I may not have the correct number of `=`.

braingram:
Thanks! Updates look good to me. I queued up regtest runs to double check.

roman: https://github.com/spacetelescope/RegressionTests/actions/runs/24851073305/job/72751370005
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/24851084371/job/72751409941

We should consider updating the wiki since I typically run the regtests at this point (so if something comes up we don't have to possibly undo the changelog updates).

braingram (approved):
Regtests passed. Looks like the CI had some hiccups but no objection from me if you want to force merge this one given all the successful other testing.

PR #2031, opened 2026-04-24
Added type hints to `_asdf.py` and `config.py`

sydduckworth:
## Changes
* Added type hints to all methods and functions in `_asdf.py` and `config.py`. Added type hints to other modules as needed to support those two files.
* No user-facing changes other than the addition of typing (and I guess the change to `NotSet` described below)
* A small number of code refactors that shouldn't change functionality:
  * Defragment command now checks if the output file is `None` before writing to it
  * In `get_file` added an `isinstance` check to the special case where the input is stdin/stdout/stderr so that type narrowing works correctly
  * In `get_file` updated `RandomAccessFile` and subclasses so that the reference to their original file handle is passed in the constructor instead of being set separately
  * Removed some import logic in `jsonschema` which only existed to support Python versions < 3.9
* Reworked the `NotSet` constant in a way that plays nicely with the Python type system
  * The underlying class has been reworked to use the same sentinel value pattern as `dataclasses.MISSING`
  * The `NotSet` constant has been renamed to `NOT_SET` because apparently a global being in all-caps makes type-checkers/linters assume it's immutable
  * Added `NotSet` as a type alias for the typing needed to use the `NOT_SET` constant.
  *  Usage in function signatures is `myparam: str | NotSet = NOT_SET`. Usage in function bodies is the same and now doing `value is not NOT_SET` correctly applies type narrowing in the type-checker.
* Added `typing.py` module which contains new type aliases and protocols
* Added some type-hints to tests where type inference failed but otherwise tried to avoid annotating tests. For tests specifically testing incorrect arguments added `# pyrefly: ignore`

### `TreeKey`
I added a `TreeKey` type alias in `asdf.typing` to be used for typing dictionaries that can be converted into ASDF trees. 
The original intention was that it would be `str | int | bool` or similar, but it turns out that 
1. For various reasons Python decided that although map *values* are covariant, map *keys* are invariant
2. The specific types in a union are considered subtypes of the union. 

The end result is that you wouldn't be able to assign a `dict[str, Any]` to a `dict[TreeKey, Any]` field without manually widening the `str` type first.

So for now I've just set `TreeKey = Any`, with the idea that if we find a way in the future to narrow the type without breaking everything we can update `TreeKey` and not have to update every individual place its used.

### `Compression`

I added a type alias `ArrayStorage` which aliases to the different string literal values that are valid for `array_storage`.   This has the nice effect that the Python LSP will auto-suggest the valid options when you're typing the value.

I added another type alias `Compression` which was intended to do the same thing for the supported options for `array_compression`.
The complication is that array compression can be one of the predefined strings or it can also be any arbitrary string supported by an extension. 
So I set `Compression = Literal["lz4", ...] | str | None`, which works with the type-checker, but it looks like adding `str` to the union means the LSP won't auto-suggest the valid options.
I think the `Literal | str` union is still worth keeping rather than just `str` since it provides more information to the user and maybe in the future we can find a way to get it to work with the LSP.

### Documentation
* I added `asdf.typing` to the User API section, but it might make more sense in the developer API section. Then again, with the addition of type hints we might want to rethink how we organize the API docs in the future anyway.
* In a lot of places the type-hints contradict the types in the associated function docstring. We should consider removing types from the docstrings of functions that have type hints.
* There are a lot of issues I ran into getting Sphinx to work with type hints that I will put in a separate issue.

----

Sorry about the size of this PR 😅


asdf/_commands/defragment.py
```diff
@@ -74,4 +74,5 @@ def defragment(input_, output=None, resolve_references=False, compress=None):
         ff2 = AsdfFile(ff)
         if resolve_references:
             ff2.resolve_references()
-        ff2.write_to(output, all_array_storage="internal", all_array_compression=compress)
+        if output is not None:
```
zacharyburnett, line 77: what happened here before if `output is None`?

asdf/_tests/test_array_blocks.py
```diff
@@ -10,6 +11,7 @@
 from numpy.testing import assert_array_equal
 
 import asdf
+import asdf.versioning
```
zacharyburnett, line 14: what is this for?

asdf/_asdf.py
```diff
@@ -452,7 +482,7 @@ def close(self):
         self._external_asdf_by_uri.clear()
         self._blocks.close()
 
-    def copy(self):
+    def copy(self) -> AsdfFile:
```
zacharyburnett, line 485: doesn't this need to be quoted if `AsdfFile` is not yet defined? I guess if runs fine then it's not an issue

asdf/_io.py
```diff
@@ -84,7 +89,11 @@ def read_comment_section(fd):
     return comments
 
 
-def find_asdf_version_in_comments(comments, default=None):
+# Required for type narrowing in `find_asdf_version_in_comments``
+MaybeVer = TypeVar("MaybeVer", AsdfVersion, None)
```
zacharyburnett, line 93: fancy! Why does this need to be a `TypeVar` instead of `AsdfVersion | None`?

asdf/generic_io.py
```diff
@@ -210,7 +219,7 @@ def __init__(self, fd, mode, close=False, uri=None):
             (read/write).
 
         close : bool, optional
-            When ``True``, close the given `fd` in the ``__exit__``
+            When ``True``, close the given ``fd`` in the ``__exit__``
```
zacharyburnett, line 222: can these be split into another PR? I assume they are from Sphinx warnings?

asdf/typing.py
```diff
@@ -0,0 +1,82 @@
...
+# Alternate version of `Extension` for use in type-hints
+# The way `Extension` works is weird enough that it can't be replaced in actual code without a lot of changes
+class ExtensionLike(Protocol):
+    """Object that contains an extension URI and can be wrapped by ``ExtensionProxy``."""
+
+    @property
+    def extension_uri(self) -> str | None: ...
+
+
+# These protocols were added to the standard library in 3.14
```
zacharyburnett: does this mean type checking and dev requires 3.14?

asdf/typing.py
```diff
@@ -0,0 +1,82 @@
...
+# fails when you try to actually index the map.
+# Hopefully someday we will find a way to specialize this type more.
+
+#: Valid ASDF tree keys
+TreeKey: TypeAlias = Any
+
+#: Local file path or remote file URI
+PathLike: TypeAlias = str | Path
+#: Readable/writable file object or the path or URI of an openable file
+FileLike: TypeAlias = PathLike | Reader | Writer | GenericFile
```
zacharyburnett, line 66: I'm not too sure on the differences in usage between `str | Path` and `PathLike`; why are they used differently here? Does `FileLike` match to a `str`?

changes/2031.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added type hints to `AsdfFile` and `asdf.config.AsdfConfig`.
+Added `asdf.typing` module containing new type aliases.
```
zacharyburnett, line 2: can you also add something about `NOT_SET`? Or is that not user-facing

asdf/typing.py
```diff
@@ -0,0 +1,82 @@
+"""This module contains type aliases and protocols."""
```
zacharyburnett:
```suggestion
"""This module contains type aliases and protocols useful for type checking."""
```

zacharyburnett:
this looks pretty comprehensive to me, I didn't know about PEP647 and type narrowing until just now 😅

asdf/_commands/defragment.py
sydduckworth: From my understanding it would raise an exception when trying to call `generic_io.get_file(output)` further down the callstack.
I considered just raising an exception here but I wasn't sure if `resolve_references()` could have some side effect that would be relevant without an output.

Since `defragment` is marked a deprecated it didn't seem super critical either way.

asdf/_tests/test_array_blocks.py
sydduckworth: There are a number of places in the code where modules are being used without actually being imported, which works because the module is already being imported by a different module.

This is just adding an explicit import for one of those cases.

asdf/_asdf.py
sydduckworth: The `from __future__ import annotations` statement switches to lazy type hint resolution (which is default in 3.14+). 
This prevents circular import issues by allowing types only used in annotations to be imported in an `if TYPE_CHECKING:` block and also allows referencing types defined later in the same file without needing quotes.

asdf/_io.py
sydduckworth: Basically if the variable is typed as `AsdfVersion | None` then even if the variable happens to contain `None` the type checker considers it valid for it to potentially contain `AsdfVersion` at any point in the future.
Using `TypeVar` causes the variable to be resolved to the specific type, meaning if the argument is `None` its able to infer that the output will also be `None` and vice-versa.

asdf/generic_io.py
sydduckworth: I needed to have Sphinx generate API docs for `generic_io` since `GenericFile` is used in existing parts of the API. 
So I needed to either resolve the new Sphinx warnings or suppress them, and since resolving them was trivial it seemed like the better option.

asdf/typing.py
sydduckworth: No, I can maybe clarify that comment. I meant that those protocols become part of the standard library in 3.14 so if we ever drop support for <3.14 then we could remove these types and use the standard library versions instead.

I try to note when I'm basically manually backporting a feature from the standard library.

asdf/typing.py
sydduckworth: `PathLike` is just a type alias for `str | Path` so they're identical from the type checker perspective.
The idea was that `PathLike` is easier to read and has more semantic meaning than `str | Path` (and in the future may grow to contain other types if needed)

zacharyburnett: oh now I see, you're not using `os.PathLike` you're defining your own, sorry I had banner blindness

zacharyburnett (approved):
LGTM especially as there are no functional differences

asdf/util.py
```diff
@@ -315,16 +317,22 @@ def is_public_member(key):
...
-from None, which may for example be a value of interest in a search.
-"""
-NotSet = _NotSetType()
+
+# This needs to be all caps because otherwise type-checkers will assume its mutable
+# https://github.com/facebook/pyrefly/issues/3042
+
+#: Special value indicating that a parameter is not set.
+#: Distinct from None, which may for example be a value of interest in a search.
+NOT_SET: Final = _NOT_SET_TYPE.NOT_SET
```
braingram, line 332: We can yearn for the day we have 3.15 as a min: https://peps.python.org/pep-0661/

braingram (approved):
Thanks for putting this together and the detailed write-up. I left one comment about NOT_SET but it's not actionable at the moment. I didn't scrutinize all the hints and don't think that's necessary at this point (and with zach's review, thanks!).

PR #2033, opened 2026-04-28
Switch Tombi precommit hooks to offline mode 

### LINKED ISSUE #2032: Tombi precommit hooks currently broken
The `precommit` CI job is currently failing for any commit that triggers Tombi to run ([example](https://github.com/asdf-format/asdf/actions/runs/25064917642/job/73445966977?pr=2031)).
It appears that jsonschema.org is having server-side issues ([relevant issue](https://github.com/SchemaStore/schemastore/issues/5621)) which is causing Tombi to fail to retrieve JSON schemas. This problem isn't necessarily manifesting locally because Tombi caches fetched schemas so if you've run it locally already it will just use those.

For now I think the best option is to switch the Tombi hooks to only using the vendored schema store maintained in the Tombi repo by passing the `--offline` flag. Currently Tombi is the only hook that requires internet access to run and this will allow it to run offline like all the other hooks.

sydduckworth:
## Description
* Added `--offline` flag to both Tombi pre-commit hooks so they don't require internet access to run
* Updated Tombi hooks to use latest version `v0.9.25`
* Applied updated TOML formatting to `pyproject.toml` (See my comment [here](https://github.com/asdf-format/asdf/issues/2032#issuecomment-4338353634))

Closes #2032

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,19 +25,6 @@ dependencies = [
 ]
 dynamic = ["version"]
 
-[project.urls]
-'documentation' = "https://asdf.readthedocs.io/en/stable/"
-'repository' = "https://github.com/asdf-format/asdf"
-'tracker' = "https://github.com/asdf-format/asdf/issues"
-
-[project.scripts]
-asdftool = "asdf._commands.main:main"
-
-[project.entry-points]
-asdf_extensions = { builtin = "asdf.extension._legacy:BuiltinExtension" }
-'asdf.extensions' = { asdf = "asdf._core._integration:get_extensions" }
-'asdf.resource_mappings' = { asdf = "asdf._core._integration:get_json_schema_resource_mappings" }
-
 [project.optional-dependencies]
 all = ["asdf[lz4]", "asdf[http]"]
 docs = [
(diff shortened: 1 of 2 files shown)


zacharyburnett:
Good catch, I'll add this to the template PR too

zacharyburnett (approved):

pyproject.toml
```diff
@@ -42,35 +55,10 @@ typing = ["pyrefly==0.62.0", "types-PyYAML", "types-jmespath"]
...
-'asdf.resource_mappings' = { asdf = "asdf._core._integration:get_json_schema_resource_mappings" }
-
-[project.scripts]
-asdftool = "asdf._commands.main:main"
-
 [build-system]
 requires = ["setuptools>=60", "setuptools_scm[toml]>=8", "wheel"]
 build-backend = "setuptools.build_meta"
 
-[tool.setuptools.packages.find]
```
braingram, line 62: Having the setuptools arguments after the build system made more sense to me than the new order.

I'd be in favor of disabling the reordering since the order has no impact on the use of this file but the reordering is making it less readable.

sydduckworth: I've reverted the formatting and disabled table sorting in `pyproject.toml`. Although I will point out that having the tables sorted makes them easier to find if you're not already familiar with the existing order!

Issue #1782, opened 2024-04-09
`AsdfFile` instances are not pickleable


braingram:
# Description of the problem

`AsdfFile` instances are not pickleable.

# Example of the problem

```python
import asdf, pickle

af = asdf.AsdfFile({'a': 1})
pickle.dumps(af)
```
Fails with
```
AttributeError: Can't pickle local object 'ValidatorManager._get_jsonschema_validator.<locals>._validator'
```

# System information

asdf version: main
python version: 3.10
operating system: mac os (m1)

braingram:
To provide a bit more context. The lack of "pickle-ability" means `AsdfFile` instances do not play well with `multiprocessing`:

```python
def get_a(af):
    return af['a']

with multiprocessing.Pool(4):
    pool.map(get_a, [af])
```
Fails with the above error.

PR #2038, opened 2026-05-04
Fix `AsdfFile` pickling for instances without file descriptors

sydduckworth:
## Description

This PR fixes two bugs that prevented `AsdfFile` instances from being pickled, which partially addresses #1782.
The result is that now asdf files can be pickled **if** they don't contain a file descriptor, which basically means files that were created from an in-memory dict.

* Fixed a bug in which `AsdfObject` instances were failing to deserialize because of their conflicting `dict` and `UserDict` base classes. Added a manual `__reduce__` override which resolves the problem.
* Updated `ValidatorManager` to not return local callables defined inside a method, which can't be pickled.
  * `ValidatorManager` now returns a new `JsonSchemaValidators` callable class 
  * Significantly simplified `ValidatorManager` implementation.
* Added test case to `_tests/test_asdf.py` to verify the limited pickling that is now supported.

### DIFF
--- a/asdf/extension/_manager.py
+++ b/asdf/extension/_manager.py
@@ -354,8 +354,35 @@ class ValidatorManager:
         }
 
     @property
-    def validators(self) -> dict[str, BoundValidators]:
+    def bound_validators(self) -> dict[str, BoundValidators]:
         """Dictionary mapping schema names to callable validator functions."""
+        return self._validators
+
+    def validate(
+        self, schema_property: str, schema_property_value: Any, node: Tagged, schema: Mapping[TreeKey, Any]
+    ) -> Iterator[ValidationError]:
+        """Validate an ASDF tree node against custom validators for a schema property.
+
+        Parameters
+        ----------
+        schema_property : str
+            Name of the schema property (identifies the validator(s) to use).
+        schema_property_value : object
+            Value of the schema property.
+        node : asdf.tagged.Tagged
+            The ASDF node to validate.
+        schema : dict
+            The schema object that contains the property that triggered
+            the validation.
+
+        Yields
+        ------
+        asdf.exceptions.ValidationError
+        """
+        yield from self.bound_validators[schema_property](None, schema_property_value, node, schema)
+
+    def get_jsonschema_validators(self) -> dict[str, BoundValidators]:
+        """Get a dictionary mapping schema names to callable validator functions."""
         return self._validators
 
 
--- a/asdf/extension/_manager.py
+++ b/asdf/extension/_manager.py
@@ -353,8 +353,9 @@ class ValidatorManager:
             for schema_property, validators in by_schema_property.items()
         }
 
+    @property
     def validators(self) -> dict[str, BoundValidators]:
-        """Get a dictionary mapping schema names to callable validator functions."""
+        """Dictionary mapping schema names to callable validator functions."""
         return self._validators
 
 
--- a/asdf/extension/_manager.py
+++ b/asdf/extension/_manager.py
@@ -1,10 +1,22 @@
+from __future__ import annotations
+
 import sys
+from dataclasses import dataclass
 from functools import lru_cache
+from typing import TYPE_CHECKING
 
 from asdf.tagged import Tagged
 from asdf.util import get_class_name, uri_match
 
 from ._extension import ExtensionProxy
+
+if TYPE_CHECKING:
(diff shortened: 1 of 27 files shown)


asdf/extension/_manager.py
braingram, line 1: How do the changes in this file relate to this PR? If unrelated would you split them into a different PR?

asdf/extension/_validator.py
braingram, line 1: Similar to https://github.com/asdf-format/asdf/pull/2038/changes#r3184559350 if these are unrelated would you split them into a different PR? If related would you help me understand how?

asdf/schema.py
braingram, line 1: Similar to https://github.com/asdf-format/asdf/pull/2038/changes#r3184559350

asdf/extension/_manager.py
sydduckworth: Changing `get_jsonschema_validators` was necessary to fix the pickling issue.
Because of that change, `validate` became both unused and untested, so I removed it.
Removing `validate` allowed me to clean up the constructor for `ValidatorManager`.

I renamed `get_jsonschema_validators` to just `validators` because it is now the only thing that `ValidatorManager` actually returns, and I found the `jsonschema` part confusing since it isn't really actually related to `jsonschema` at all (or any more than the input validators are).

asdf/extension/_validator.py
sydduckworth: I added type hints here to support the changes I made to `ValidatorManager`. Since I was adding type hints to the code I wrote there I wanted to make sure it agreed with the input typing.
It didn't seem worth opening a new PR for a couple of type hints.

asdf/extension/_manager.py
```diff
@@ -317,70 +329,51 @@ def _get_cached_extension_manager(extensions):
...
                 schema_property,
+                frozenset(validators),
             )
+            for schema_property, validators in by_schema_property.items()
+        }
 
-    def validate(self, schema_property, schema_property_value, node, schema):
-        """
-        Validate an ASDF tree node against custom validators for a schema property.
+    def validators(self) -> dict[str, BoundValidators]:
```
zacharyburnett: would it make more sense / be more ergonomic for this to be a property?

sydduckworth: Yeah, probably. Fixed!

sydduckworth:
@braingram I've reverted the public API for `ValidatorManager`.

braingram:
I think we'd benefit from including a wider set of possible `AsdfFile` instances for initial pickling support.

Testing this PR if `validate` is called on an `AsdfFile` (created from an in-memory tree) prior to pickling the pickling fails with:
```python
import asdf, pickle, numpy as np
af = asdf.AsdfFile({"arr": np.arange(42)})
pickle.loads(pickle.dumps(af))  # passes
af.validate()
pickle.loads(pickle.dumps(af))  # fails
```
```
TypeError: cannot pickle 'weakref.ReferenceType' object
```
The same error appears if array storage type is set on "arr" prior to pickling.
```python
af.set_array_storage(af["arr"], "inline")
pickle.loads(pickle.dumps(af))  # fails
```
As part of this we should address what operations cause pickling to fail and why? What are the "typical" AsdfFile instances we want to support (probably at least a tree created from in-memory objects but also one read from a file and read then modified)?

I think we shouldn't claim pickling support until we support whatever we decide are the "typical" cases.

sydduckworth:
Is there a different changelog entry you would prefer for a PR that specifically fixes the two bugs addressed here but that does not fully enable pickling?

asdf/extension/_manager.py
```diff
@@ -317,32 +329,26 @@ def _get_cached_extension_manager(extensions):
...
-        self._jsonschema_validators_by_schema_property = {}
-        for schema_property in self._validators_by_schema_property:
-            self._jsonschema_validators_by_schema_property[schema_property] = self._get_jsonschema_validator(
-                schema_property,
-            )
-
-    def validate(self, schema_property, schema_property_value, node, schema):
-        """
-        Validate an ASDF tree node against custom validators for a schema property.
+    def validate(
```
braingram, line 348: Based on the current state and the previous comments should we deprecate this since it's now used (except for the new test)? That can be handled in a separate PR if that's easier.

braingram (approved):
Thanks! One minor question about if we deprecate `ValidatorManager.validate` but that can go in a separate PR if that's easier.

PR #2040, opened 2026-05-05
Add `typing-extensions` dependency

sydduckworth:
## Description
* Added `typing-extensions` as a dependency
* Removed custom `Reader` and `Writer` protocols in `asdf.typing` and replaced them with the backports from `typing_extensions`

`typing-extensions` backports various typing features so that they're available in older Python versions. 
In addition to letting us remove the custom protocols as described above, it also gives access to the `deprecated` decorator which was added to `warnings` in Python 3.14 and will be very useful as we start to deprecate parts of the API.

Worth noting that `typing-extensions` is already an indirect dependency of several optional dependencies.

### DIFF
--- a/asdf/typing.py
+++ b/asdf/typing.py
@@ -4,9 +4,10 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from pathlib import Path
-from typing import Any, AnyStr, Literal, Protocol, TypeAlias
+from typing import Any, Literal, Protocol, TypeAlias
 
 import numpy.typing as npt
+from typing_extensions import Reader, Writer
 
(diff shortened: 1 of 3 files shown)


zacharyburnett (approved):
LGTM, how does the `deprecated` decorator look to users? It's https://peps.python.org/pep-0702/ right?

zacharyburnett:
nvm found it https://docs.python.org/3.13/library/warnings.html#warnings.deprecated, looks neat!

braingram (approved):
Hopefully this wasn't waiting on my review. It would be good to get this in before the typing submodule is released.

PR #2041, opened 2026-05-05
Replace `pre-commit` with `prek`

### LINKED ISSUE #2034: Switch precommit CI job to prek
The [precommit action](https://github.com/pre-commit/action) is in maintenance-only mode.
The [prek action](https://github.com/j178/prek-action) is actively maintained and significantly faster than precommit.

sydduckworth:
## Description
* Replaced pre-commit Github action with [j178/prek-action](https://github.com/j178/prek-action)
* Updated contributor docs to suggest prek instead of pre-commit

Closes #2034 

I kept the name of the CI job as `pre-commit` to avoid breaking branch protections and such, plus I think the name is still fairly accurate.

There's also a "pre-commit enabled" badge on the readme. Not sure if we wanted to leave it as-is, update it, or remove it.

Let me know if you think this needs a changelog entry.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,13 +4,13 @@ Thanks for contributing to asdf!
 Your PR should trigger the CI (after approval for first-time contributors)
 which will:
 
-- check your code for 'style' using pre-commit
+- check your code for 'style' using prek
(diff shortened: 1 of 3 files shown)


braingram (approved):
Thanks!

PR #2044, opened 2026-05-07
fix add_history_entry validation to ignore custom schema

braingram:
## Description

This fixes a bug introduced by https://github.com/asdf-format/asdf/pull/2029 where if a user:
- provides a custom schema that constrains the entire tree (like is the case for [dkist](https://github.com/DKISTDC/dkist/blob/a7dc385cc0c30523a93fe1e14a4f4a926bfcda2d/dkist/io/level_1_dataset_schema.yaml#L6))
- calls add_history_entry

The call to `add_history_entry` can fail due to the validation:
- not including the entire tree
- (incorrectly) using the custom schema

This PR fixes the issue by providing a blank schema to `validate` (which will be used instead of the custom schema). The tagged entry will still be validated.

Hopefully we can get some confirmation from dkist that this fixes the issues they had with asdf 5.3.0. Unfortunately this was only revealed in tests for https://bitbucket.org/dkistdc/dkist-inventory/src which we did not have in our downstream. This PR also adds that to our downstream.

## Tasks

</details>

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -334,7 +334,7 @@ commands_pre =
     pip install -r {env_tmp_dir}/asdf_requirement.txt
     pip freeze
 commands =
-    pytest
+    pytest -n auto
 
 [testenv:abacusutils]
 change_dir = {env_tmp_dir}
--- a/tox.ini
+++ b/tox.ini
@@ -321,6 +321,21 @@ commands_pre =
 commands =
     pytest --benchmark-skip
 
+[testenv:dkist-inventory]
+change_dir = {env_tmp_dir}
+allowlist_externals =
+    git
+    bash
+extras =
+commands_pre =
+    git clone https://bitbucket.org/dkistdc/dkist-inventory.git .
+    bash -c "pip freeze -q | grep 'asdf @' > {env_tmp_dir}/asdf_requirement.txt"
+    pip install -e ".[test]"
+    pip install -r {env_tmp_dir}/asdf_requirement.txt
+    pip freeze
+commands =
+    pytest
+
 [testenv:abacusutils]
 change_dir = {env_tmp_dir}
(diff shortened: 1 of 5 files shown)


braingram:
From offline conversation with @Cadair this fixes the CI failures they are seeing with 3.5.0.

@sydduckworth any objection if we put this in a patch release (3.5.1)? Should we include the other changes on main?

tox.ini
```diff
@@ -321,6 +321,21 @@ commands_pre =
...
+    bash
+extras =
+commands_pre =
+    git clone https://bitbucket.org/dkistdc/dkist-inventory.git .
+    bash -c "pip freeze -q | grep 'asdf @' > {env_tmp_dir}/asdf_requirement.txt"
+    pip install -e ".[test]"
+    pip install -r {env_tmp_dir}/asdf_requirement.txt
+    pip freeze
+commands =
+    pytest
```
Cadair: You probably want to run this in parallel, they are quite slow?

braingram: Thanks. I see xdist is a dependency:
https://bitbucket.org/dkistdc/dkist-inventory/src/19054f04b721d9eea22767344986ff84d07122b9/pyproject.toml#lines-58
Would you recommend `-n auto`?

Cadair: Yeah probably

sydduckworth:
I'm fine with a patch release!

tox.ini
braingram: `-n auto` added in https://github.com/asdf-format/asdf/pull/2044/commits/79a673928f7cbb5607f5acd92763899285aec3a9

braingram: Worked! Thanks.

sydduckworth (approved):

PR #2048, opened 2026-05-12
Add type hints to block and IO APIs

sydduckworth:
## Description
* `asdf._block` and its submodules now have 100% type coverage. 
* `GenericFile` and its subclasses are mostly fully typed now.
* The advantage of adding this type coverage is now its possible to trace the precise usage of `generic_io` classes in the rest of the codebase, which will aid in future refactors.
* Added new type aliases to `asdf.typing`:
  * `ByteArray1D` as an alias for a 1-D uint8 numpy array which is used all over the place in block IO
  * `BlockDataCallback` as an alias for a callable that returns `ByteArray1D`, which is used for lazy block loading 
* Added a `BlockHeader` typed dictionary to describe the block header fields
* The tagged container types are now generic in the same way as their parent types, e.g. you can now annotate a value with `TaggedDict[str, int]` or `TaggedList[float]`
* Similarly, `Tagged` is now also generic over its inner type, which means type checkers will now recognize `TaggedList[str]` as a subtype of `Tagged[list[str]]`.
* Converted `asdf._block.external.UseInternal` to be a sentinel value that type checkers will understand: the actual value has been renamed to `USE_INTERNAL` and `UseInternal` is now a type alias for the value's type.

As with the previous typing PR there should be no functional changes and no changes to the public API (other than adding type hints and the change to `UseInternal`, which I'm not sure is public).

I made a few refactoring changes when necessary. Most of them are downstream of the fact that `ReadBlock` and `WriteBlock` can have an offset of `None` if the source file isn't seekable (or if the file being read was not seekable when written). There are a lot of places that assume that offset is an integer and would break if offset were ever actually `None`, so I had to add some checks in those places to handle the `None` case. 
In the future I think that we should just drop support for non-seekable files, which would allow offset to always be an integer.

### DIFF
--- a/asdf/_block/external.py
+++ b/asdf/_block/external.py
@@ -8,26 +8,42 @@ the block manager to have a reference to the `AsdfFile`
 (that references the block manager).
 """
 
+from __future__ import annotations
+
+import enum
 import os
 import urllib
+import urllib.request
+from typing import TYPE_CHECKING, Final, Literal
 
 import numpy as np
 
 from asdf import generic_io, util
 
+if TYPE_CHECKING:
+    from asdf.typing import ByteArray1D
 
-class UseInternalType:
-    pass
+
+class _USE_INTERNAL_TYPE(enum.Enum):
+    USE_INTERNAL = "UseInternal"
+
+    def __repr__(self) -> str:
+        return str(self.value)
 
 
-UseInternal = UseInternalType()
+USE_INTERNAL: Final = _USE_INTERNAL_TYPE.USE_INTERNAL
+
+#: Type corresponding to ``USE_INTERNAL`` value for use in type-checking
+UseInternal: Final = Literal[_USE_INTERNAL_TYPE.USE_INTERNAL]
 
 
 class ExternalBlockCache:
     def __init__(self):
         self.clear()
 
-    def load(self, base_uri, uri, memmap=False, validate_checksums=False):
+    def load(
+        self, base_uri: str | None, uri: str, memmap: bool = False, validate_checksums: bool = False
+    ) -> ByteArray1D | UseInternal:
         key = util.get_base_uri(uri)
         if key not in self._cache:
             resolved_uri = generic_io.resolve_uri(base_uri, uri)
@@ -37,7 +53,7 @@ class ExternalBlockCache:
             if resolved_uri.endswith("#"):
                 resolved_uri = resolved_uri[:-1]
             if resolved_uri == "" or resolved_uri == base_uri:
-                return UseInternal
+                return USE_INTERNAL
 
             from asdf import open as asdf_open
 
@@ -47,7 +63,7 @@ class ExternalBlockCache:
                 blk = af._blocks.blocks[0]
                 if memmap and blk.header["compression"] == b"\0\0\0\0":
                     parsed_url = util._patched_urllib_parse.urlparse(resolved_uri)
-                    if parsed_url.scheme == "file":
+                    if parsed_url.scheme == "file" and blk.data_offset is not None:
                         # deal with leading slash for windows file://
                         filename = urllib.request.url2pathname(parsed_url.path)
                         arr = np.memmap(filename, np.uint8, "r", blk.data_offset, blk.cached_data.nbytes)
@@ -58,11 +74,11 @@ class ExternalBlockCache:
             self._cache[key] = arr
         return self._cache[key]
 
-    def clear(self):
+    def clear(self) -> None:
         self._cache = {}
 
 
-def relative_uri_for_index(uri, index):
+def relative_uri_for_index(uri: str, index: int) -> str:
     # get the os-native separated path for this uri
     path = util._patched_urllib_parse.urlparse(uri).path
     _, filename = os.path.split(path)
--- a/asdf/_block/key.py
+++ b/asdf/_block/key.py
@@ -26,6 +26,8 @@ of these Key instances all methods and attributes
 are private.
 """
 
+from __future__ import annotations
+
 import weakref
 from typing import Any
 
@@ -69,7 +71,7 @@ class Key:
             return False
         return r is obj
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Key):
             return NotImplemented
         if self._key != other._key:
@@ -78,6 +80,6 @@ class Key:
             return False
         return other._matches_object(self._ref())  # pyrefly: ignore [not-callable]
 
-    def __copy__(self):
+    def __copy__(self) -> Key:
         obj = self._ref if self._ref is None else self._ref()
         return type(self)(obj, self._key)
--- a/asdf/_block/key.py
+++ b/asdf/_block/key.py
@@ -27,6 +27,7 @@ are private.
 """
 
 import weakref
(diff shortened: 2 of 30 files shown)


asdf/_block/external.py
```diff
@@ -8,26 +8,42 @@
...
+        return str(self.value)
 
-class UseInternalType:
-    pass
 
+USE_INTERNAL: Final = _USE_INTERNAL_TYPE.USE_INTERNAL
 
-UseInternal = UseInternalType()
+#: Type corresponding to ``USE_INTERNAL`` value for use in type-checking
+UseInternal: Final = Literal[_USE_INTERNAL_TYPE.USE_INTERNAL]
```
zacharyburnett, line 37: should this be moved into `if TYPE_CHECKING:`?

asdf/_block/key.py
```diff
@@ -38,26 +41,26 @@ def _next_key(cls):
         cls._next += 1
         return key
 
-    def __init__(self, obj=None, _key=None):
+    def __init__(self, obj: Any | None = None, _key: int | None = None):
```
zacharyburnett, line 44: does `Optional` work here?

asdf/_compression.py
```diff
@@ -267,7 +279,7 @@ def decompress(fd, used_size, data_size, compression, config=None):
     """
     buffer = np.empty((data_size,), np.uint8)
 
-    compression = validate(compression)
+    compression = typing.cast("str", validate(compression))
```
zacharyburnett, line 282: how is this different from `compression: str =`?

zacharyburnett (approved):
looks good to me, I like all the new signatures. This should make using ASDF with a lsp much easier

asdf/_block/key.py
sydduckworth: `<type> | None` is the preferred syntax over `Optional[<type>]`. The latter will generate a Ruff linter warning.

asdf/_compression.py
sydduckworth: That would only work if `validate` had an untyped return value. In this case `validate` can return values other than `str` so we need `cast` to override the return type.

Really the better solution would be to either update `compress` and `decompress` to handle the case where `validate` returns `None` (since currently its relying on the fact that `validate` was actually already called), or remove the calls to `validate` and require that the `compression` parameter be a pre-validated string.

But that would involve functional changes so for now I just left it as-is.

asdf/_block/external.py
sydduckworth: I think in general `if TYPE_CHECKING:` is only used for imports not for type definitions. In this case the user of `UseInternal` would presumably gate the import behind an `if TYPE_CHECKING:` block so it should work out to the same thing either way.

asdf/generic_io.py
```diff
@@ -1122,7 +1125,7 @@ def get_file(
             except NotImplementedError as err:
                 msg = f"Unable to open {init} with mode {mode}"
                 raise ValueError(msg) from err
-            return get_file(fd, uri=uri or init, close=True)
+            return get_file(fd, uri=str(uri or init), close=True)
```
braingram, line 1128: Would you comment on the need for `str`? It looks reasonable but I'm curious how it came about.

asdf/_tests/_block/test_writer.py
```diff
@@ -47,7 +47,7 @@ def test_write_blocks(tmp_path, lazy, index, padding, compression, stream, seeka
                 assert r.header["compression"] == compression
             if padding:
                 assert r.header["allocated_size"] > r.header["used_size"]
-        if stream:
+        if streamed_block is not None:
```
braingram: This is fine but to nitpick I like using `stream` better since it's the parameterized argument to the test and not a result of the parameterization. Understanding `stream` just requires looking at the arguments whereas `streamed_block` required finding where `stream` then conditionally created `streamed_block` (in this case). If for some reason we change how `streamed_block` is constructed it's possible these checks could silently incorrectly pass whereas using `stream` would keep these checks independent.

asdf/generic_io.py
sydduckworth: At this point `init` can be either a `str` or `Path`, but we expect `uri` to be a `str`, so this does the conversion if needed.

asdf/_tests/_block/test_io.py
```diff
@@ -12,7 +12,9 @@
 def test_checksum(tmp_path):
     my_array = np.arange(0, 64, dtype="<i8")
     target_checksum = b"\xcaM\\\xb8t_L|\x00\n+\x01\xf1\xcfP1"
+    # pyrefly: ignore [bad-argument-type]
```
braingram, line 15: Should we rerun the baseline instead of adding these ignores? Would that work here?

asdf/generic_io.py
braingram: Thanks. And this came about because this function was type annotated? Does that mean we don't have a test for this bug?

sydduckworth: I don't think there is a bug now? `str(uri or init)` isn't doing a type-system cast: if `init` is a `Path` it will correctly convert it to a `str` at runtime. `str(Path("/foo")) == "/foo"`

But yeah this is because of type annotations. Initially I was typing uri as `uri: str | Path`, but then there were places inside asdf that are doing string manipulation on `uri`, so I had to convert it to just be a `str`.

asdf/_block/manager.py
```diff
@@ -625,6 +651,9 @@ def update(self, new_tree_size, pad_blocks, include_block_index, write_checksums
             break
         if last_block is None:
             new_block_start = new_tree_size
+        elif last_block.data_offset is None:
+            msg = "Unable to update ASDF file because source file is not seekable"
```
braingram, line 655: We're already checking if the file is seekable:
https://github.com/asdf-format/asdf/blob/a883bff7e15645fe3c4eb66dda9c482609ada0fb/asdf/_asdf.py#L902
If we add another check inside this update (at the start) would that be enough to appease the type checker?

asdf/_tests/_block/test_io.py
sydduckworth: I'd prefer not to add more things to the baseline file if possible. Ideally we'd eventually fix all of the items in the file and remove it.
Plus, if we ever want to switch to a different type-checker without baseline file support it would be a lot easier to just grep for and replace `pyrefly: ignore` instances.

asdf/_block/manager.py
```diff
@@ -661,8 +690,15 @@ def update(self, new_tree_size, pad_blocks, include_block_index, write_checksums
                 src += n
                 dst += n
 
+            def update_offset(offset: int | None) -> int:
+                if offset is None:
+                    msg = "Unable to update ASDF file because source file is not seekable"
```
braingram, line 695: Same as https://github.com/asdf-format/asdf/pull/2048/changes#r3288891809

asdf/_block/manager.py
sydduckworth: So unfortunately the type checker is not able to infer that "file is not seekable" means "`data_offset` is not `None`".  Technically even if a file is seekable, somewhere else in the code could set `data_offset = None` without the type checker complaining, which means it can't be assumed to be non-`None` here.

In general this type of pattern is not handled well by the type system.

PR #2049, opened 2026-05-13
Remove `_InheritDocstrings`

sydduckworth:
## Description
* Removed `_InheritDocstrings` metaclass

`_InheritDocstrings` seems unnecessary because its intended behavior is already built in to Sphinx (and other docs generators).
Additionally it will overwrite the docstring of a child method with the parent docstring even if the child is already documented, which doesn't seem like desired behavior.

This PR produces no changes in the public API documentation.


braingram (approved):
Thanks!

PR #2054, opened 2026-05-29
Pin `globus-sdk` in dkist downstream CI

sydduckworth:
## Description
* Updated `dkist` environment in `tox.ini` to pin `globus-sdk==4.5.0`

This change allows the `dkist` downstream CI job to pass.
See DKISTDC/dkist#725 for details.

Once `dkist` is updated to either pin `globus-sdk` itself or resolve the incompatibility with `>=4.6.0` this change can be reverted.


braingram (approved):
Thanks for fixing it. We can remove it when the issues are addressed in dkist.

braingram:
Oh fun... a pytestdev failure. Any clue what's going on with that one?

sydduckworth:
The pytestdev failure is also happening on main. I was just about to put an issue in for it. No idea what's causing it yet.

PR #2056, opened 2026-06-01
Change block index verification

braingram:
Prior to this PR the first and last blocks were loaded.

Now, prior to the index finding the start of the first block is found
The first blocks typically starts immediately after the tree.
This is nice because the code just finished reading the tree meaning
checking for the first block should usually only take reading 4 bytes
to read the block magic. The exception is when the tree is padded.
Here the entire padding needs to be read.

Now that the code knows where the first block starts, it uses
that offset to check the block index (if one is found). If the
first index doesn't match, the code falls back to reading serially.
If it matches, the code uses the block index.

This access pattern should be more efficient (and a better match
to cloud access) where the old code did something like the following.

=== old routine ===

Assume we have a file with:
[ ---- tree --- _padding_ blk0 blk1 ... blkN index ]

For old and new reading starts with the tree
             fp index
                |
[ ---- tree --- ********************************** ]

For old next the index was found which required a seek to the end
then reading backwards to find the index
                |----------------------- seek --->|
[ ---- tree --- **************************** index ]

then a seek to the start to read the first block (after padding)
                          |<-seek-----------------|
[ ---- tree --- _padding_ blk0 ************* index ]

then a seek to the last block and read it
                              ---seek->|
[ ---- tree --- _padding_ blk0 ******** blkN index ]

=== new routine ===

For the new routine the process is simplier and starts the same with
reading the tree:
             fp index
                |
[ ---- tree --- ********************************** ]

Then the first block is found (just the magic)
                          |
[ ---- tree --- _padding_ ************************ ]

Then the file is searched for a block index (and read):
                          |-------------- seek --->|
[ ---- tree --- _padding_ *******************index ]

and that's it.

One possible (and very unlikey) downside is if a file somehow has
- a valid block index
- a valid first block offset
- an invalid last block offset

we would not catch the error (but we would before). It seems impossible
to generate such a file without manually modifying block headers. If
that's happening there is no reason non first/last blocks might also
be incorrect so the old routine was also not absolute.

## Tasks

</details>

### DIFF
--- a/asdf/_block/reader.py
+++ b/asdf/_block/reader.py
@@ -232,7 +232,7 @@ def read_blocks(
         # load all blocks serially
         return _read_blocks_serially(fd, memmap, lazy_load, validate_checksums, after_magic)
 
-    # storet starting offset
+    # store starting offset
     starting_offset = fd.tell()
     magic_len = len(constants.BLOCK_MAGIC)
 
--- a/asdf/_block/reader.py
+++ b/asdf/_block/reader.py
@@ -278,4 +278,7 @@ def read_blocks(
         return _read_blocks_serially(fd, memmap, lazy_load, after_magic)
 
     # skip magic for each block
-    return [ReadBlock(offset + magic_len, fd, memmap, lazy_load, validate_checksums) for offset in block_index]
+    return [
+        ReadBlock(offset + magic_len if offset is not None else None, fd, memmap, lazy_load, validate_checksums)
+        for offset in block_index
+    ]
--- a/asdf/_block/reader.py
+++ b/asdf/_block/reader.py
@@ -232,8 +232,21 @@ def read_blocks(
         # load all blocks serially
         return _read_blocks_serially(fd, memmap, lazy_load, validate_checksums, after_magic)
 
-    # try to find block index
+    # storet starting offset
     starting_offset = fd.tell()
+    magic_len = len(constants.BLOCK_MAGIC)
+
+    # we should be at the first block, have no blocks, or have padding
+    if not after_magic:
+        # seek until the first magic is found
+        try:
+            fd.seek_until(b"(" + constants.BLOCK_MAGIC + b")", magic_len)
+        except DelimiterNotFoundError:
+            fd.seek(starting_offset)
+            return []
+    first_block_offset = fd.tell() - magic_len
+
+    # try to find block index
     index_offset = bio.find_block_index(fd, starting_offset)
     if index_offset is None:
         # if failed, load all blocks serially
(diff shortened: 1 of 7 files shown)


asdf/_block/reader.py
```diff
@@ -218,21 +231,21 @@ def read_blocks(fd, memmap=False, lazy_load=False, validate_checksums=False, aft
...
+    # check that the offset for the first block matches
+    if block_index[0] != first_block_offset:
+        msg = "Invalid block index contents for block 0, falling back to serial reading"
         warnings.warn(msg, AsdfBlockIndexWarning)
         fd.seek(starting_offset)
         return _read_blocks_serially(fd, memmap, lazy_load, after_magic)
-    return blocks
+
+    # skip magic for each block
+    fd.seek(starting_offset)
```
braingram: This seek is here only to keep sphinx-asdf from crashing for asdf directives that show blocks. Once https://github.com/asdf-format/asdf/pull/2057 is merged we can remove this seek.

braingram: linked PR merged and released

sydduckworth (commented):
There's a chunk of code not being hit by unit tests, which is the logic described in #2062 that emits a warning if there's an empty block index but no blocks.
Should probably add tests to capture that behavior.

Otherwise LGTM

braingram:
> There's a chunk of code not being hit by unit tests, which is the logic described in #2062 that emits a warning if there's an empty block index but no blocks. Should probably add tests to capture that behavior.
> 
> Otherwise LGTM

Thanks! Good eye finding the uncovered code. I pushed a change in [2e43751](https://github.com/asdf-format/asdf/pull/2056/commits/2e43751ab347ac712ec7378a35a2b46d11b5b16c) that:
- makes the warning within that block consistent with the other warnings
- updates the `test_invalid_block_index` test to cover that code (I noticed this test wasn't truncating the file after trashing the index and since truncation is a mess on windows I updated the test to use an in-memory buffer instead of writing to disk)

The updated test was causing issues with pyrefly. Part of this seemed to be an incorrect type for the return value of `read_block_index` so I updated that to be `list[int]` instead of `list[int | None]`.

sydduckworth:
> The updated test was causing issues with pyrefly. Part of this seemed to be an incorrect type for the return value of `read_block_index` so I updated that to be `list[int]` instead of `list[int | None]`.

Ah yeah I think I set the typing of that function because when constructing the block index the values can be `None`, and I hadn't yet found the logic that prevents writing out a block index that contains `None` values.

sydduckworth (approved):

PR #2059, opened 2026-06-03
Add deprecation warning for reading non-seekable files

### LINKED ISSUE #2050: Deprecate non-seekable file support
Currently asdf supports reading and writing non-seekable files, which adds a significant amount of complexity to the file reading logic (less so on the write side).
By deprecating and removing non-seekable file support we would be able to significantly streamline file reading logic.

The only common cases where you encounter a non-seekable file are:
* Sockets
* Network responses (e.g. HTTP)
* Standard streams (stdin/stdout)

For stdin/stdout it is worth noting that those streams actually *are* seekable in the case of file redirection, e.g. `python read_stdin.py < myfile.asdf` and `python write_stdout.py > myfile.asdf` both produce seekable streams. 

## Open Questions

We could retain support for *writing* non-seekable files. The only place where seek is used while writing a file is to pad a section with zeros, which could pretty easily be replaced with an explicit `write` call.

## Steps
* Verify no downstream projects are passing non-seekable files to asdf
* Add deprecation warning when asdf encounters a file where `.seekable() ==  False`
* Remove seek support:
  * Remove `InputStream` and `OutputStream` types from `asdf.generic_io`
  * Remove `if f.seekable():` checks in read logic

sydduckworth:
## Description
* This PR deprecates reading from non-seekable files. 
  * By extension this requires deprecating support for duck-typed files (e.g. objects with just a `read` or `write` method)
  * The only non-deprecated file objects supported moving forward are ones that subclass `IOBase`.
  * We could potentially continue to support duck-typed files by checking if they have a `seekable` attribute, but requiring `IOBase` provides better future opportunities for simplifying/removing `generic_io` since we can assume the availability of significantly more file methods.
* Added overloads for `AsdfFile.write_to`, `open_asdf`, and `generic_io.get_file`:
  * First overload matches the supported file argument types (`Path | str | IOBase | GenericFile`)
  * Second overload matches duck-typed files (`Reader | Writer`) and has a `@deprecated` decorator
* Added runtime deprecation warnings to `generic_io.get_file` if file is duck-typed or if `mode != "w"` and file is not seekable.
* Updated relevant tests to add a `with pytest.deprecated_call():` block around tests that use non-seekable or duck-typed files.

Closes #2050

### DIFF
--- a/asdf/_tests/_regtests/test_1542.py
+++ b/asdf/_tests/_regtests/test_1542.py
@@ -17,7 +17,7 @@ def test_failure_to_write_blocks_to_non_seekable_file():
     https://github.com/asdf-format/asdf/issues/1542
     """
     r, w = os.pipe()
-    with os.fdopen(r, "rb") as rf:
+    with os.fdopen(r, "rb") as rf, pytest.deprecated_call():
         with os.fdopen(w, "wb") as wf:
             arrs = [np.zeros(1, dtype="uint8") + i for i in range(10)]
             af = asdf.AsdfFile({"arrs": arrs})
--- a/asdf/_tests/test_generic_io.py
+++ b/asdf/_tests/test_generic_io.py
@@ -44,7 +44,8 @@ def test_mode_fail(tmp_path):
     path = os.path.join(str(tmp_path), "test.asdf")
 
     with pytest.raises(ValueError, match=r"mode must be 'r', 'w' or 'rw'"):
-        generic_io.get_file(path, mode="r+")  # pyrefly: ignore[bad-argument-type]
+        # pyrefly: ignore[no-matching-overload]
+        generic_io.get_file(path, mode="r+")
 
 
 @pytest.mark.parametrize("mode", ["r", "w", "rw"])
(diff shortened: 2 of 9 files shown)


asdf/_tests/_regtests/test_1542.py
```diff
@@ -17,7 +17,7 @@ def test_failure_to_write_blocks_to_non_seekable_file():
     https://github.com/asdf-format/asdf/issues/1542
     """
     r, w = os.pipe()
-    with os.fdopen(r, "rb") as rf:
+    with os.fdopen(r, "rb") as rf, pytest.deprecated_call():
```
braingram, line 20: One benefit of using `match` for these deprecation checks is that it makes it easy to find what tests should be removed once the feature is removed (like this one).

asdf/_tests/test_generic_io.py
```diff
@@ -478,19 +482,22 @@ class All(Reader, Writer, Random):
...
 
     buff = io.BytesIO()
-    assert isinstance(generic_io.get_file(Reader(buff), "r"), generic_io.InputStream)
-    assert isinstance(generic_io.get_file(Writer(buff), "w"), generic_io.OutputStream)
-    assert isinstance(generic_io.get_file(RandomReader(buff), "r"), generic_io.MemoryIO)
-    assert isinstance(generic_io.get_file(RandomWriter(buff), "w"), generic_io.MemoryIO)
-    assert isinstance(generic_io.get_file(All(buff), "rw"), generic_io.MemoryIO)
-    assert isinstance(generic_io.get_file(All(buff), "r"), generic_io.MemoryIO)
-    assert isinstance(generic_io.get_file(All(buff), "w"), generic_io.MemoryIO)
+    with pytest.deprecated_call():
```
braingram, line 485: I would have been ok with a filterwarnings mark for this whole function.

braingram (approved):
Thanks! Just a couple comments about the use of `deprecated_call` but nothing that holds up approval.

PR #2065, opened 2026-06-08
forward port: render changelog for 5.3.1

braingram:
Update changelog on main for 5.3.1 release.

## Tasks

</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ exclude = ["asdf/_jsonschema/json/*"]
 
 [tool.setuptools_scm]
 version_file = "asdf/_version.py"
+version_scheme = "release-branch-semver"
 
 [tool.black]
 line-length = 120
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,13 @@
+5.3.1 (2026-06-08)
+==================
+
+Bugfix
+------
+
+- Fix bug where ``add_history_entry`` incorrectly used a custom schema if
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -72,6 +72,7 @@ exclude = ["asdf/_jsonschema/json/*"]
 
 [tool.setuptools_scm]
 version_file = "asdf/_version.py"
+version_scheme = "release-branch-semver"
```
braingram, line 75: This configures setuptools-scm to know about our release branch process:
https://setuptools-scm.readthedocs.io/en/latest/extending/#vcs_versioning.ScmVersion.branch

While looking up docs for this settings it looks like this scheme is now "semver-pep440-release-branch" (but the scheme in this PR still works). The new scheme was only introduced in March and since the old one works I'm inclined to use the old one. Let me know if I should change this to the new scheme.

I had to add this to fix the compatibility tests since on main (without this configure) the version is incorrectly 5.3.1.dev... which is <5.3.1 whereas it should be 5.4.0.dev on main. With this setting the version on main is correct.

braingram:
@zacharyburnett would you give this a review? The 5.3.1 release combined with main reporting the incorrect versino (5.3.1.dev instead of 5.4.0) mans that the compatibility job is failing on main:
https://github.com/asdf-format/asdf/actions/runs/27158451155/job/80167903654#step:11:242

This fixes that issue (and updates the changelog on main to reflect the 5.3.1 release).

sydduckworth (approved):

PR #2068, opened 2026-06-11
Fix failure to load files with untagged history metadata

### LINKED ISSUE #2063: Assumed tag for file metadata leads to error
The following file:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.6.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
history:
  extensions:
    - extension_class: "asdf.extension._manifest.ManifestExtension"
      extension_uri: "asdf://asdf-format.org/core/extensions/core-1.6.0"
      manifest_software:
        name: "asdf_standard"
        version: "1.1.1"
      software:
        name: "asdf"
        version: "4.1.0"
...
```
is valid as the asdf-1.1.0 schema:
https://github.com/asdf-format/asdf-standard/blob/main/resources/stable/schemas/stsci.edu/asdf/core/asdf-1.1.0.yaml
doesn't require that the history/extensions items have a specific tag (even if we ignore https://github.com/asdf-format/asdf-standard/issues/436). This can be seen by loading and validating the above file (example.asdf) via:
```python
import asdf
asdf.schema.validate(asdf.util.load_yaml("example.asdf", tagged=True))
```

However if we try to load the file with `asdf.open` we encounter an error:
```
File ~/projects/src/asdf/asdf/_asdf.py:278, in AsdfFile._check_extensions(self, tree, strict)
    276 installed = None
    277 for ext in self._user_extensions + self._plugin_extensions:
--> 278     if (extension.extension_uri is not None and extension.extension_uri == ext.extension_uri) or (
    279         extension.extension_uri is None and extension.extension_class in ext.legacy_class_names
    280     ):
    281         installed = ext
    282         break

AttributeError: 'dict' object has no attribute 'extension_uri'
```
since the python library assumes the extension objects are tagged (and then deserialized to a custom object).

This might extend to other objects/tags in the standard which are linked to the asdf object schema (which uses refs) including:
- history_entry
- software (which is also referenced in history_entry)

The expectation here is that these are valid files and the python library shouldn't produce an error on loading.

sydduckworth:
## Description
* Fixed files failing to load if their history section contains extension entries without schema tags

Closes #2063

### DIFF
--- a/asdf/_tests/test_history.py
+++ b/asdf/_tests/test_history.py
@@ -334,3 +334,5 @@ history:
 
     with asdf.open(BytesIO(file.encode())) as af:
         af.validate()
+
+        af.write_to(BytesIO())
--- a/asdf/_tests/test_history.py
+++ b/asdf/_tests/test_history.py
@@ -325,7 +325,7 @@ history:
       extension_uri: "asdf://asdf-format.org/core/extensions/core-1.6.0"
       manifest_software:
         name: "asdf_standard"
-        version: "1.1.1"
+        version: "1.1.0"
       software:
         name: "asdf"
         version: "4.1.0"
--- a/asdf/_tests/test_history.py
+++ b/asdf/_tests/test_history.py
@@ -334,5 +334,7 @@ history:
 
     with asdf.open(BytesIO(file.encode())) as af:
         af.validate()
+        # check that extensions entries remain untagged
+        assert type(af["history"]["extensions"][0]) is dict
 
         af.write_to(BytesIO())
--- a/asdf/_tests/test_history.py
+++ b/asdf/_tests/test_history.py
@@ -312,8 +312,8 @@ def test_history_ignores_custom_schema(tmp_path):
     assert af.get_history_entries()[0]["description"] == "test"
 
 
-def test_history_missing_schema_tag():
-    """Test that a file can be loaded and validated even if its extension list entries don't have schema tags."""
+def test_history_extensions_missing_tags():
+    """Test that a file can be loaded and validated even if its extension list entries don't have tags."""
     file = """#ASDF 1.0.0
 #ASDF_STANDARD 1.6.0
 %YAML 1.1
--- a/asdf/_tests/test_history.py
+++ b/asdf/_tests/test_history.py
@@ -1,6 +1,7 @@
 import datetime
 import fractions
 import warnings
+from io import BytesIO
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


changes/2068.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed files failing to load if their history section contains extension entries without schema tags.
```
braingram:
```suggestion
Fixed files failing to load if their history section contains extension entries without tags.
```
Minor edit since these are yaml tags.

asdf/_tests/test_history.py
```diff
@@ -309,3 +310,27 @@ def test_history_ignores_custom_schema(tmp_path):
 
     af.add_history_entry("test")
     assert af.get_history_entries()[0]["description"] == "test"
+
+
+def test_history_missing_schema_tag():
```
braingram:
```suggestion
def test_history_extensions_missing_tags():
```

asdf/_tests/test_history.py
```diff
@@ -309,3 +310,27 @@ def test_history_ignores_custom_schema(tmp_path):
...
+    """Test that a file can be loaded and validated even if its extension list entries don't have schema tags."""
```
braingram:
```suggestion
    """Test that a file can be loaded and validated even if its extension list entries don't have tags."""
```

braingram (commented):
`_update_extension_history` needs a similar update:
https://github.com/asdf-format/asdf/blob/9c89dc2d72583f1e482bfc4f3833263d117e7ab7/asdf/_asdf.py#L432-L437

This can be seen by attempting to add a `af.write_to` to the unit tests fails with:
```
*** AttributeError: 'dict' object has no attribute 'extension_uri'
```

asdf/_tests/test_history.py
```diff
@@ -309,3 +310,29 @@ def test_history_ignores_custom_schema(tmp_path):
...
+      software:
+        name: "asdf"
+        version: "4.1.0"
+...
+    """
+
+    with asdf.open(BytesIO(file.encode())) as af:
+        af.validate()
+
+        af.write_to(BytesIO())
```
braingram, line 340:
```suggestion
    with asdf.open(BytesIO(file.encode())) as af:
        af.validate()
        # check that extensions entries remain untagged
        assert type(af["history"]["extensions"][0]) is dict

        af.write_to(BytesIO())
```
On minor request to expand this to check that we don't silently convert the extensions on read.

braingram (approved):
Thanks!

PR #2072, opened 2026-06-17
Fix zero-length ndarrays failing to round-trip

### LINKED ISSUE #2071: 0 dimension inline arrays fail to roundtrip
```python
import asdf, numpy as np
t = {"arr": np.zeros((0, 0), dtype="f4")}
asdf.loads(asdf.dumps(t, all_array_storage="inline"))
```

returns:
```
ValueError: inline data doesn't match the given shape
```

tested on asdf main

sydduckworth:
## Description
* Fixed arrays with zero-length axes failing to round trip correctly
* Added test case to verify arrays with zero-length axes work as expected

Closes #2071


changes/2072.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed ndarrays with zero-length dimensions failing to round-trip correctly.
```
braingram:
```suggestion
Fixed ndarrays with zero-length dimensions failing to round-trip correctly when saved inline.
```

asdf/_tests/tags/core/tests/test_ndarray.py
```diff
@@ -1129,3 +1129,20 @@ def test_lazy_load_array_class(tmp_path, lazy_load, lazy_tree, array_class):
 
     with asdf.open(file_path, lazy_load=lazy_load, lazy_tree=lazy_tree) as af:
         assert type(af["arr"]) is array_class
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (0,),
+        (0, 10),
```
braingram, line 1139:
```suggestion
        (0, 0),
        (0, 10),
```
To add an all-0 multidimensional case.

braingram (approved):
2 minor comments.

I tested this with roman_datamodels (and setting an inline_array_threshold) and unfortunately it doesn't fix all the failures. Running `pytest tests/test_models.py::test_create_minimal_copies[FpsModel]` produces an error (with this PR and asdf main):
```
E           On instance['meta']['groundtest']['frames_temp']:
E               {'shape': [], 'data': 0.0, 'datatype': 'float64'}

../asdf/asdf/_jsonschema/validators.py:312: ValidationError
```
so it looks like in addition to the structured array inline issue there's an issue for scalar arrays not correctly being stored inline. I'll open 2 more issues.

PR #2075, opened 2026-06-22
add 3.15 job

### LINKED ISSUE #2043: Allow CI jobs to run in parallel
Currently in the `ci.yml` workflow the jobs are configured so that `core` doesn't run until after `pre-commit`, and most of the other jobs don't run until after `core`.
Removing those dependencies would speed up execution by allowing the jobs to run in parallel.

Also, because of that configuration the later jobs only run if the earlier jobs are successful. I've run into issues with this when I've pushed a PR where the `pre-commit` job failed, and then after fixing that I found that the later jobs were failing for a different reason.

braingram:
## Description

Add a 3.15 CI job (with devdeps to pick up the dev numpy that supports 3.15).

Also fixes #2043 

All tests pass so looks like 3.15 support is good-to-go.

No LLMs were used for this PR.

## Tasks

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
         - linux: py313-coverage-devdeps-parallel
         - linux: py314-coverage-devdeps-parallel
         - linux: py315-coverage-pytestdev-parallel
-          python-version: 3.15-dev
+          python-version: 3.15
         # separate pytest so a failure here doesn't cause the whole suite to fail
         - linux: py314-coverage-pytestdev-parallel
       coverage: codecov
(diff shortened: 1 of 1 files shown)


sydduckworth:
Do we also want a 3.15 step for the `core` job?

braingram:
> Do we also want a 3.15 step for the `core` job?

Eventually yes. Numpy has not yet published stable 3.15 wheels:
https://pypi.org/project/numpy/2.5.0/#files
so we can only (easily) test against dev numpy.

Once numpy has stable 3.15 wheels we can add 3.15 to core.

sydduckworth (approved):

PR #2076, opened 2026-06-22
Update PR template

sydduckworth:
## Description

* Updated local PR template to include changes from organization-level template

We could alternatively remove this template so that the repo can just inherit the org-level template, but having the specific guidance on changelog and CI seems helpful.


zacharyburnett (approved):
it's too bad we can't patch on inherited files

PR #2091, opened 2026-07-07
remove NOT_SET restore NotSet

### LINKED ISSUE #2090: passing asdf.util.NotSet for all_array_storage to write_to no longer works on main
I suspect this is related to:
https://github.com/asdf-format/asdf/pull/2031

With asdf 5.3.1 the following works as expected:

```python
import asdf
af = asdf.AsdfFile()
af.write_to("foo.asdf", all_array_storage=asdf.util.NotSet)
```

However on asdf main the above fails with:
```python
ValueError: Invalid value for all_array_storage: 'typing.Literal[NotSet]'
```

This is causing devdeps for romancal tests:
https://github.com/spacetelescope/RegressionTests/actions/runs/28851326247

braingram:
## Description
Fixes https://github.com/asdf-format/asdf/issues/2090

Remove `asdf.util.NOT_SET` (we haven't released a version that contains this) and restore `asdf.util.NotSet`. This retains backwards compatibility.

I was able to reproduce the romancal regtest failure locally and this PR fixes the local failure. Running full regtests with this PR to confirm the fix:
https://github.com/spacetelescope/RegressionTests/actions/runs/28875897376
all passing.

## AI Disclosure
No AI was used.

## Tasks

</details>

### DIFF
--- a/asdf/_asdf.py
+++ b/asdf/_asdf.py
@@ -892,11 +892,11 @@ class AsdfFile:
 
         with config_context() as config:
             if all_array_storage is not NotSet:
-                config.all_array_storage = all_array_storage  # type: ignore
+                config.all_array_storage = typing.cast("ArrayStorage", all_array_storage)
             if all_array_compression is not NotSet:
-                config.all_array_compression = all_array_compression  # type: ignore
+                config.all_array_compression = typing.cast("Compression", all_array_compression)
             if compression_kwargs is not NotSet:
-                config.all_array_compression_kwargs = compression_kwargs  # type: ignore
+                config.all_array_compression_kwargs = typing.cast("dict", compression_kwargs)
 
             fd = self._fd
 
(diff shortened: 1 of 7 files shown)


sydduckworth (changes requested):
Can we instead do something like this? 
Keep `NOT_SET` and just alias `NotSet = NOT_SET`. 
That way type checking still works and we could potentially deprecate the alias at some point.

```py
#: Special value indicating that a parameter is not set.
#: Distinct from None, which may for example be a value of interest in a search.
NOT_SET: Final = _NOT_SET_TYPE.NOT_SET
NotSet: Final = NOT_SET

#: Type corresponding to ``NOT_SET`` value for use in type-checking
NotSetType: TypeAlias = Literal[_NOT_SET_TYPE.NOT_SET]
```

braingram:
> Can we instead do something like this? Keep `NOT_SET` and just alias `NotSet = NOT_SET`. That way type checking still works and we could potentially deprecate the alias at some point.
> 
> ```python
> #: Special value indicating that a parameter is not set.
> #: Distinct from None, which may for example be a value of interest in a search.
> NOT_SET: Final = _NOT_SET_TYPE.NOT_SET
> NotSet: Final = NOT_SET
> 
> #: Type corresponding to ``NOT_SET`` value for use in type-checking
> NotSetType: TypeAlias = Literal[_NOT_SET_TYPE.NOT_SET]
> ```

No objection from me but let's handle it in a follow-up PR where we add tests to make sure the changes don't break backwards compatibility.

sydduckworth (approved):

PR #2096, opened 2026-07-10
Port downstream CI from tox to nox

sydduckworth:
## Description
- Added `noxfile.py` 
- Added `downstream` nox session session parameterized for all downstream test environments
- Updated `downstream.yml` CI workflow to use nox instead of tox 
- Note that `noxfile.py` also contains implementations for most of the test suite run by `ci.yml`, but I will be putting in a separate PR to finalize and enable those sessions.

The new nox sessions should functionally be the same as the tox environments (same dependencies, same pytest arguments, etc.). There are a few small behavioral changes I've noted below:

- Nox is configured to use the `uv` backend when available, which means there will be some differences in the build/installation process compared to `pip`. In the one test suite where that seems to be a problem (`oldestdeps`) I've configured it to not use `uv`.
-  I didn't include the step of running `pip freeze` at the end of setup. Instead in CI we run nox with `nox -v` which causes it to log all of the installed dependencies anyway.
- I've omitted the step in `abacusutils` that creates an empty `pytest.ini`. The comment in the tox file says this is necessary to prevent pytest from using asdf's pytest configuration, but in my testing this isn't actually needed because pytest uses `abacusutil`'s `pyproject.toml` instead and doesn't try to search the parent directory.

## AI Disclosure
No AI tools used

## Tasks

</details>


noxfile.py
```diff
@@ -0,0 +1,324 @@
...
+    session.install("-e", ".[all,tests]")
+
+    session.run(*pytest_args(parallel=False), "--jsonschema")
+
+
+@nox.session(
+    tags=["test", "core"],
+    python="3.10",
+    # This test doesn't work with the uv backend
+    # Probably due to differences in build isolation
```
zacharyburnett: if the issue here was with `minimum_dependencies`, maybe using uv's `lowest-direct` resolution if using `uv` as a backend would work?

zacharyburnett (approved):
this LGTM, and using the noxfile makes the downstream package definition cleaner

braingram:
Looking at the asdf-astropy job (as an example). Here is the log on main:
https://github.com/asdf-format/asdf/actions/runs/29242147391/job/86790483912#step:11:141
and the log with this PR:
https://github.com/asdf-format/asdf/actions/runs/29266711406/job/86874236463?pr=2096#step:5:91

one thing I'm not seeing that has been useful is a `pip list` (or other output of the packages in the test environment) prior to the pytest run. This is helpful for a number of reasons:
- to confirm the tests picked up the appropriate asdf commit
- to inspect dependency versions (useful when investigating sudden failures which might be caused by a release of a dependency)

It would be helpful to keep this feature.

noxfile.py
braingram: `lowest-direct` and `minimum_dependencies` aren't equivalent (the former silently bumps versions).

.github/workflows/downstream.yml
```diff
@@ -23,55 +23,31 @@ concurrency:
...
-        - linux: roman_datamodels
-
-  third-party:
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v2
+    runs-on: ubuntu-latest
+    outputs:
+      session: ${{ steps.set-matrix.outputs.session }}
+    steps:
+    - uses: actions/checkout@v7
+    - uses: wntrblm/nox@3284968a16fa5bd7fc0cc64a69a531404ccb4756 # 2026.07.11
```
braingram: Let's not mix hash-pins and version pins. If you'd like to use hashes would you open a pr switching the CI to that (IIRC zizmor has a command that will automatically make the changes) first and then use hashes here?

pyproject.toml
```diff
@@ -46,18 +46,21 @@ benchmark = ["asdf[tests]", "pytest-benchmark"]
 test = ["asdf[tests]"]
 
 [project.urls]
-'documentation' = "https://asdf.readthedocs.io/en/stable/"
-'repository' = "https://github.com/asdf-format/asdf"
-'tracker' = "https://github.com/asdf-format/asdf/issues"
+"documentation" = "https://asdf.readthedocs.io/en/stable/"
```
braingram, line 49: What caused this reformatting?

noxfile.py
sydduckworth: Unfortunately `lowest-direct` produces different behavior than `minimum_dependencies`. Namely that `minimum_dependencies` seems to only impact required dependencies while `lowest-direct` also affects extra dependencies (e.g. from `test`).

noxfile.py
```diff
@@ -0,0 +1,324 @@
...
+def downstream(session, pkg: Package):
+    """Run the test suite for a downstream package against the local asdf version."""
+    dir = pkg.download_and_install(session)
+    session.install("-e", ".[all,tests]")
+
+    with session.cd(dir):
+        pkg.test(session)
+
+
+@nox.session(tags=["test", "core"], python=["3.10", "3.11", "3.12", "3.13"])
```
braingram: Is this used anywhere? Is the plan to switch the entire CI to nox? That sounds good to me. Can we do this in 1 PR (and remove tox)?

pyproject.toml
sydduckworth: Seems like as of a recent release the Tombi formatter now applies double quote formatting by default. Didn't seem like a big deal since we use double quotes everywhere else except those few lines.

noxfile.py
sydduckworth: Yes that is the plan. I wanted to split it into multiple PRs since replacing `ci.yml` will probably require more iteration to deal with things like code coverage.

pyproject.toml
braingram: Can we wait until prek picks that up that version? It's distracting to have unrelated formatting in this PR.

.github/workflows/downstream.yml
sydduckworth: From our prior conversation it sounded like hashes were preferred over fully pinned versions (`X.X.X`) for actions that don't publish major version tags. Would you prefer using the fully pinned version tags?

noxfile.py
braingram: Thanks. Splitting works for me (if it seems doable to you). Since this is only addressing downstream would you:
- remove the sessions in the noxfile that aren't used in downstream
- remove the downstream environments in the tox file

.github/workflows/downstream.yml
braingram: I'd prefer major version pinned workflows but that won't work with uv (added in this PR) so second to that hash-pinning would be the next best.

sydduckworth:
> Looking at the asdf-astropy job (as an example). Here is the log on main: https://github.com/asdf-format/asdf/actions/runs/29242147391/job/86790483912#step:11:141 and the log with this PR: https://github.com/asdf-format/asdf/actions/runs/29266711406/job/86874236463?pr=2096#step:5:91
> 
> one thing I'm not seeing that has been useful is a `pip list` (or other output of the packages in the test environment) prior to the pytest run. This is helpful for a number of reasons:
> 
>     * to confirm the tests picked up the appropriate asdf commit
> 
>     * to inspect dependency versions (useful when investigating sudden failures which might be caused by a release of a dependency)
> 
> 
> It would be helpful to keep this feature.

It does show the list of installed packages. It's using `uv`'s +/- syntax but it is a fully reproducible list of packages.

If that format isn't acceptable I can add the `pip freeze` step.

braingram:
> It does show the list of installed packages. It's using `uv`'s +/- syntax but it is a fully reproducible list of packages.
> 
> If that format isn't acceptable I can add the `pip freeze` step.

Thanks. The +/- syntax is difficult to follow (it requires reading much more of the log and tracking what is installed and uninstalled at each step). Re-adding `pip freeze` or something equivalent would be great.

pyproject.toml
```diff
@@ -58,6 +58,9 @@ asdf_extensions = { builtin = "asdf.extension._legacy:BuiltinExtension" }
 [project.scripts]
 asdftool = "asdf._commands.main:main"
 
+[dependency-groups]
+mocks3 = ["boto", "fsspec", "moto[s3,server]", "s3fs>2023.12.1"]
```
braingram:
```suggestion
mocks3_test = ["boto", "fsspec", "moto[s3,server]", "s3fs>2023.12.1"]
```
Maybe adding `test` is worthwhile here to make it more obvious these are only for testing.

noxfile.py
```diff
@@ -0,0 +1,324 @@
...
+
+    session.install("-r", "requirements-dev.txt")
+    session.run(*pytest_args(parallel=True), "-W", "ignore::asdf_standard.exceptions.UnstableCoreSchemasWarning")
+
+
+@nox.session(tags=["test", "core"], python="3.12")
+def mocks3(session):
+    """Set up AWS mocks and run S3 integration tests"""
+    session.install("-e", ".[all,tests]")
+    session.install(*nox.project.dependency_groups(PYPROJECT, "mocks3"))
```
braingram:
```suggestion
    session.install("-e", ".[all,tests,mocks3_test]")
```
Does this work?

.github/workflows/downstream.yml
sydduckworth: Nox also doesn't provide major version tags.

pyproject.toml
sydduckworth: Dependency groups aren't visible to users in the same way that package extras are. They're only available during development and don't show up in the built package/wheel. So the fact its for testing seemed implied, but I can change it if it still seems confusing.

noxfile.py
sydduckworth: I just removed that session from the PR so I'll have to get back to you in the next PR

pyproject.toml
braingram: Good to know. Either name works for me. Should this go in a follow-up PR since it's not part of the downstream tests?

tox.ini
```diff
@@ -72,36 +72,6 @@ commands_pre =
...
-commands_pre =
-    bash -c "pip freeze -q | grep 'asdf @' > {env_tmp_dir}/requirements.txt"
-    git clone https://github.com/asdf-format/asdf-zarr.git
-    pip install -e asdf-zarr[tests]
-    pip install -r {env_tmp_dir}/requirements.txt
-    pip freeze
-commands =
-    pytest asdf-zarr
-
 [testenv:asdf-transform-schemas]
```
braingram: Should we take this opportunity to move these tests to downstream? They're not there on main (they're in asdf-schemas) but probably should be.

sydduckworth: Done

.github/workflows/downstream.yml
```diff
@@ -25,55 +25,37 @@ concurrency:
...
+      matrix:
+        session: ${{ fromJson(needs.generate-jobs.outputs.session) }}
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+          # Need tags for dynamic versioning
+          fetch-depth: 0
+          persist-credentials: false
+    - uses: | # v8.3.2 # zizmor: ignore[cache-poisoning] this workflow doesn't publish any artifacts
+        astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990
```
braingram, line 57:
```suggestion
```
Is there a benefit to using uv here?
Testing here:
https://github.com/braingram/asdf/actions/runs/29282578163
the workflow runs fine without it and any speed gained by uv (if there is any, it sounds like setup-python can be faster: https://github.com/astral-sh/setup-uv/#do-i-still-need-actionssetup-python-alongside-setup-uv) isn't practically realized since time to run tests limits when this workflow completes. 1 benefit to removing it is 1 less CI workflow to ping us for updates.

tox.ini
braingram: Thanks! Would you also remove these from ci.yml:
https://github.com/asdf-format/asdf/blob/c1cb05426dc6142d9b14d4a5245c1f0fa2d247e4/.github/workflows/ci.yml#L82-L90

they are currently just re-running the asdf tests:
https://github.com/asdf-format/asdf/actions/runs/29284241570/job/86932786668?pr=2096

That's unexpected tox behavior to me so another reason to move to nox!

.github/workflows/downstream.yml
```diff
@@ -25,55 +25,37 @@ concurrency:
...
-  third-party:
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@afacbe8ebafa67b369e1f605b4f6989284ba2e87 # v2.6.5
+    runs-on: ubuntu-latest
+    outputs:
+      session: ${{ steps.set-matrix.outputs.session }}
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        persist-credentials: false
+    - uses: wntrblm/nox@3284968a16fa5bd7fc0cc64a69a531404ccb4756 # 2026.07.11
```
braingram, line 37: One question about the nox version here. Is it correct that pinning to this workflow also pins the nox version? Put another way, we'd expect that when nox publishes a new version we'll get a dependabot PR updating this workflow (and won't silently get a new nox version like we currently get with tox)?

.github/workflows/downstream.yml
sydduckworth: Having the uv backend allows nox to automatically pull python toolchains, which without uv requires installing a package extra that the nox action doesn't include by default.
You can set the range of pre-fetched Python versions in the nox action, but I'd personally much rather have a single source of truth in the `noxfile` such that if we add, e.g. a session that needs Python 3.15, it just gets downloaded automatically instead of needing to remember/know to also go into the workflow and add the new Python version to the list.

.github/workflows/downstream.yml
sydduckworth: The nox action is part of the main nox repo, and as far as I can tell the action directly installs the package from repo, so the package version would always be the same as the action version.

tox.ini
sydduckworth: Done

.github/workflows/downstream.yml
braingram: Thanks for the details.

To check I'm following, the downstream tests all work without this since the runners already have python 3.12 (the version we've given to nox). If we asked for some version that wasn't available on the runner (say for the 3.10 or 3.15 jobs added in a later PR) we'd need uv or the nox extra (which isn't included by default). So we don't see an issue removing uv for this PR but may for the follow-ups.

Can we just disable the cache rather than ignore the warning?
```suggestion
    - uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
       with:
         enable-cache: false
```
We're not using this for a release but providing it cache access allows it to write a potentially poisoned cache.

sydduckworth: What's the advantage of disabling it? A sufficiently compromised action can always *write* to the cache (disabling caching doesn't prevent this). The vulnerability is that cache poisoning can be used to inject code into a non-compromised action if that action is reading from the cache. But that's only relevant if the reading action is publishing artifacts, which this workflow isn't. 
The only reason that zizmor is flagging the uv action is because this workflow triggers on tags and it assumes workflows that trigger on tags publish artifacts.

braingram: One advantage is that we don't have a relatively hidden ignore here for zizmor for an option that doesn't seem to be of benefit. Since this is currently our only direct use of setup-uv it seems harmless to have the ignore (although I think we do have indirect uses given the current cahes). My understanding of the cache poisoning is that action A (run with limited privileges) writes to a cache (in this case a malicious package). Then action B (with more privileges) uses the poisoned cache thinking it's safe (in this case using the malicious package instead of the 'safe' one from pypi). I admit that we're discussing a hypothetical here since we don't have any action B (at least directly) but why not stop the cache creation if it's of little to no benefit? https://github.com/astral-sh/setup-uv/issues/745

braingram: I'm not going to block progress for this. I think we should use the suggestion but given that we don't have a known issue we can either go forward with the ignore or suggestion.

braingram (approved):
Once this and the other PRs are in the readme should be updated to describe the nox usage.

.github/workflows/downstream.yml
sydduckworth: It's not possible to prevent "action A" from occurring. Any action can write to the cache regardless of whether the `enable-cache` input is set, and any action can write arbitrary cache keys: there's no per-action isolation and all namespacing is voluntary.
That's why the zimor lint exists to detect and prevent "action B", which is triggering a false positive in this workflow. 
In a scenario where we were using `setup-uv` with caching enabled and we *were* publishing artifacts it wouldn't matter if that was the only place we were using `setup-uv`: any arbitrary other action could theoretically poison the uv cache.

braingram: Sure we can't guard against a malicious action but we can tell setup-uv to not write a cache which might contain malicious packages.

PR #2104, opened 2026-07-14
Port core CI tests to nox

### LINKED ISSUE #2098: Port tox test suite to nox
## Background

Currently our CI uses [tox](tox.wiki) to set up and run all of the test environments. This generally works but has a few disadvantages:

* Since tox uses a plain INI/TOML file it can end up being very repetitive. All of our downstream test environments basically follow the same sequence of steps (clone repo, install repo, run tests) but the steps have to be repeated for every environment.
* Tox assumes all environments share a subset of dependencies, which makes it very poorly suited to setting up completely distinct environments. For example, if we wanted to add a new environment for building documentation it would inherit all of the test dependencies from the base configuration, but if we removed those dependencies from the base configuration we'd have to re-specify them in every test environment.
* Limited support for other backends. There are plugins for tox that allow you to use other backends (e.g. uv instead of pip) but that only affects the dependencies specified in the environment configuration. Our toxfile uses custom pip commands which break if you install the `tox-uv` plugin.

[nox](nox.thea.codes) addresses all of these issues:

* Nox environments are defined using Python code, so it's easy to create reusable functions for common behaviors. 
* With nox, each environment is entirely independent
* Nox has integrated support for multiple backends (pip/virtualenv, uv, mamba), with the ability to specify backends per-environment and set a fallback if a backend isn't available.

sydduckworth:
## Description
* Updated `noxfile.py` to add session definitions matching all of the remaining environments from `tox.ini`
* Updated `ci.yml` to use nox instead of tox
* Added `mocks3` dependency group to `pyproject.toml` to track AWS mock dependencies
* Added `nox` dependency to `typing` package extra to enable better type hints and code completion in the noxfile
* Deleted `tox.ini`
* Updated readme to reference nox instead of tox

This PR incorporates the changes to the CI jobs from #2086 where coverage is only run for Python 3.14.
Once this PR is ready to merge we will need to update the branch protections since some of the job names have now changed.

Closes #2098 

## AI Disclosure
No AI tools used

## Tasks

</details>


sydduckworth:
As was discussed in a previous PR, we're running into an issue where because our workflows are triggered on `tag: '*'` zizmor is assuming that the workflows are potentially being used to publish artifacts, which causes it to generate a `cache-poisoning` linter error on every action that uses caching.

At least for `ci.yml` I think we could probably just remove the `tag: '*'` trigger since it's triggered by any push to `main` or to a release branch, so it should already be run on all tagged releases anyway.

For `downstream.yml` (and also for `ci.yml` if we didn't want to change the workflow triggers) I think it would be a good idea to create a `zizmor.yml` file to just ignore `cache-poisoning` for that workflow, e.g.:

```yaml
# zizmor.yml
rules:
  cache-poisoning:
    ignore:
      - downstream.yml
```

.github/workflows/downstream.yml
```diff
@@ -39,7 +39,6 @@ jobs:
       shell: bash
       run: echo session=$(nox --json -l -t downstream | jq -c '[.[].session]') | tee --append $GITHUB_OUTPUT
   nox:
-    if: (github.repository == 'asdf-format/asdf' && (github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'Downstream CI')))
```
braingram, line 42: Would you restore this?

noxfile.py
```diff
@@ -71,22 +95,68 @@ def _generate_pytest_command(self):
...
     "CRDS_PATH": "/tmp/crds_cache",  # noqa: S108
     "CRDS_CLIENT_RETRY_COUNT": "3",
     "CRDS_CLIENT_RETRY_DELAY_SECONDS": "20",
 }
 
 # Downstream packages with test suites to run against the local asdf version
-DOWNSTREAM = [
+DOWNSTREAM: list[Package] = [
     ### asdf ###
-    Package(
```
braingram, line 84: Are these being run somewhere else? If not would you restore this and asdf-transform-schemas?

noxfile.py
```diff
@@ -192,22 +256,92 @@ def _generate_pytest_command(self):
...
+def core(session: Session) -> None:
+    """Run asdf test suite"""
+    Asdf(parallel=True).install(session).test(session)
+
+
+@nox.session(tags=["test", "coverage"], python="3.14")
+def coverage(session: Session) -> None:
+    """Run asdf test suite with coverage"""
+    (
+        Asdf(parallel=False)
```
braingram: Is there something new preventing these from being run in parallel?

tox.ini
```diff
@@ -1,58 +0,0 @@
...
-    stpipe
-    roman_datamodels
-    weldx
-    sunpy
-    dkist
-
-[testenv]
-set_env =
-    devdeps: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
-    devdeps: ASDF_UNSTABLE_CORE_SCHEMAS = 1
```
braingram, line 21: I don't see this being set anywhere. Are there tests using the unstable core schemas?

tox.ini
```diff
@@ -1,58 +0,0 @@
...
-    stdatamodels
-    stpipe
-    roman_datamodels
-    weldx
-    sunpy
-    dkist
-
-[testenv]
-set_env =
-    devdeps: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
```
braingram, line 20: I don't see this being set anywhere. Are there tests using the nightly numpy wheels?
I would expect:
https://github.com/asdf-format/asdf/actions/runs/29355946690/job/87163625457?pr=2104#step:5:120
to be:
2.6.0.dev:
https://github.com/asdf-format/asdf/actions/runs/29418039407/job/87361117653#step:11:41
I think we can add this index to requirements-dev.txt as one possible solution.

.github/workflows/ci.yml
```diff
@@ -48,79 +48,75 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+          fetch-depth: 0
+          persist-credentials: false
+    - uses: | # v8.3.2 # zizmor: ignore[cache-poisoning] this workflow doesn't publish any artifacts
+        astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990
+    - uses: wntrblm/nox@3284968a16fa5bd7fc0cc64a69a531404ccb4756 # 2026.07.11
+    - run: nox -v -s "core-3.11"
```
braingram:
```suggestion
    - run: nox -v -s "core-3.12"
```
Maybe we should bump the version here since 3.11 has been dropped by some key packages. We could also handle this in a separate PR if that's preferred.

.github/workflows/ci.yml
```diff
@@ -48,79 +48,75 @@ jobs:
...
+      with:
+          fetch-depth: 0
+          persist-credentials: false
+    - uses: | # v8.3.2 # zizmor: ignore[cache-poisoning] this workflow doesn't publish any artifacts
+        astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990
+    - uses: wntrblm/nox@3284968a16fa5bd7fc0cc64a69a531404ccb4756 # 2026.07.11
+    - run: nox -v -s "${MATRIX_SESSION}"
+      env:
+        MATRIX_SESSION: ${{ matrix.session }}
+  coverage:
```
braingram: Is there a slightly different way to arrange this where we just tag (almost) every session used in this workflow as "ci" (or some other tag) and then handle the selection of sessions within the nox file? That would have the benefit of less spread of the configuration between the workflow and nox file. I think the one exception is the platform test but maybe there is a way to make the runs-on conditional to something in the matrix? Similarly we'll need to make the coverage upload conditional to the session containing "coverage". Overall the goal is as you suggested to push as much of the parametrization/control of the sessions into the nox file and have less of it in the workflow.

braingram (commented):
I left a few comments about missing jobs and test changes. One suggestion about changing the relationship of workflow and noxfile configuration might require more extensive changes so I'll hold off on further review until that is discussed.

noxfile.py
sydduckworth: Whoops. This is what happens when you copy and paste. Fixed

noxfile.py
sydduckworth: No, I was just matching the existing behavior. Updated to now run in parallel.

tox.ini
sydduckworth: Whoops no I just forgot to add that. `devdeps` has been updated and I verified that the dependencies now match the previous implementation.

.github/workflows/downstream.yml
sydduckworth: The conditional isn't necessary since `nox` already depends on `generate-jobs` which has the same conditional. To me not having the same conditional defined twice seems less error-prone.

.github/workflows/ci.yml
sydduckworth: So at the moment all of the sessions are being run using a single tag (`core`) with a few exceptions:

- `jsonschema`, since it has a different trigger than the other tests. As a side note, that test only takes like a minute to run so it might be worth just having it always run then we could remove this block.
- `coverage`, since it seemed better to me from a readability perspective to just have the small amount of duplication rather than adding potentially fragile conditional logic. 
- The platform-specific tests, since again the current version seemed more readable than the alternative.

For the platform tests I could combine them into the same matrix as the core tests, which would look like this:

```yaml
core:
  name: ${{ matrix.session }}-${{ matrix.os }}
  needs: [core-generate-jobs]
  strategy:
    fail-fast: false
    matrix:
      session: ${{ fromJson(needs.core-generate-jobs.outputs.session) }}
      os: [ubuntu-latest]
      include:
      - session: core-3.12
        os: windows-latest
      - session: core-3.12
        os: macos-latest
  runs-on: ${{ matrix.os }}
 # rest of job definition
```

.github/workflows/downstream.yml
braingram: Thanks. The diff doesn't show the other conditional so I assumed it was inadvertently removed (since I didn't see a mention of the change). Removing it sounds good.

.github/workflows/ci.yml
braingram: This alternative for the platform-specific tests seems easy enough to follow with less workflow duplication. Running the jsonschema tests by default seems fine (especially since it allow for further cleanup of the workflow file).

From some testing adding:
```yaml
     - if: ${{ contains(matrix.session, 'coverage') }}
```
(followed by the existing `codecov/codecov-action` usage) handles the coverage session. I don't think this is any more fragile than the current setup (both depend on the coverage session name) and it further simplifies the workflow.

With these changes the workflow should only have steps:
- pre-commit
- type-checking
- core-generate-jobs
- core

is there a plan to also move the pre-commit and type-checking to nox?

sydduckworth: Okay, `jsonschema`, platform tests, and `coverage` are all merged into the main test job now.

A concern I'd have about moving type-checking into nox is that right now the pyrefly action emits linter errors as PR annotations, which I think is quite nice. I suppose we could add a positional argument to the jobs to tell them they're running on github, but that seems potentially hard to maintain.

zacharyburnett: I think most popular CI workflows set env var `CI=true`, so maybe that could be a simple way to check?

noxfile.py
```diff
@@ -94,16 +170,16 @@ def _generate_pytest_command(self):
...
     ),
     Package("asdf-zarr", "https://github.com/asdf-format/asdf-zarr.git", tags=["asdf"], extras=["tests"]),
-    Package("asdf-wcs-schemas", "https://github.com/asdf-format/asdf-wcs-schemas.git", tags=["asdf"], extras=["test"]),
     Package(
-        "asdf-coordinates-schemas",
-        "https://github.com/asdf-format/asdf-coordinates-schemas.git",
-        tags=["asdf"],
+        "asdf-transform-schemas",
+        "https://github.com/asdf-format/asdf-transform-schemas.git",
+        tags=["asdf-schemas"],
```
braingram: Where in the CI are sessions with the `asdf-schemas` tag run?

sydduckworth: They're run as part of the downstream CI.

braingram: Thanks. It's confusing that these have tags defined here but then also tags defined in the parameterization in downstream. What's the benefit of the tags here?

sydduckworth: The additional tags are based on the way that the downstream tests were grouped in the old CI. Originally I was going to replicate that in the new CI but we ended up not going that route.

The benefit is just additional organization of the tests, so if you want to just run the third-party downstream tests or just the asdf downstream tests there's a convenient way to do that instead of needing to individually specify a lot of tests individually.
It's definitely a workflow that will be helpful for me in the future.

sydduckworth: Would it be less confusing to call the field `extra_tags`?

braingram: `extra_tags` sounds good to me or removing the extra tags since we're not using them.

sydduckworth: I ended up renaming the field to `category` and making it a scalar since `extras` is already used to refer to package extras and wanted to avoid confusion there.
I'd prefer to keep the extra tags since as mentioned I would find them useful for running tests locally.

braingram: `category` sound great!

braingram (approved):

PR #2109, opened 2026-07-23
Add nox session for building documentation

### LINKED ISSUE #2035: Sphinx 8 vs sphinx 9 compatibility
In implementing type hints (#2031) I learned that Sphinx 9+ is only available for Python versions `>= 3.12` and older Python versions use Sphinx 8. 
The problem is that Sphinx completely reworked how autodoc works, which significantly changes the behavior with respect to type hints.
Since `asdf` supports Python `>=3.10`, depending on which Python version you use you will encounter completely different Sphinx behavior.

Right now running the doc generation code with older Python versions fails because sphinx 8 doesn't correctly handle imports gated behind `TYPE_CHECKING` blocks (among other reasons). 
I was originally using the [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) plugin to resolve these issues but having that plugin enabled in Sphinx 9 causes a whole set of new problems. Additionally, regardless of version it seems to break cross-references for type aliases.

## Potential Solutions
* Move the doc generation code into its own unpublished package (in the same repo), set its required Python version to `>=3.12`, and `sphinx >=9.0`. To me this seems like the best solution because it allows us to use the latest Sphinx version and it doesn't impact the main package.
* Update the sphinx config to enable [`autodoc_use_legacy_class_based`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_use_legacy_class_based), which forces sphinx 9+ to use the old autodoc implementation. This would ensure parity across Python versions, but it would also break type hints in the ways described above.
* Change nothing, and just add a note for developers that you need to be using Python `>=3.12` in order to build the documentation.

sydduckworth:
## Description
- Moved docs dependencies from a package extra to a `docs` dependency group since the docs dependencies shouldn't be visible to downstream packages
- Added minimum python version >= 3.12 for `docs` group. This is only enforced for `uv` currently.
- Added `docs` session to `noxfile.py`
- Updated `.readthedocs.yaml` to use `uv` since RTD doesn't support dependency groups with `pip` (even though pip itself does)

I'd previously mentioned that a problem I had with using tox to build the documentation is that it adds a lot of overhead to local docs builds. 
Now that we're using nox this isn't an issue because you can use nox's `-R` flag to reuse the existing virtual environment and skip re-installing dependencies, so there is essentially no overhead to using nox vs building directly.

Closes #2035 (in my opinion)

## AI Disclosure
No AI tools used

## Tasks

</details>

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -412,7 +412,11 @@ def type_checking(session: Session) -> None:
 
 @nox.session(tags=["docs"], python="3.14", default=False)
 def docs(session: Session) -> None:
-    """Build sphinx documentation."""
+    """Build sphinx documentation.
+
+    If the environment hasn't changed, you can speed up the build by using nox's `-N` flag
+    to reuse the virtual environment and skip reinstalling dependencies, e.g. `nox -N -s docs`.
+    """
     Asdf(extras=[]).install(session, *nox.project.dependency_groups(PYPROJECT, "docs"))
 
     session.run(
--- a/noxfile.py
+++ b/noxfile.py
@@ -414,8 +414,8 @@ def type_checking(session: Session) -> None:
 def docs(session: Session) -> None:
     """Build sphinx documentation.
 
-    If the environment hasn't changed, you can speed up the build by using nox's `-N` flag
-    to reuse the virtual environment and skip reinstalling dependencies, e.g. `nox -N -s docs`.
+    If the environment hasn't changed, you can speed up the build by using nox's `-R` flag
+    to reuse the virtual environment and skip reinstalling dependencies, e.g. `nox -R -s docs`.
     """
     Asdf(extras=[]).install(session, *nox.project.dependency_groups(PYPROJECT, "docs"))
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -113,7 +113,7 @@ class Package:
 class Asdf:
     """Helper class for managing ASDF installation and testing options."""
 
-    parallel: bool
+    parallel: bool = False
     show_slowest: int | None = 10
     extras: list[str] = field(default_factory=lambda: ["all", "tests"])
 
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -402,9 +402,30 @@ def type_checking(session: Session) -> None:
...
+    Asdf(extras=[]).install(session, *nox.project.dependency_groups(PYPROJECT, "docs"))
+
+    session.run(
+        "sphinx-build",
+        "-E",
+        "-T",
+        "-W",
+        "--keep-going",
+        "docs",
+        "build/html",
```
braingram:
```suggestion
        "docs/_build/html",
```
The existing MakeFile uses the suggested directory. We don't need to stick with that but we shouldn't use `build` as it's already used for packaging.

Any reason to not use the MakeFile here?

sydduckworth: I copied the invocation from RTD. In general I don't think putting putting stuff into `build` subdirectories should be a problem since tools should never assume exclusive access to the build directory, but agreed that in this case there's no reason to override the makefile.

sydduckworth: After some additional thought, I actually prefer not using the makefile. IMO the makefile doesn't really provide much benefit over just invoking the command directly from nox, while also adding an extra layer of indirection. 
It also makes it less intuitive to modify the command, since with direct invocation extra arguments can just be appended to the nox CLI invocation, whereas with the makefile you have to set the `SPHINXOPTS` variable.

I'd also prefer to keep the directory as `build` rather than `_build`. The use of `_build` comes from the `sphinx-quickstart` command, and is done so that if the sphinx source files are in the project root `build` doesn't get incorrectly picked up as a source directory. This isn't an issue in our case since the source files are located in `docs`.
Typically projects have a single `build` directory (or `target` directory if you're Rust) with subdirectories used by different tools as needed, so for me a `_build` directory is a bit unexpected. I generally expect built documentation to be in `build/docs` or `build/html` along with all other generated artifacts.

braingram: The makefile pattern is common across many repositories so I don't think we should remove it. Having the built docs put in different places for nox vs Makefile is confusing. I don't think we should put anything in `build` since that's used for package building. How about `docs/build` and updating the Makefile to match?

sydduckworth: Putting it into `docs/build` runs into the problem above where sphinx may pick it up as a source directory.

We can stick with `_build` but I don't see the advantage of invoking the makefile from nox? It doesn't seem worth it to pull in an entire additional build system (and a 134 line file) to avoid repeating a couple of CLI flags. Also see my above comment about making it harder to modify the arguments.

zacharyburnett (approved):
looks good to me at least. I imagine there might be a possibility of `uv` updating and breaking something with the RTD build, but RTD can handle that and we can deal with that if it comes up. Are there any other issues other than the makefile discussion?

noxfile.py
zacharyburnett: I think this really depends on who's using the makefile, and who's building the docs for asdf. As long as the docs build process has a clear documentation in the readme or something I think users who want to build docs locally will be able to figure it out without doing `make html`. I agree with @sydduckworth that nox takes over the utiliy of the makefile here, so having a makefile as another entry point to the docs build is one more place that something can be misconfigured

braingram (approved):
This looks fine to me. I don't see myself using the nox session for building the docs as the Makefile is more convenient, handles other formats and supports `make clean` for refreshing the auto-generated api docs. As long as we're keeping the Makefile I don't want to hold up this PR.

PR #2125, opened 2026-08-28
Add warning when conversion fails with default `warn_on_failed_conversion`

sydduckworth:
## Description

Part of #2127

- Added `is_set` function which can be used to check if attributes of an object have been modified along with an `_IsSet` base class necessary for it to function
- Added `asdf.exceptions.AsdfFutureWarning`
- Updated file loading logic so that if node conversion fails and `warn_on_failed_conversion` hasn't been manually set then `AsdfFutureWarning` is emitted before an exception is raised. 

PRs to address downstream test failures:
- https://github.com/astropy/asdf-astropy/pull/337
- https://github.com/spacetelescope/stdatamodels/pull/814

## AI Disclosure
No AI tools used

## Tasks

</details>


asdf/config.py
```diff
@@ -476,6 +478,13 @@ def warn_on_failed_conversion(self) -> bool:
         Returns
         -------
         bool
+
+        Warnings
+        --------
+        In a future release the default value will change from `False` to `True`.
```
braingram, line 482: This helpfully describes the change.
Would you add similar text to the docs?
https://asdf-website--2125.org.readthedocs.build/projects/asdf/en/2125/asdf/config.html#warn-on-failed-conversion

asdf/util.py
```diff
@@ -36,10 +40,14 @@
     "get_base_uri",
     "get_class_name",
     "get_file_type",
+    "is_set",
```
braingram: Let's keep the new API private. Would you move it to a private submodule or add underscore prefixes (pending response to other comments)?

asdf/yamlutil.py
```diff
@@ -351,6 +354,16 @@ def _walker(node):
                     warnings.warn(f"A node failed to convert with: {err}", AsdfConversionWarning)
                     obj = node
                 else:
+                    if not is_set(get_config()).warn_on_failed_conversion:
```
braingram, line 357: What are options for consolidating this code with the updates in lazy_nodes to avoid duplicating the error message?

asdf/config.py
```diff
@@ -464,7 +466,7 @@ def lazy_tree(self) -> bool:
     def lazy_tree(self, value: bool) -> None:
         self._lazy_tree = value
 
-    @property
+    @tracked_property
```
braingram, line 467: Would it be simpler to add a `__setattr__` to `AsdfConfig` and track what was set? Something like (with `_set_attrs` set to an empty set on `__init__`)?
```python
    def __setattr__(self, attr, value):
        if not attr.startswith("_"):
            self._set_attrs.add(attr)
        super().__setattr__(attr, value)

    def _is_set(self, attr):
        return attr in self._set_attrs
```
This could be extended to something like:
```
    def _warn_if_not_set(self, attr):
        ....
```
to consolidate the warning message (as mentioned in another comment).

braingram (commented):
Thanks for putting this together. I left a few suggestions.

What's your take/plan for the downstream errors?

asdf/yamlutil.py
sydduckworth: Idk the message is only in two places and will be removed following the next major release, so to me it doesn't seem like a problem. If we really wanted to avoid duplication we could create an `AsdfFutureWarning` subtype with that text as its message, but that seems like more overhead than the current approach.

asdf/config.py
sydduckworth: That is also a workable solution but I don't see any particular advantage over the implementation I went with. 
One reason I went with `is_set` as a separate wrapper type as opposed to adding an `_is_set` method to `AsdfConfig` is to avoid cross-module calls to a private method, which IMO is a code smell.

The current version is also generic and requires no modification to the containing class (other than the decorator) so it can easily be re-used for other classes in the future if needed.

asdf/util.py
sydduckworth: Moved from `asdf.util` to `asdf._helpers`

asdf/config.py
braingram: If the simpler alternative is workable let's go with that one.

I don't see a way we can avoid a cross-module call to a private method since we don't want these methods to be public and we need to use them from a different module (this applies to the current approach as well). But maybe there is something I've overlooking.

sydduckworth: Sorry, so you're saying you want me to replace the current implementation with the one you described above? To me it doesn't seem meaningfully simpler than the current version.

I'm also not clear on how the current implementation results in a private method call? It involves public methods imported from a private module, which to me conveys the intended use more clearly, i.e. I expect a private module to be used in other modules in the same package, but I don't expect private methods of a class to be used outside the current class/module.

braingram (approved):
Thanks! This warning will hopefully inform some downstream users and the docs and details helpfully describe the upcoming changes and how code can be adapted.

Issue #2036, opened 2026-04-29
Path forward for integrating types in docstring with type hints


sydduckworth:
Now that we've started added type hints to the codebase we need to decide how we want to handle the existing docstring-based typing.

Many (most?) major open-source projects are moving away from docstring-based typing in favor of type hints.
On the other hand the maintainers of numpy are [pretty firmly in favor](https://github.com/numpy/numpydoc/issues/664#issuecomment-3555080020) of maintaining docstring type info in addition to type hints.

The argument in favor of keeping docstring typing, as described by the numpy devs in the link above, is that the docstring types can be more human-readable than type hints, e.g. a function accepting a `list of strings or integers` is more readable than `Sequence[str | int]`.
They also point out that if your project keeps type annotations in separate `.pyi` files then the type hints may not show up in a user's editor whereas docstrings will.

The argument against docstring typing is that it is really hard to keep the docstrings in sync with the actual code.

Personally I'm pretty strongly in favor of removing docstring typing (obviously only as we replace it with type hints). 
I exclusively write code in an editor with an LSP and a type-checker so I'm pretty much never relying on the raw docstring for type information. 

### Removing Docstring Typing

If we do want to remove docstring typing then we probably want Sphinx to inject the types from the type-hints into the documentation body.
By setting `autodoc_typehints = "both"` the type-hints should be injected for docstrings without types while docstrings with types should be left unchanged.
Unfortunately from my testing there seems to be a conflict between `autodoc` and `numpydoc` that causes the type injection not to work properly (you end up with a second parameter list with just the types). A workaround would be to switch from `numpydoc` to the builtin `sphinx.ext.napoleon`, which largely does the same thing and doesn't conflict with `autodoc`. From my limited testing making that change doesn't seem to break the existing formatting but I also didn't check all of the documentation.

A side-effect of the change is that `napoleon` tries to resolve the docstring type(s), which generates a huge number of missing ref warnings.
There may be a way to disable that but if there is I couldn't figure it out (setting `napoleon_preprocess_types = False` doesn't make a difference). We also might be able to add ignores for the warnings without a huge amount of effort (most of the warnings are from trying to resolve `optional` as a type).

Another option is [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints), which is supposed to work with `numpydocs` but in my testing it also didn't work correctly (it would add types to the first parameter in a function but not the rest).

braingram:
[...]

PR #2139, opened 2026-09-18
Replaced `automodapi` in Sphinx docs with `autosummary`

sydduckworth:
## Description
- Removed usage of `automodapi` in Sphinx docs and replaced with `autosummary`
- Removed `graphviz` dependency since its no longer needed 
- Nox `doc` job now always removes the generated `api` directory before building since the overhead of regenerating it is negligible.

I tried to minimize changes to the downstream docs, but removing `automodapi` has had a few effects:
- Slight change to the organization and wording of the API docs
  - The pages in the user API are not suffixed with package/module now, e.g. it's now just "asdf.typing" instead of "asdf.typing Module"
  - The modules in the developer API are now each nested on their own page instead of all being on one big page
- `__version__` is no longer part of the generated docs. This seems fine to me in part because the documentation currently is just the generic docstring for `str`, but I can add it back if this would break anything in other docs.
- Classes no longer have a graphviz inheritance graph

I originally planned to swap out `numpydoc` for `sphinx.ext.napoleon` and add `sphinx-autodoc-typehints` but those create conflicts with our current docstrings so will have to be a future PR.

## AI Disclosure
No AI tools used

## Tasks

</details>


asdf/util.py
```diff
@@ -269,6 +269,7 @@ def get_class_name(obj, instance=True):
 #: Special value indicating that a parameter is not set.
 #: Distinct from None, which may for example be a value of interest in a search.
 NOT_SET = sentinel("NOT_SET", repr="NotSet")
+#: Deprecated alias for `NOT_SET`
```
braingram:
```suggestion
#: Alias for `NOT_SET`. Will be deprecated in a future release.
```
Let's reword this since we don't yet produce a DeprecationWarning on usage of NotSet.

docs/_templates/autosummary/base.rst
```diff
@@ -0,0 +1,5 @@
+{{ name | escape | underline}}
```
braingram, line 1: Would you point to an example where this edit from the stock template (name vs fullname) changes the docs?

docs/_templates/autosummary/class.rst
braingram, line 1: What necessitates adding these templates to the repository? What issues do we have to deal with when using the default templates?

braingram (commented):
I left a few comments. Overall the docs look reasonable but the addition of custom templates and sphinx hooks makes me think this solution is too custom. I'm not opposed to changing the documentation structure if we can use more sphinx (or non-sphinx) defaults and have a simpler documentation.

docs/_templates/autosummary/base.rst
sydduckworth: Using `fullname` results in the full module path being used as the page name in the side navigation TOC. 
So currently we have `User API -> asdf.config -> AsdfConfig`, whereas with `fullname` it would be `UserAPI -> asdf.config -> asdf.config.AsdfConfig`, which is fine for one item but becomes very hard to read/navigate when there are multiple items listed.

docs/_templates/autosummary/class.rst
sydduckworth: The custom template does a few things:

- It keeps the formatting largely the same as `automodapi` (which is where the templates originally came from)
- The class template implements filtering to only document inherited members of classes within certain modules. I originally was going to just show inherited members for every class, but it turns out that a bunch of upstream class methods break `numpydoc` when included. So this approach preserves the previous behavior and works around that issue.
- The class template also handles checking if a class field is a plain attribute or a property and then calling `autoattribute` or `autoproperty` as appropriate. This isn't handled by the default `autosummary` template which results in class properties not being properly annotated with type information.
- The module template handles filtering items that we want to ignore. Ignoring certain items is necessary because some items are included in the `__all__` list of multiple public modules (e.g. `ValidationError`) which Sphinx doesn't support, and some items break autodoc and need to be excluded entirely (e.g. `ByteArray1D`)

A lot of the functionality from the templates is taken from the GWCS Sphinx templates.

sydduckworth:
My goal is that in the future we can remove a lot of the custom processing in `conf.py`.

Vanilla Sphinx has a lot of limitations which need to be worked around. Right now we're in kind of a chicken-egg situation where `automodapi` is providing a few useful features but it's also blocking us from adopting other tools that would let us simplify the docs further. 

Originally my plan was to just switch everything over all at once, but as I described [here](https://github.com/asdf-format/asdf/issues/2036#issuecomment-5734750967) it turns out that `napoleon` and `sphinx-autodoc-typehints` will both require some changes to our documentation (or to how we handle warnings).
So for now I tried to get rid of `automodapi` while minimizing the number of big changes to the rendered docs.

docs/_templates/autosummary/class.rst
braingram: > The custom template does a few things:
> 
>     * It keeps the formatting largely the same as `automodapi` (which is where the templates originally came from)

I don't think keeping formatting the same is necessary (and outweighs the complexity of these templates and custom hooks).

>     * The class template implements filtering to only document inherited members of classes within certain modules. I originally was going to just show inherited members for every class, but it turns out that a bunch of upstream class methods break `numpydoc` when included. So this approach preserves the previous behavior and works around that issue.

Would you provide some examples of this? What docs would look different if we dropped these templates? 

>     * The class template also handles checking if a class field is a plain attribute or a property and then calling `autoattribute` or `autoproperty` as appropriate. This isn't handled by the default `autosummary` template which results in class properties not being properly annotated with type information.

Is there an upstream issue for this? Following up on what you mentioned in the description, what issue does the cause for type annotations? An example would be helpful.
 
>     * The module template handles filtering items that we want to ignore. Ignoring certain items is necessary because some items are included in the `__all__` list of multiple public modules (e.g. `ValidationError`) which Sphinx doesn't support, and some items break autodoc and need to be excluded entirely (e.g. `ByteArray1D`)

What issues do these objects cause? Can we address those instead of working around them in the docs?

> A lot of the functionality from the templates is taken from the GWCS Sphinx templates.

That's part of my concern. At the moment these templates live in a centralized location and is maintained (due to the use in astropy). I'm all for removing the use of automodapi if there is a good alternative but this PR seems like we're vendorizing the parts of it we currently use putting us on the hook to address future dependency issues (which often come up for sphinx), fix bugs, add new features, etc. Some of which we'll have to do for every repository that vendorizes that functionality.

braingram:
> My goal is that in the future we can remove a lot of the custom processing in `conf.py`.
> 
> Vanilla Sphinx has a lot of limitations which need to be worked around. Right now we're in kind of a chicken-egg situation where `automodapi` is providing a few useful features but it's also blocking us from adopting other tools that would let us simplify the docs further.
> 
> Originally my plan was to just switch everything over all at once, but as I described [here](https://github.com/asdf-format/asdf/issues/2036#issuecomment-5734750967) it turns out that `napoleon` and `sphinx-autodoc-typehints` will both require some changes to our documentation (or to how we handle warnings). So for now I tried to get rid of `automodapi` while minimizing the number of big changes to the rendered docs.

Thanks for the extra context. If the full solution is to add more extensions and remove the custom processing I'm in favor of doing that all at once. How much work would be needed to see what the full solution looks like (even if it's not 100% working with cross-linking etc)? It would be good to discuss the plan with some examples.

docs/_templates/autosummary/class.rst
sydduckworth: * Anything inheriting from `dict`, `UserDict`, `OrderedDict`, etc breaks because the formatting of the docstrings for some of the methods causes numpydoc to treat the entire docstring as a return type, which of course fails to resolve. This generates dozens of warnings and also generates incorrect annotations. Several of the methods in `semantic_version.Version` generate errors due to unexpected indentation. The `str.maketrans` method generates a list assignment out of range warning.
* I'm not aware of an upstream issue. With the default template the property doesn't get a type annotation at all. 
* Sphinx breaks when using subscript notation on certain types, e.g. `np.ndarray[tuple[int], np.uint8]`. Because of the way that Python works this only affects type aliases (and maybe other uses?) that don't include a union. So `np.ndarray[...]` causes the parser to break, but `np.ndarray[...] | int` doesn't because the inclusion of the union `|` causes Python to convert the whole thing into a string. I have spent a long time looking into it and I don't think there is really a fix other than hacking the docs.

